PR #3, opened 2014-12-31
CODE-2790

thpr:
Remove unused Imports


thpr:
My lack of understanding got and GitHub.

I opened a PR with only one patch committed and something decided my second
commit should be included in the PR even though the order of ops was
commit, PR, commit.

I'm very interested in knowing how to avoid that given that I don't want to
have to stage commits/PRs over days to avoid combining things...

TP

Zaister:
I guess the best tactic is to use a branch for each individual issue. 
These branches can then easily be deleted once the issue is closed.

Stefan.

LegacyKing:
Hi Tom,

That's the reason for one issue per branch. If you need to amend your PR, pushing a fix into the branch includes it in the PR (till it's pulled into the requested) since it's a request to pull an update from that branch - though no warning is given to the reviewer if any additional changes have been made. I'm curious why this is still staged though. James comment looks like he's requesting a javadoc.

thpr:
The java doc is now in another commit.  I'm guessing James is enjoying New
Years Day...

TP

jdempsey-au:
Interesting - I didn't get an email when a new commit was added. I'll merge it in now.

PR #4, opened 2015-01-01
Fix: DATA-2256

LegacyKing:
Paladin Mounts are not generated correctly in 3.5e


LegacyKing:
Silly me has a project underway on my master branch, and since I didn't want to inadvertently push in Tom's code work and those commits, I figured this was the easier work around. (I branched off an earlier commit of mine, which is really cool!)

PR #79, opened 2015-01-14
[PF] Animal Archive - Companion archetype not enable additional feats se...

mjmeans:
...lection


LegacyKing:
Mark, in the future please include the JIRA ID:

DATA-2310
[PF] Animal Archive archetype enabled feats not selectable 

When your PR comes over the Subject will have: DATA-2310

And the comment will be:
[PF] Animal Archive archetype enabled feats not selectable 

That way the JIRA can grab the fact this issue ties to the JIRA.

Cheers,

mjmeans:
I think I did that on the last two PRs. But I thought since the PR was coming from a branch with the correct JIRA ID also in the subject would be redundant.

 

Sorry,

Mark

PR #112, opened 2015-01-24
NEWTAG-476

thpr:
Create a DISPLAYNAME for SizeAdjustment so that the first token can be consistent in LST files


code/src/java/plugin/lsttokens/sizeadjustment/DisplayNameToken.java
```diff
@@ -0,0 +1,65 @@
...
+package plugin.lsttokens.sizeadjustment;
+
+import pcgen.cdom.enumeration.StringKey;
+import pcgen.core.SizeAdjustment;
+import pcgen.rules.context.LoadContext;
+import pcgen.rules.persistence.token.AbstractNonEmptyToken;
+import pcgen.rules.persistence.token.CDOMPrimaryToken;
+import pcgen.rules.persistence.token.ParseResult;
+
+public class DisplayNameToken extends AbstractNonEmptyToken<SizeAdjustment>
```
jdempsey-au: A comment would be useful here.

PR #118, opened 2015-01-24
Remove error message seeking svn info.

LegacyKing:


LegacyKing:
@jdempsey-au please review. (I ran a build on my local and it ran without an issue)

PR #137, opened 2015-01-27
J2816: Updated Permissions on Gradlew to 777

StumpyFrostreaver:
Fore Jira CODE-2816: Updated Permissions on gradlew Script (I haven't moved the Jira through any states yet as I don't have any permissions in Jira except that of user :-))


StumpyFrostreaver:
Scratch that... will resubmit from Stumpy/PCGen/master

PR #143, opened 2015-01-29
made price MAGIC_COST 8500 + MASTERWORK (MWORKW). Total is 8810 (10 for ...

StumpyFrostreaver:
...short sword, 300 for Masterwork and 8500 Magic). I could not see a property that would bestow the magic on the weapon.


LegacyKing:
Which magic are you looking for? +1?
EQMOD:PLUS1W
aka Plus 1 for "Weapon" Typical is up to 5. Other properties are found as well.

PR #198, opened 2015-03-03
[Pathfinder] Psionics Expanded updated files. [I thought I had already d...

Distant-Scholar:
...one this.]

This should add in the files for Psionics Expanded, if they are missing.  Which I guess they are.


Distant-Scholar:
Maybe I did already do this.  Or something.

PR #223, opened 2015-03-15
[Pathfinder] BotD Vol 1: Princes of Darkness- Host Devil Noxious Breath ...

DreamingPsion:
...is not allowing for Ability Focus

now allows


LegacyKing:
just an FYI, missing the JIRA tag - i.e. DATA-2382

PR #227, opened 2015-03-22
DATA-2387: Load fixes to to SGGAH to reflect changes in core sets

ElroTheOnk:
Fixed load errors from SGGAH to reflect two changes:
1) "Ability-ising" of feats
2) "FACT-ising" of SpellType

Both relatively simple fixes!


LegacyKing:
Not exactly sure about the FACT-ising, didn't see any FACT at the review, but looks clean and good.

ElroTheOnk:
Re the FACT-ising: the console log reported it as a problem from the FACT system (I'm guessing some simplification/use of FACT of the SpellType tag upstream in the core set which broke the PREABILITY I was trying to use in the CLASS line in this set). I imagine it never worked as I intended, and the fix I made is a much cleaner answer to what I was trying to do - so all-round an improvement :-)

PR #234, opened 2015-03-31
Correct Companion Archetypes - prereqs acting up and a bad entry

LegacyKing:
This isn't tied to JIRA. But basically three Archetypes for companions were unable to be taken. Altered those to work. Also racer had the charger descriptive text, fixed that as well. Lastly, claws needed to be claw.


LegacyKing:
This came in as a plea from the help desk email channel... in case people were wondering.

PR #240, opened 2015-04-21
DATA-2408

Zaister:
Fix bite attack strength bonus damage and damage type for Toothy half-orc
racial ability.


Zaister:
Test failure seems to be from a previous commit.

PR #331, opened 2015-07-08
Martial Flexibility Archetype Tweaks

jamrenaissance:
Modified the Martial Master archetype to prevent it from selecting
Weapon Trainings unnecessarily.

Modified the Warsighted archetype to properly prevent the selection of
Oracle Revelations unnecessarily.


LegacyKing:
I'll review this after sleep. Something is very off, if the abilitypool tags are actually required. Means something broke in the archetype code, or the types are not coded up correctly.

jamrenaissance:
Any update? Is there something I should look for. I went back and doublechecked and was still getting the faulty pools.

LegacyKing:
Sorry, busy monkey. Anyways, to the topic - we should not be doing the work twice, means we'll have bad results when someone like me comes along and fixes it.

ArchetypeName <> CATEGORY:Archetype <> TYPE:ClassFeatureToDisable <> !PRExxx:1,CAT=Arch,TYPE.ClassFeatureToDisable exists and works in all the other existing class archetypes. 

Problem here is you are encountering a broken piece of code, and fixed it to the best of your ability, but the fix isn't local to the set, it's local to the classes that should be working. 

Sadly, I think this requires a chimp to investigate and fix correctly. 

Let me do some comparisons and tinker a bit. I'll have an answer shortly.

jamrenaissance:
Thanks. I'll similarly avoid other "patches" until a more expansive solution is found.

LegacyKing:
Just an update: Oracle Revelations had been missing any limiters. I'm patching a fix in to handle that. Next I'll investigate the Fighter Weapon Training.

LegacyKing:
CR and APG both had issues that hadn't been caught. I've addressed those.

Retest and you should find your patch isn't required to negate the pools now.

LegacyKing:
Issue resolved by fixing bad code in the core book.

PR #359, opened 2015-07-15
Run HTML Tidy on all docs. Convert HTML 4 to XHTML 1.0.

LiaungYip:
Automatically corrected a multitude of sins (whitespace in URLs, unescaped entities, empty `<p>`, missing tags...)
(Command used: `find -name "*html" | xargs -L 1 -t tidy -m -q -asxhtml 2> tidy_logs_run_1.txt`)


LiaungYip:
This is a very noisy commit, but entirely automated - no *content* has changed.

There are several hundred HTML errors remaining after the conversion process, that HTML Tidy couldn't figure out on its own - these errors will be manually squashed.

PR #373, opened 2015-07-18
Update links from `sourceforge` to other places as appropriate

LiaungYip:
There are some specific files still linked to on Sourceforge - leaving those links as-is for now.


stphnclysmth:
Ironically, this build is failing for the very reason that the links need to be updated.

LiaungYip:
@stphnclysmth : you mean how SourceForge's house has melted, and they're running around with their pants on fire?

LegacyKing:
Yeah... however, these links won't fix that to my knowledge unless documentation has that drastic of an effect. I believe it's a code side deal. But the code team wasn't going to move on it since SF is down.

PR #381, opened 2015-07-25
CODE-2830

thpr:
Fix: Race condition on Custom Equipment Load v Size Load needs resolution
Fix: CODE-2863: Option Save Custom Equipment can cause PCGen to be uanble to load sources


jdempsey-au:
Looks good to me

PR #387, opened 2015-07-26
CODE-2861 - Campaign Story is not saved

thpr:
Allow multiple empty Chronicle Entries to exist on a PC


code/src/java/pcgen/cdom/facet/fact/ChronicleEntryFacet.java
```diff
@@ -19,10 +19,14 @@
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.IdentityHashMap;
```
jdempsey-au: Are these new imports needed?

PR #395, opened 2015-07-26
CODE-2829 Make Ability Tokens editor friendly

thpr:
Note this is not a complete fix for CODE-2829 since the converter still doesn't leverage this new function, but this is a separately usable unit of work, so I'm proposing as its own PR.

Should be noted for future editor work - note that the responsibility for converting:
AbilityCategory.class "SomeCat" is now transferred from the token into the Reference Context (by using a different method on the Context).  

This does not change (materially) runtime behavior, since it still checks if the returned reference manufacturer is null

However, this allows an editor/converter to unconditionally return a Reference Manufacturer (or otherwise check an external perhaps unloaded in memory list) and decide whether to return null or not in order to indicate an error to the token (the editor may have an external list, a converter would probably do unconditional return of a manufacturer)

### DIFF
--- a/code/src/java/pcgen/cdom/base/CategorizedAbilitySelectionChooseInformation.java
+++ b/code/src/java/pcgen/cdom/base/CategorizedAbilitySelectionChooseInformation.java
@@ -24,7 +24,8 @@ package pcgen.cdom.base;
 
 import pcgen.cdom.choiceset.CollectionToAbilitySelection;
 import pcgen.cdom.content.AbilitySelection;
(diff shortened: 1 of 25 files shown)


LegacyKing:
@thpr this has conflicts.

jdempsey-au:
The conflicts are with the files I fixed the EOLs on. To avoid pulling those issues back in again, I recommend cancelling this PR, making a new branch from master and copying over the desired changes to the new branch.

thpr:
Closing per James' comments, will base on current master

PR #398, opened 2015-07-26
DOCS-323 Deprecate FEAT tag and friends.

LiaungYip:


LegacyKing:
Just one thing to be mindful of - notice the branch wasn't made from the "master" but brought forward the other commit changes. Not an issue in this case, but just something to be aware of. Cheers,

LiaungYip:
I did it intentionally - the commits both touched the same files, and didn't want to create a manual merge situation again. What's the etiquette for separate commits touching the same file?

LegacyKing:
That's a bit tricky. Ideally, each issue in a PR should be separated by the changes. I'm guilty of cross-polluting my commits in some cases, so not unheard of. However, each branch PR should focus on the specific issue being addressed. The first commit was all about the UDAM change (also, something to note, is you want to reference the issue in the title, like you've done, so the JIRA catches the change. Which is the main reason to keep issues separated. Keeps from JIRA capturing the wrong commits. If the cross pollution is unavoidable. Separate them out by different commits as best you can, and then do the PR for them all. At least with separate commits, it is easier to cherry pick the changes if an issue arises. Cheers,

PR #402, opened 2015-07-28
Archetype fact replacement

LegacyKing:
@jdempsey-au I appear to have a couple of Code patches in this??? Can you verify these are safe to merge...

PR #403, opened 2015-07-28
These are stalking each branch...

LegacyKing:
@jdempsey-au 
Probably more EOL.


jdempsey-au:
Uggh, that's not a good thing at all. Was this branched from master after 402 was merged?

@thpr

LegacyKing:
James, I pulled in his PREFACT fixes... but ever since, these three files are persistently following me around wanting to be committed. It's extremely annoying.

Yes, I branched from master and no matter which branch I switch too (pre-existing or otherwise) they pop up. I'm starting to suspect my git has corrupted, cause data files are popping up in master forcing me to delete before I can merge.

thpr:
What's the heritage of the PREFACT changes in 402?  Those deprecations were made a LONG time ago, well before EOL was corrected.  I don't think those changes should have been part of #402

thpr:
And what PREFACT fixes?  You shouldn't be pulling in any code that isn't in master...

LegacyKing:
I can reverse #402 - but I have a ton of work in that merge I don't want to lose.

LegacyKing:
I could have sworn I pulled this from Master - https://github.com/LegacyKing/pcgen/commit/6e8909c6d929baf27aa7598d35ae879522d0a076. If not, I can reverse those, cherry pick into a new branch and re done the work tied in there.

thpr:
So I think this is starting to demonstrate why some ReBase is important
The entire history is rewritten due to an old commit, with LOTS of new branches that are hard to track down who did what

thpr:
IMHO, 402 should be backed out and rebased and then checked to see what it actually does against a master, not have master merged into it.  Too complicated with showing data team modifying code

LegacyKing:
Okay, let me back out the change, and see how this rebase works out.

thpr:
So actually... revert doesn't take it out.  We should define a new head without it ever there... how does one do that in git?

thpr:
commit 8e03781445b87a7258db75ea2c52e5c8179e3814 should be made the master

LegacyKing:
Hard reset to that commit, not exactly pretty, but should be manageable. Not sure how well it'll translate up to github though.

thpr:
won't let me push to github - "not fast forward"

jdempsey-au:
Hi,

Please don't use force push or the like - I've seen too many repos
destroyed that way. I am concerned we are trying to do low level work too
quickly here. Lets stop, plan out our work and then act on it.

Cheers,
James Dempsey

thpr:
Not sure I'd even know HOW to force push :D 
But I think we do need a policy that includes the data team that long lived branches are rebased.  I had thought those were the policies we agreed to, but maybe that did not include enough people and needs wider participation.  I'm extremely uncomfortable that it's showing code files modified by data team since that really makes it hard to know who is making changes.

thpr:
Also think this probably needs to move to _dev or _experimental rather than in a PR

LegacyKing:
Let's move this to _exp.

PR #428, opened 2015-08-05
Null Commit

thpr:
For James' review & comment

Unless I figure out another way to do it, this is a "null commit" that brings my master back into harmony with pcgen master, so I can again sensibly use my repo.  (I'm open to other ideas)


jdempsey-au:
It looks safe

PR #431, opened 2015-08-05
CODE-2790 - Code Cleanup

thpr:
Generics Fixes
Appropriate Warning Suppression
Other minor cleanup


LegacyKing:
Quite a read.

code/src/java/pcgen/base/geom/GridPoint.java
```diff
@@ -121,8 +121,8 @@ public Number getPreciseY()
 	@Override
 	public void setLocation(double x, double y)
 	{
-		this.x = x;
-		this.y = y;
+		this.x = java.lang.Double.valueOf(x);
```
jdempsey-au: Why do you need fully qualified references?

code/src/java/pcgen/cdom/primitive/CompoundAndPrimitive.java
```diff
@@ -58,7 +58,7 @@ public CompoundAndPrimitive(Collection<PrimitiveCollection<T>> pcfCollection)
 				Logging.log(Level.WARNING, "Found duplicate item in "
 						+ pcfCollection);
 			}
-			primCollection.add(PrimitiveCollection.INVALID);
+			primCollection.add(PrimitiveCollection.FIXED.<T>invalid());
```
jdempsey-au: I don;t think I;ve seen that syntax before. Is it valid? If so, do we need to be that clever?

code/src/java/pcgen/rules/context/AbstractReferenceContext.java
```diff
@@ -127,13 +128,17 @@ public boolean validate(UnconstructedValidator validator)
 		return getManufacturer(c, cat).getTypeReference(val);
 	}
 
-	public <T extends Loadable> T constructCDOMObject(Class<T> c, String val)
+	public <T extends Loadable, U extends Loadable & CategorizedCDOMObject<U>> T constructCDOMObject(Class<T> c, String val)
```
jdempsey-au: This needs some robust commenting. I really don't follow what U is doing here.

code/src/java/pcgen/rules/persistence/util/TokenFamilySubIterator.java
```diff
@@ -23,7 +23,6 @@
 import java.util.Set;
 
 import pcgen.rules.persistence.token.CDOMSecondaryToken;
-import pcgen.rules.persistence.token.CDOMToken;
 
 public class TokenFamilySubIterator<C> implements
```
jdempsey-au: Comments here would be useful, particularly around the nature of C

jdempsey-au:
What's the motivation for the generic work? Was there an overall strategy to this?

code/src/java/pcgen/base/geom/GridPoint.java
thpr: Becuase it imports Point2D, which has Point2D.Double and thus there are two "Double"s that need disambiguation.

code/src/java/pcgen/cdom/primitive/CompoundAndPrimitive.java
thpr: Yes, this is valid Java.  It's how one eliminates the generics warnings in this case.
As to need to be clever or not??  Just trying to reduce the warnings in our code, since generics can help catch problems too

code/src/java/pcgen/rules/context/AbstractReferenceContext.java
thpr: The complexity provides error catching, in the sense that instead of generics warnings, we get the compiler checking that we have the correct object type.

thpr:
Overall to reduce warnings and have the compiler do more work in catching errors.  I have done this type of cleanup steadily over time, and almost never failed to find errors as I have cleaned up generics warnings, so I think it is a good set of changes to be continuing.

thpr:
Comments added to all of pcgen.rules.persistence.util (addresses what C is among many other things :)

jdempsey-au:
Thanks Tom, the javadoc is quite helpful

code/src/java/pcgen/rules/context/AbstractReferenceContext.java
jdempsey-au: Could you add a comment on this method please? I'm still uncertain what U is and how it affects things.

thpr: I have removed the "U" types from the file since they are confusing.  I am looking at other strategic changes to the context files, which should eliminate the need for that behavior, but is 6.7 behavior since it is dependent on AbilityCategory.FEAT not being "preloaded"

jdempsey-au:
Thanks for that last change Tom. I think this is fine to merge now.

PR #437, opened 2015-08-10
NEWTAG-508 - Convert DESC output token to Freemarker

thpr:
- Convert PC-based String storage to PCStringKey for isolation from
CDOMObject based storage
- Deprecate DESC token
- Enable PCStringKey based information for output


code/src/java/pcgen/cdom/enumeration/PCStringKey.java
```diff
@@ -0,0 +1,69 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
+ */
+package pcgen.cdom.enumeration;
+
+/**
+ * This is a Typesafe enumeration of legal String Characteristics of a PC.
```
jdempsey-au: Were these moved to a separate class to designate that they are output enabled? If so that might be useful in the docs

thpr: Doc'd

PR #444, opened 2015-08-14
DATA-2517: Removed unnecessary Ultimate Magic dependency from source

ElroTheOnk:
Separated PF Bestiary dependencies into separate lsts & PRECAMPAIGNed their loading to avoid PF Bestiary hard dependency (not necessary for character build)
(Also fixed a trivial typo with the CR of the donkey rat I spotted while I was at it...)


LegacyKing:
Trying to piece together the logic here:
B1 is required to make the Kit and Race work?

ElroTheOnk:
In an attempt to explain - though sure you've figured it out, the "why" might be of interest...

Yes (or at least for a subset of the races/kits). The source contains two types of races:

1) "New content" (pages 306-315), which are rules-complete in the source and don't have logical prereqs on the Bestiary. As well as being rules complete (and hence useful without the Bestiary), this category includes a couple of 0HD races and therefore is useful to the Player. These ones I have left in _races.lst and _kits_race.lst, and these always load when the source is loaded.

2) "Reskinned Bestiary creatures" (pages 254-255). The source merely has a list that says things like "Donkey Rat: ... Stats: As dire rat (Pathfinder RPG Bestiary 232), but without the disease ability". These logically absolutely require the Bestiary, but are minor reskins of creatures with use only to GMs making monsters. These ones I have moved to _races_bestiary.lst and _kits_bestiary.lst, and these only load when the Bestiary is loaded (i.e. when in "GM mode").

I play with quite a lot of near-novice players, and so very much dislike the proliferation of unnecessary sources on the Player side (it clutters both the UI and - for example - spell lists of Clerics etc. on the character sheets), so try and minimise unnecessary prereqs, especially just to get a couple of minor things for GM use only.

The Inner Sea World Guide is a critical source for even very simply built characters in Golarion though, as it has the master list of Golarion languages and deities (beyond the core 20). So I think it's a worthy goal to reduce its essential dependencies to just Core and the APG. Hope you agree!

LegacyKing:
No worries, just seemed off when I glanced at it.

PR #454, opened 2015-08-16
CODE-2790 Simplify Categorization to extend Loadable, not dependent on CDOMObject

thpr:


jdempsey-au:
@thpr looks good to me

PR #459, opened 2015-08-17
CODE-2881 Fix test and token

vincent-lhote:


jdempsey-au:
@thpr does this fit in with your feat changes?

thpr:
Could have been a side effect of them.  This is safe

PR #461, opened 2015-08-18
Fix Arcanist Exploit Count

ckwalsh:
Previously a level 1 Arcanist would have 10 exploits available to them immediately, which is incorrect. This increases the exploit pool size as a proportion of the character level and subtracts from the exploit pool if archetypes remove exploits at certain levels.

This is modeled after the Brawler Bonus Feats


LegacyKing:
@jdempsey-au I see no travis check included with this PR. (Not that it will matter for this submission).

ckwalsh:
This is my first pull request to pcgen, apologies for not being super familiar with your development/merge flow

LegacyKing:
@ckwalsh Nothing wrong with your end. I'm flagging the developer. There is something amiss with the fact this PR should have triggered a Travis CI build check.

jdempsey-au:
@LegacyKing I can't see why the build hasn't kicked in. We'll have to keep an eye on it.

PR #462, opened 2015-08-19
DATA-2462 [PF] Animal Archive - Fix archetypes not replacing companion abilities

mjmeans:
I've tested this with 3rd level druid, cavalier and samurai. Now all
normal animal companion class features will show in the UI and this
helps confirm that the archetype has replaced the correct animal
companion abilities.


mjmeans:
Hold off on this. It's not replacing the higher level abilities.

mjmeans:
There is still a problem with the implementation of multiattack.

mjmeans:
Trying to track this down.... Wolf (and Eagle and many other companions) get 1 primary attack (like bite) and no secondary attack. At companion level 9 it will be BAB +6, but that fact doesn't grant it another attack. Animal companions don't get second or third attacks due to high BAB. Only multi-attack can give them an additional attack. Companions at level 9 with less than 3 natural attacks get multiattack which gives them a second attack at the primary attack BAB-5. No matter how many hit dice they eventually get, companions never get any additional attacks due to high BAB. So BAB +10 would NOT have yet another attack.

mjmeans:
I'll create a new PR for the multiattack calculation issue because is it not related to the reported problem of the archetype not removing the replaced abilities. This PR is ready to merge.

PR #469, opened 2015-08-19
Newtag 560

Zaister:
Add VISIBLE support for equiment.


code/src/java/plugin/lsttokens/equipment/VisibleToken.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+		else if (value.equals("NO"))
+		{
+			vis = Visibility.HIDDEN;
+		}
+		else
+		{
+			ComplexParseResult cpr = new ComplexParseResult();
+			cpr.addErrorMessage("Unexpected value used in " + getTokenName()
+					+ " in Skill");
```
LegacyKing: If this is equipment, why is the error message referencing a skill?

Zaister: It seems I did not catch all occurences. I'll correct that.

code/src/java/plugin/lsttokens/equipment/VisibleToken.java
```diff
@@ -0,0 +1,99 @@
...
+import pcgen.cdom.enumeration.ObjectKey;
+import pcgen.core.Equipment;
+import pcgen.rules.context.LoadContext;
+import pcgen.rules.persistence.token.AbstractNonEmptyToken;
+import pcgen.rules.persistence.token.CDOMPrimaryToken;
+import pcgen.rules.persistence.token.ComplexParseResult;
+import pcgen.rules.persistence.token.ParseResult;
+import pcgen.util.enumeration.Visibility;
+
+public class VisibleToken extends AbstractNonEmptyToken<Equipment>
```
jdempsey-au: This class needs javadoc and a license header please

PR #470, opened 2015-08-19
DATA-2524: Ultimate Equipment PRECAMPAIGN & dependency clean up

ElroTheOnk:
Clean up interactions & dependencies between UE and UC & UM.

- UE does not now depend on UM or UC (previously it depended on UM but the UC dependencies had been stripped out).
- UE now loads cleanly with none, one or both of these sources also loaded (previously it threw a couple of errors if UC was not loaded, despite not having a PRECAMPAIGN on UC).
- Also added some TYPEs to UE weapons & weapon profs necessary for playing nicely with UC feats.


LegacyKing:
CI failing.

LegacyKing:

    org.junit.ComparisonFailure: Errors encountered while loading Pathfinder RPG Core Supplements expected:<[]> but was:<[file:/home/travis/build/PCGen/pcgen/data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_profs_weapon.lst (52): PObject Amentum (Javelin) not found; .MOD skipped.,

    file:/home/travis/build/PCGen/pcgen/data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_profs_weapon.lst (76): PObject Great Terbutje not found; .MOD skipped.]>

        at pcgen.persistence.lst.DataLoadTest.testLoadSources(DataLoadTest.java:169)

pcgen.persistence.lst.DataLoadTest > testLoadSources[Pathfinder Society] FAILED

    org.junit.ComparisonFailure: Errors encountered while loading Pathfinder Society expected:<[]> but was:<[file:/home/travis/build/PCGen/pcgen/data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_profs_weapon.lst (52): PObject Amentum (Javelin) not found; .MOD skipped.,

    file:/home/travis/build/PCGen/pcgen/data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_profs_weapon.lst (76): PObject Great Terbutje not found; .MOD skipped.]>

        at pcgen.persistence.lst.DataLoadTest.testLoadSources(DataLoadTest.java:169)

ElroTheOnk:
Ok - this is an expected unit test fail, as UC & UE used different names for the same weapon proficiency. I fixed them to use the same name, so would expect test chars with the old name to be different. What do I need to do to update the unit tests?

LegacyKing:
That isn't unit tests, that's just load tests. Unit tests are updated running pftests in gradle or the IDE. 

Pathfinder RPG Core Supplements - basic loader
Pathfinder Society - basic loader

ElroTheOnk:
Nah, already fixed those & repushed - the latest fail is indeed unit tests. Error on the latest Travis CI build is

pcgen.inttest.game_pathfinder.pcGenGUIPfrpgPaladinTest > testCode FAILED

LegacyKing:
Ah, my screen failed to update... it still shows only two commits. Do you know how to run the PF unit tests, or do I need to attempt to explain the process?

ElroTheOnk:
Afraid I need the whole explanation - never come across something that passed load but failed unit tests before. I understand roughly what happens in the background, but not what files to edit with the new expected output.

LegacyKing:
You never manually "edit" the Unit tests. 

We run the Unit Tests through gradle - I use the Eclipse IDE for this. It runs (and fails) and places the unit test files in the code/testsuite/output folder. Those are copied or compared against the files in csheets. If the changes are expected, you overwrite the files in the csheets and push those in. The CI reruns against those and compares the new results against the csheets to see if they match. Even an order of output will cause a fail. HENCE why we never edit, since the expected output and new output must be identical to not trigger a failure.

ElroTheOnk:
Any tricks to getting the IDE established with pcgen or is it pretty self-explanatory?

LegacyKing:
That is a whole different level. The concepts should be identical though. Install an IDE of your choice. Add Gradle support to the IDE if not native. Then in your "run" dialog, there should be the ability to see a list of commands - you'll want just the pfinttest (which is just the pathfinder unit tests). This can also be done in command line, but that is something @thpr or @jdempsey-au would be better at explaining.

LegacyKing:
Oh - http://wiki.pcgen.org/Building_PCGen

ElroTheOnk:
That was a voyage of discovery - let's hope I haven't destroyed anything :-) Thanks for your help!

data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_profs_weapon.lst
```diff
@@ -73,7 +73,7 @@ Taiaha.MOD				TYPE:Weapon Group Melee OneHanded Piercing
 Double Chicken Saber.MOD	TYPE:Weapon Group Melee OneHanded Slashing
 Broadsword (Nine Ring).MOD	TYPE:Weapon Group Melee OneHanded Slashing
 Terbutje.MOD			TYPE:Weapon Group Melee OneHanded Slashing
-Great Terbutje.MOD	TYPE:Weapon Group Melee OneHanded Slashing
+Terbutje (Great).MOD		TYPE:Weapon Group Melee OneHanded Slashing
```
LegacyKing: Is this weapon listed as "Terbutje, Great" somewhere? That would be the only reason to alter the KEY here.

data/pathfinder/paizo/roleplaying_game/ultimate_equipment/ultimate_equipment.pcc
```diff
@@ -51,8 +51,9 @@ EQUIPMENT:ue_equip_general.lst
 EQUIPMENT:ue_equip_magic_items.lst
 EQUIPMOD:ue_equipmods.lst
 KIT:ue_kits.lst
-# ue_spells.lst only required if UC is not loaded - it avoids a dependency there
-SPELL:ue_spells.lst|!PRECAMPAIGN:1,INCLUDES=Ultimate Combat
+# ue_spells_*.lst only required if UC/UM are not loaded - avoids dependencies
+SPELL:ue_spells_uc.lst|!PRECAMPAIGN:1,INCLUDES=Ultimate Combat
```
LegacyKing: I have nothing against the dependencies removal, however, I am not very keen on duplicating objects. If errata comes out, I need to track down a spell in multiple sources. Is the INCLUDE/EXCLUDE code not up to this task?

ElroTheOnk:
Yep, it's listed like that in UC (and the other way in UE, though it's the same weapon). To clean up dependencies between the sources I had to go with one or the other - I chose UC's as the "original" source (and the nearest to our more usual style, as it happens), and used OUTPUTNAME to get the output looking right in UE.

ElroTheOnk:
... on the duplication: I do agree, but it doesn't seem very user-friendly to REQUIRE loading Ultimate Magic just because one piece of equipment (out of the thousands) in Ultimate Equipment grants it as an SLA. There are a quite a lot of potential solutions I see, none of which are entirely trouble free:

1) PRECAMPAIGN the original source. Benefit: dead easy, Downside: but I don't *want* the original source & it's hardly required just because one item uses an SLA
2) INCLUDE the SLA from the original source. Benefit: easy to manage data, Downside: despite the user having specified not to load the original source, its spells appear in the class lists (& hence on the OS for e.g. Clerics that get every spell)
3) Duplicate the SLA in the new source, stripping out the CLASS entry so it doesn't appear in spell lists. Benefit: nice presentationally, Downside: Errata to spell text etc. would need to be fixed in two sources (if we remember!)
4) !PRECAMPAIGN on the items in question to only include them if the original spell source is loaded. Benefit: this is clean and probably what was meant when the user didn't choose to include UM, Downside: might well lead to "bug reports" when a user can't find an item in UE despite having loaded the source.
5) Don't have the item grant the SLA via SPELLS, just note the name in SPROP and move on. Benefit: nothing to manage in data, Downside: Not so nice presentationally on the OS.

No doubt there are others, but those are the ones that immediately occur!

Personally, I don't like #1 and #2; as I said a couple of days ago, overwhelming new players with unwanted options is a problem for (some of) my groups, which was the driver for this change in the first place. I could go for any of #3-#5.

The version of UE previously in the repo (James' work, I think) had already set up the duplication solution (#3) to remove the dependency on Ultimate Combat, so I went with that for Ultimate Magic as well.

If you have a brainwave to avoid the downsides of any of the above, I'd be delighted to implement :-)

LegacyKing:
6) Utilize FORWARDREF to remove any notifications, items not in book are not included, but otherwise dependencies are subsumed.

ElroTheOnk:
I'll give that a whirl later - need to go and deal with life for a bit...

ElroTheOnk:
Ok, looks good...

PR #471, opened 2015-08-19
DATA-2525: Fix typos in casterlevel variables

ElroTheOnk:

PR #472, opened 2015-08-19
DATA-2525: Fix typo in Alchemist caster level variables

ElroTheOnk:
Replace Alchemist_Caster_Level with Caster_Level_Alchemist


ElroTheOnk:
This is a more limited replacement for #471 that just fixes the Alchemist typo & appears to work...

PR #477, opened 2015-08-21
More companion work

mjmeans:
Okay so this is just a little cleanup. Pony has only hooves attack listed and that means they should be considered primary, because there are no primary attacks for the Pony. So I've fixed that by adding HoofPony natural weapon which is primary. And Pony uses that. Boar gore attack was using the wrong die size. Multiattack was not applying to all the correct animals, so that was fixed. Multiattack will no longer add a second attack to primary attacks when the animal companion has less than 3 natural attacks. This is because some animal companions that have less than 3 natural attacks have two different single primary attacks and no secondary attacks. The old way would add an additional attack on both primary attacks (giving the AC two additional attacks) which is wrong. So now the multiattack will simply list in its description that they get a second attack in one of the primary attacks if the total number of natural attacks is less than 3. And also fixed the plant companions to show the class features like the other animal companions and be able to use the archetypes.


mjmeans:
Ack. It doesn't like DISPLAYNAME in ability HoofPony. Strange that it actually works though!

data/pathfinder/paizo/roleplaying_game/core_essentials/ce_abilities_race.lst
```diff
@@ -225,15 +225,16 @@ CATEGORY=Internal|Base Size ~ Colossal.MOD
...
-Hoof			CATEGORY:Internal		TYPE:NaturalAttack.Secondary	DEFINE:NaturalAttacks|0	DEFINE:HoofAttacks|0		BONUS:VAR|HoofAttacks|2	BONUS:VAR|NaturalAttacks|1				BONUS:WEAPONPROF=Hoof|TOHIT|-5	BONUS:WEAPONPROF=Hoof|DAMAGE|-(STR/2)
-Tentacle		CATEGORY:Internal		TYPE:NaturalAttack.Secondary	DEFINE:NaturalAttacks|0	DEFINE:TentacleAttacks|0	BONUS:VAR|TentacleAttacks|2	BONUS:VAR|NaturalAttacks|1			BONUS:WEAPONPROF=Tentacle|TOHIT|-5	BONUS:WEAPONPROF=Tentacle|DAMAGE|-(STR/2)
-Wing			CATEGORY:Internal		TYPE:NaturalAttack.Secondary	DEFINE:NaturalAttacks|0	DEFINE:WingAttacks|0		BONUS:VAR|WingAttacks|2	BONUS:VAR|NaturalAttacks|1				BONUS:WEAPONPROF=Wing|TOHIT|-5	BONUS:WEAPONPROF=Wing|DAMAGE|-(STR/2)
+Hoof			CATEGORY:Internal		TYPE:NaturalAttack.Secondary	DEFINE:NaturalAttacks|0	DEFINE:HoofAttacks|0		BONUS:VAR|HoofAttacks|2	BONUS:VAR|NaturalAttacks|HoofAttacks				BONUS:WEAPONPROF=Hoof|TOHIT|-5	BONUS:WEAPONPROF=Hoof|DAMAGE|-(STR/2)
+HoofPony	CATEGORY:Internal		TYPE:NaturalAttack.Primary	DEFINE:NaturalAttacks|0	DEFINE:HoofAttacks|0		BONUS:VAR|HoofAttacks|2	BONUS:VAR|NaturalAttacks|HoofAttacks				BONUS:WEAPONPROF=Hoof|DAMAGE|MAX(STR/2,0)|PREVAREQ:NaturalAttacks,1
```
LegacyKing: Is there a reason we needed to add a duplicate attack when the fix could be placed directly on the Companion?

LegacyKing: BONUS:WEAPONPROF=Hoof|DAMAGE|MAX(STR/2,0)|PREVAREQ:NaturalAttacks,1 (Also, unless there has been a change to the var, don't equine companions get two attacks with their hoofs?  Attack 2 hooves (1d3); which means the code never triggers...Since NaturalAttacks is set to 2.

mjmeans: It’s not duplicate. The type is changed from TYPE.Secondary to TYPE.Primary. While the STR bonus and tohit bonus could be changed directly on the companion to “get the right number”, it would not actually change its hoof attack to primary, which it is. I don’t know if any other ability that depends on whether the attack is actually typed as primary or secondary. But just in case, that’s how I did it.

mjmeans: There is no “rule” that says “all equine companions get 2 hoof attacks”. While there are no companions I know of that only have 1 hoof attack (and I looked extensively), it doesn’t mean there won’t ever be any, because it is not a rule that all of them must have 2 hoof attacks.

 

In this case this attack was added exclusively to handle the one special situation of a pony where its hoof attacks are considered primary. The “BONUS:WEAPONPROF=Sting|DAMAGE|MAX(STR/2,0)|PREVAREQ:NaturalAttacks,1” code exists on all the other TYPE.Primary natural attacks. So to be consistent with the others, it was included here. However, you are right that it will not run unless something else changes the number of hoof attacks on the companion to 1.

LegacyKing:
Please post a comment when something has been updated. It is the only method for us to get notified of a change.

mjmeans:
I didn’t realize that. Sorry.

PR #481, opened 2015-08-21
NEWTAG-506 Deprecate SPEECHTENDANCY output token

thpr:


Zaister:
When this is deprecated, whatever replaces is should also fix the spelling error.

PR #486, opened 2015-08-22
NEWTAG-484

thpr:
Deprecate AGE.CATEGORY output token
Add pc.age as FreeMarker interpolation
Add pc.ageset as FreeMarker interpolation


LegacyKing:
All the other deprecate output stuff was merged, why is this one still pending?

thpr:
The others were only a move and an xml file change, what I rate as trivial,
this has code which doesn't hurt to have a second set of eyes on it

jdempsey-au:
Looks good

PR #490, opened 2015-08-22
NEWTAG-488 Deprecate BIRTHPLACE output token

thpr:


LiaungYip:
Related: does this mean that the `PREBIRTHPLACE` list file tag, and its brethren `PRECITY`, `PREREGION`, etc. are also deprecated? (Replaced by the `FACT` system?)

thpr:
No, they are not facts

PR #551, opened 2015-09-05
NEWTAG-527

thpr:
Deprecate UNITSET output token
NEWTAG-527 Export UnitSet as gamemode.unitset

### DIFF
--- a/code/src/java/pcgen/io/ExportHandler.java
+++ b/code/src/java/pcgen/io/ExportHandler.java
@@ -336,11 +336,9 @@ public final class ExportHandler
 
 			// data-model
 			Map<String, Object> pc = OutputDB.buildDataModel(aPC.getCharID());
-			Map<String, Object> mode = OutputDB.buildModeDataModel();
 			Map<String, Object> input = new HashMap<String, Object>();
 			input.put("pc", ObjectWrapper.DEFAULT_WRAPPER.wrap(pc));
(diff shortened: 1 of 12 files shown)


LegacyKing:
@jdempsey-au sounds like @thpr is waiting on this to be merged before he can look into the info / infovar rule export. (as per last post on _exp)

code/src/java/pcgen/io/ExportHandler.java
```diff
@@ -336,8 +336,10 @@ private void exportCharacterUsingFreemarker(PlayerCharacter aPC, BufferedWriter
 
 			// data-model
 			Map<String, Object> pc = OutputDB.buildDataModel(aPC.getCharID());
+			Map<String, Object> mode = OutputDB.buildModeDataModel();
```
jdempsey-au: Would it be better to have this call pass in the game mode to be used? That way the assumption of SettingsHandler.getGame() is localised.

thpr: Actually headed the direction that SettingsHandler.getGame() is not the way to access the game mode, that game mode information should be in facets under a DatasetID, so that there is a consistent method of access between different "levels" of information (and we get away from static items at runtime)

I have submitted an additional commit to this PR that makes game mode handling in output more self-consistent (since gamemodename was not consistent)

thpr:
Reevaluating

PR #552, opened 2015-09-05
NEWTAG-562, NEWTAG-563, NEWTAG-564

thpr:


thpr:
Note the SAVE/CHECK object consistency from previous PR was not resolved. Is the intent to rename PCCheck.class to PCSave.class?

LegacyKing:
CHECK > SAVE for consistency has my vote.

jdempsey-au:
@thpr Up to you on the rename - just pointing out that we had two names for the same thing

thpr:
ok, I think even if we are going to do it, I'd have it in a separate PR since it's just a refactor (and will hit a lot of files).

LegacyKing:
Sounds like a good plan to me. @jdempsey-au this get the stamp of approval then?

jdempsey-au:
Yes that sounds good to me.

LegacyKing:
@jdempsey-au  I should choose my words carefully, I meant to ask if we can merge this in, since Tom would submit another PR for the CHECK to SAVE change.

thpr:
I think we should merge this and wait on a rename until after the formula merge.  I am not thrilled about a rebase on top of a naming refactoring that is going to cut across PCCheck, the facets, PRExxx tokens, et al., so really don't want to have that as a priority at this moment.

PR #554, opened 2015-09-05
NEWTAG-528 Export Using Output name preference

thpr:
Add ability to export preferences

### DIFF
--- a/code/src/java/pcgen/output/publish/OutputDB.java
+++ b/code/src/java/pcgen/output/publish/OutputDB.java
@@ -240,7 +240,7 @@ public final class OutputDB
 	}
 
 	/**
-	 * Registers anew Boolean Preference for inclusion in the global Models.
+	 * Registers a new Boolean Preference for inclusion in the global Models.
 	 * 
 	 * @param pref
 	 *            The preference name, as identified in the preference file
--- a/code/src/java/pcgen/output/publish/OutputDB.java
+++ b/code/src/java/pcgen/output/publish/OutputDB.java
@@ -30,7 +30,6 @@ import pcgen.output.base.NamedModel;
 import pcgen.output.factory.ItemModelFactory;
 import pcgen.output.factory.SetModelFactory;
 import pcgen.output.model.BooleanOptionModel;
-import pcgen.system.PCGenSettings;
 import freemarker.template.TemplateModel;
 
 /**
@@ -216,7 +215,6 @@ public final class OutputDB
 	{
 		outModels.clear();
 		globalModels.clear();
-		triggerLoad();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


code/src/java/pcgen/output/publish/OutputDB.java
```diff
@@ -206,6 +215,68 @@ public static boolean isLegal(String interpolation)
...
+		Map<String, TemplateModel> hm = new HashMap<String, TemplateModel>();
+		for (NamedModel nm : globalModels.values())
+		{
+			hm.put(nm.getModelName(), nm);
+		}
+		return hm;
+	}
+
+	/**
+	 * Registers anew Boolean Preference for inclusion in the global Models.
```
LegacyKing: Possible small typographical error "anew" thinking this was meant to be "a new"

code/src/java/pcgen/output/publish/OutputDB.java
```diff
@@ -206,6 +215,68 @@ public static boolean isLegal(String interpolation)
...
+	 * Initialize the preferences we want to export to FreeMarker
+	 */
+	static
+	{
+		triggerLoad();
+	}
+
+	private static void triggerLoad()
+	{
+		registerBooleanPreference(
```
jdempsey-au: I'm worried about the order of operations as this is run from a static. Have the options been loaded by that stage of program startup? In general I like to avoid static initialisation, particularly if it has external dependencies - you lose a lot of control over the  order of things. I would generally setup this sort of object as a singleton instance (perhaps managed by spring, or some master object) to allow for easier ordering and testing.

thpr: Moved to constructor of PCGenSettings

LegacyKing:
@jdempsey-au  looks like Tom's updated this.

code/src/java/pcgen/system/PCGenSettings.java
```diff
@@ -104,6 +106,8 @@ private PCGenSettings()
 		setProperty(VENDOR_DATA_DIR, "@vendordata");
 		setProperty(HOMEBREW_DATA_DIR, "@homebrewdata");
 		setProperty(CUSTOM_DATA_DIR, "@data/customsources".replace('/', File.separatorChar));
+		OutputDB.registerBooleanPreference(
```
jdempsey-au: This constructor is called from a static initialiser and thus this code is run before the properties are loaded. See pcgen.system.Main.loadProperties(boolean) for where the properties are loaded. The OutputDB would need to be initialised after that method is complete (or after its current processing).

thpr: I feel I'm not properly interpreting your concern here.  This is capturing the preference name (not the value).  (We don't want to capture the value at ANY point because we always want to be able to react to what the setting actually is at that time rather than what it was at PCGen code load)

This registers the (constant) preference name, and that constant is loaded into a BooleanOptionModel.  When FreeMarker is run (and pcgen.prefName is used as an interpolation), then the getAsBoolean() method called, which chains through PropertyContext.getProperty() to determine what the setting actually is.... so the setting is only done "at runtime" - long after both the properties are loaded and OutputDB is initialized.

The use of false is because we have to load the default early on since that is otherwise hardcoded elsewhere that the preference is referred to.

Does that help?

jdempsey-au: Thanks Tom, that makes sense to me now. The initialization as written should be fine then.

PR #558, opened 2015-09-05
NEWTAG-520 Support face in FreeMarker

thpr:
Change GridPoint to OrderedPair per Formula discussion
Support OrderedPair type in FreeMarker
Deprecate FACE output token
support pc.face in FreeMarker


LegacyKing:
@thpr CI still failing

PR #559, opened 2015-09-05
NEWTAG-565 Eliminate Spell Point Cost

thpr:


LegacyKing:
I suppose this means I should remove that code from the data?

LegacyKing:
@jdempsey-au Associated Data was removed. This is clear to be merged once reviewed.

PR #573, opened 2015-09-07
Code 2896 - Reinstate MSRD integration tests

LegacyKing:
Re-enables MSRD unit tests
* Updates Elwood to use prefix naming
* Adds a MSRD Int test to the build, in the same style as the RSRD.


code/src/test/pcgen/inttest/game_modern/pcGenGUIIlyanaTest.java
```diff
@@ -0,0 +1,50 @@
+package pcgen.inttest.game_modern;
```
jdempsey-au: License header needed here.

code/src/test/pcgen/inttest/game_modern/pcGenGUIElwoodTest.java
```diff
@@ -0,0 +1,50 @@
+package pcgen.inttest.game_modern;
```
jdempsey-au: LGPL license header needed here.

build.gradle
```diff
@@ -346,6 +346,22 @@ task rsrdinttest(type: Test, dependsOn: 'jar') {
...
+    testClassesDir = sourceSets.slowtest.output.classesDir
+    classpath = sourceSets.slowtest.runtimeClasspath
+    forkEvery = 1
+    maxHeapSize = "512m"
+    include 'pcgen/inttest/game_modern/*Test.class'
+    testLogging {
+        exceptionFormat = 'full'
+    }
+
+	// Report each test as it is started.
```
jdempsey-au: The formatting seems a bit off here, or that a tabs vs spaces thing?

jdempsey-au:
@LegacyKing The xml and pcg files should be moved out of the archive rather than duplicated. That way their history is preserved also.

Have you considered making one or both of these tests use the FTL output engine rather than the legacy engine? See pcgen.inttest.game_pathfinder.pcGenGUIPfrpgPaladinTest for an example of the code. The hardest part is that the output is very close but not exactly the same as the legacy engine output so a comparison is needed to validate the new output.

Oh and could you put yourself down on the list of code monkeys in PCGenProp.properties please? :)

build.gradle
LegacyKing: Probably a tab vs spaces. There was a noted inconsistency when I copied it from the rsrd version.

code/src/test/pcgen/inttest/game_modern/pcGenGUIElwoodTest.java
LegacyKing: Then it was missing when we archived this. How do I do the license header?

LegacyKing:
Yeah, I can add myself to the properties. Haven't covered the license stuff (just got through the loops and how not to make the three most common mistakes). Very nice to finally understand some terms being used around here now. :)

LegacyKing:
I actually hadn't considered moving to the FTL output engine. Main goal was just to re-instate the unit tests first. Fix what I knew would be broken in the unit tests, and then get those back in place before the big migration push. 

As a bonus, I added the MSRD inttest to it's own group so I could select those like the Pathfinder, and a slight rename to make the MSRD tests consistent in being named "msrd_xxx".

LegacyKing:
@jdempsey-au - Addressed the requested items. 
Migration to the FTL output will be a separate PR. Just to keep things simple.

jdempsey-au:
@LegacyKing Looks good.

PR #582, opened 2015-09-08
CODE-2898 Update active unit tests to use FTL Engine, and add missing License Headers

LegacyKing:


code/src/test/pcgen/inttest/game_35e/pcGenGUIAliceTest.java
```diff
@@ -1,6 +1,28 @@
+/*
+ * pcGenGUIAliceTest.java
+ * Copyright 2015 (C) Andrew Maitland <amaitland@users.sourceforge.net>
```
jdempsey-au: The original authors have to be the ones to release copyright to the project for these files. We'll have to go back through the history to find who they were and get their written release, or even better have them do a PR adding the headers. However for most it will be wither Tom or myself.

So, please remove the license headers you've added unless they are new files you've added (i.e. the msrd test classes you added are fine, even though they are derived form older work).

LegacyKing:
@jdempsey-au updated to remove the invalid license header - I left the file name to be nice. I believe @thpr did the majority of the "Named" 35e unit tests, Alice, Bob, Charlies, etc. I think you did GoldieLocks and the Cleric PF test. The older 35e tests and 3e tests are beyond my time. Cheers,

jdempsey-au:
@LegacyKing That looks great. I had discussed with Tom leaving one or two tests that ran the old output engine until we officially retire that engine. Could you set one each of the 35e and pathfinder tests back to the old engine please and then I think this is right to go.

LegacyKing:
@jdempsey-au yeah, I suppose I can revert two tests back...

LegacyKing:
@jdempsey-au completed.

PR #592, opened 2015-09-09
Support PR#590, Have DataTest use "common" file list

thpr:


LegacyKing:
@thpr - Investigating the test failure to see if the files actually exist.

LegacyKing:
@thpr & @jdempsey-au  looks like legit files missing. I'll put in a patch for that.

LegacyKing:
Is jpg not a valid file format? I'm getting a lot of those, but the files actually exist.

LegacyKing:
@thpr & @jdempsey-au resolved the data tests side.

jdempsey-au:
@LegacyKing Did you need to comment out the cover files? Did they refer to files that exist? If so then the code is the problem not the data!

LegacyKing:
@jdempsey-au I think there are two maybe three issues going on.
1) The pcc files with a cover file that I commented out are never seen by the user (No gamemode entry). Thus a file reference to any visual items is wasted memory. They piggyback on the actual coded source books, and alter the rules.
2) There were legit problems - first few pccs I found they had files name mismatches. xxx.jpeg instead of xxx.jpg or xxx missing the .jpg.
3) The file references from the pfs hidden pccs were legit data issues, not code.

As an example, this was flagged as failing, which on review is correct
/pathfinder/paizo/roleplaying_game/core_rulebook/core_rulebook.jpg
the correct reference is
@/pathfinder/paizo/roleplaying_game/core_rulebook/core_rulebook.jpg

Since I didn't have any bestiary entries failing in the tests, and they had proper references (started with @) I felt it was probably a wiser course of action to disable all the hidden PFS reference calls, since as I mentioned before they are never seen by the user.

Cheers,

jdempsey-au:
Thanks for the clarification!

PR #596, opened 2015-09-11
NEWTAG-525 add pcgen.version interpolation

thpr:
Deprecate EXPORT output token
add pcgen.version interpolation


jdempsey-au:
@thpr I don't follow what this one is doing? Why the need to wrap a freemarker class?

thpr:
Closing due to conflicts

PR #602, opened 2015-09-11
DATA-2561 [RSRD] Animal Companion Progression Confusing to end users.

LegacyKing:


LegacyKing:
This PR displays all the abilities clearly, while doing the behind the scenes work.

PR #621, opened 2015-09-16
Fix bomb stats for the Crypt Breaker

pieterlexis:
Before, when one choose the Crypt Breaker archetype for the Alchemist, the number of alkahest bombs per days was not increased with the level of the PC. Also, the amount of bonus dice, the minimal damage for the splash-effect shown in the description and weapon statblock were not calculated correctly.


pieterlexis:
Hrm, I just realized the weapon statblock still tells me the type is fire, while the alkahest bomb does acid damage. I could create a new weapon (Alkahest Bomb) with the correct properties though (including the property text that it does Xd8 damage vs undead+constructs and Xd4 against other creatures). Comments?

LegacyKing:
Do we have a jira ticket attached to this issue?

Changing the type is not possible, without a new equipment item. The natural attack block could handle it easier, since that is just a simple ability. Maybe the better method would be a ASPECT:CombatBonus|xxx indicating the damage type by the bombs.

LegacyKing:
AUTO:EQUIP|Bomb	ABILITY:Natural Attack|AUTOMATIC|Bomb - change that if you want a different outcome. Since this fixes an existing issue I'll merge it in, however, it is best to log any issues in our jira instance - pcgenorg.atlassian.net is where our jira is held. Keeps track of issues, and that is how we generate the release notes. Cheers,

PR #622, opened 2015-09-18
DATA-2578 OUTPUTNAME causes magic items created via .COPY= to inherit…

LegacyKing:
… that name. Switch to KEY.


LegacyKing:
The error indicates a flaw in our testing system. Reruning travis to confirm.

However, the item "Sunblade" has two copies, and one has "Sunblade (Bastard)" while the other has "Sunblade (Short)" as their respective keys.

PR #631, opened 2015-09-25
Remove deprecated functions and calls 	- clean up calls to various internal and external deprecated 	  calls.

grimreaper:


LegacyKing:
theApplication = theApplication = new com.apple.eawt.Application(); -- looks like that file entry was removed. Everything else looks fine from my newbie review. Not sure if we normally edit things in the deprecated folder. Normally means they have been replaced by something else. That would be @thpr or @jdempsey-au to make that call.

thpr:
The challenge is that most of the changes here are based on the assumption that the files in plugin.exporttokens.deprecated can be removed immediately.  They cannot.  They are part of the software's "external API" and were just deprecated earlier in the 6.5.x cycle.  Thus, they are not yet ready to be removed.

The output sheets and users of our program are dependent on those items and we "guarantee" even the deprecated capabilities will work for some number of future versions (and in this case, we haven't even defined what version that is where the function will break).  So this patch is useful "someday" (but will likely be easier to reproduce when that time comes), but not at the present time.

grimreaper:
FWIW I run on a mac.  It works for me (OS X 10.11.1 (15B17c)).

jdempsey-au:
Looks great - thanks for the contribution!

PR #632, opened 2015-09-27
CODE-2790 Output Rebuild

thpr:
This rebuilds the output system to be dataset/charid dependent.
CDOMObjectWrapper is no longer a "static" that has to be loaded/cleared (helps unit tests and overall design)


code/src/java/pcgen/cdom/facet/CDOMWrapperInfoFacet.java
```diff
@@ -0,0 +1,145 @@
...
+
+	/**
+	 * Initializes this facet for the given DataSetID.
+	 * 
+	 * @param dsID
+	 *            The DataSetID for which this facet should be initialized
+	 */
+	public void initialize(DataSetID dsID)
+	{
+		set(dsID, CDOMObject.class, "source", new SourceActor());
```
jdempsey-au: Do these ever get cleaned up? Or do they build up as we switch sources?

code/src/java/pcgen/cdom/facet/base/AbstractSubAssociationFacet.java
```diff
@@ -0,0 +1,187 @@
...
+import java.util.Map;
+
+import pcgen.base.util.WrappedMapSet;
+import pcgen.cdom.base.PCGenIdentifier;
+
+public abstract class AbstractSubAssociationFacet<IDT extends PCGenIdentifier, S1, S2, A>
+		extends AbstractStorageFacet<IDT>
+{
+
+	public A get(IDT id, S1 obj1, S2 obj2)
```
jdempsey-au: It would be useful to have javadoc on these

code/src/java/pcgen/output/actor/FactKeyActor.java
```diff
@@ -40,6 +42,9 @@
  */
 public class FactKeyActor<T> implements OutputActor<CDOMObject>
 {
+	private static final ObjectWrapperFacet WRAPPER_FACET = FacetLibrary
```
jdempsey-au: Will the FacetLibrary always be initialized before any reference to the FactKeyActor class is made?

code/src/java/pcgen/output/actor/ObjectKeyActor.java
```diff
@@ -39,6 +41,9 @@
  */
 public class ObjectKeyActor<T> implements OutputActor<CDOMObject>
 {
+	private static final ObjectWrapperFacet WRAPPER_FACET = FacetLibrary
```
jdempsey-au: As with FactKeyActor I'm a bit concerned about order of operations here.

code/src/java/pcgen/output/model/CDOMObjectModel.java
```diff
@@ -75,14 +74,29 @@ public CDOMObjectModel(CDOMObjectWrapperInfo info, CDOMObject cdo)
 	@Override
 	public TemplateModel get(String key) throws TemplateModelException
 	{
-		OutputActor<CDOMObject> actor = info.get(key);
+		//TODO may not be entirely correct - really about ... what? formula scope??
```
jdempsey-au: Does this need to be addressed?

code/src/java/pcgen/output/actor/FactKeyActor.java
thpr: FacetLibrary.getFacet is "self-initializing" in that if the items is not already known, it is requested from Spring... so this usage is not an order of operations problem.

code/src/java/pcgen/cdom/facet/CDOMWrapperInfoFacet.java
thpr: They get cleaned up.  Here is the "process":

All the facets are weak references of the Identifier (CharID or DataSetID) to their stored information, so the question can be rephrased: Does the DataSetID get appropriately cleaned up when we switch sources?

The DataSetID is "owned by" the LoadContextInst, and I have no reason to believe we have a LoadContext-binding memory leak (we had one at one point, but that has been addressed).

So the assertion becomes: Assuming no one "holds onto" the DataSetID and the LoadContext gets forgotten, the DataSetID should be cleaned up. 

Thus in my mind the remaining question is: "Are there any 'bad actors' (non-facets) that maintain a reference to the DataSetID?"  Since the only other place the DataSetID is "captured" is in the CharID (so the PC can reference back to the loaded data), any retention will also be associated with a PlayerCharacter based memory leak, which I also think we don't have at the moment.  So I believe we are appropriately managing memory in this case.

code/src/java/pcgen/output/model/CDOMObjectModel.java
thpr: oops.  I will fix this.  It was addressed.  The use case that was a problem was something like SubClass (hence my comment about formula scope - which is actually not right either), but the solution was CDOMWrapperInfoFacet.getActor() will "fall up" - so SubClass can use PCClass and both can use CDOMObject.class (just like the tokens "fall up" to allow global tokens.

code/src/java/pcgen/output/actor/FactKeyActor.java
jdempsey-au: Spring still has to instantiate the classes it manages. So if class A is managed by spring and A refers to FactKeyActor, then the startup will fail. Even if it works today, it still creates a startup dependency order which will be lying in wait for us to trip over in the future.

code/src/java/pcgen/cdom/facet/CDOMWrapperInfoFacet.java
jdempsey-au: Right, so the part I missed was that the cachedMap only holds a weak reference to the keys.

LegacyKing:
@thpr how is this coming along?

LegacyKing:
@thpr we have a conflicted PR, any status update?

LegacyKing:
@jdempsey-au 3 months without contact, or updates. Branch is conflicted. Permission to close this for now. When @thpr returns, he can re-open when it's ready again.

PR #635, opened 2015-09-27
CODE-2790 Swap to using external base library

thpr:


jdempsey-au:
@thpr This looks fine, but I'll leave them merge to you so that you can do them in the right order.

LegacyKing:
@thpr this has a greenlight... cheers,

PR #639, opened 2015-09-29
CODE-2790: Clean up CharacterSheetPanel

grimreaper:
- change enum to include data attached instead of using a
  separate switch statement
- Use resource based try statements
- Be explicit about caught exceptions
- Don't explicitly init to null
- Use final where possible
- Use string formatting instead of manual concat
- Cut off lines at consistent 80 char mark

### DIFF
--- a/code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
+++ b/code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
@@ -23,8 +23,10 @@ package pcgen.gui2.csheet;
 import java.io.BufferedWriter;
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.IOException;
 import java.io.StringWriter;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -40,12 +42,14 @@ import org.lobobrowser.html.parser.InputSourceImpl;
 import org.lobobrowser.html.test.SimpleHtmlRendererContext;
 import org.lobobrowser.html.test.SimpleUserAgentContext;
 import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
 
 import pcgen.core.Globals;
 import pcgen.facade.core.CharacterFacade;
 import pcgen.gui2.PCGenFrame;
 import pcgen.gui2.PCGenStatusBar;
 import pcgen.gui2.tools.CharacterSelectionListener;
+import pcgen.io.ExportException;
 import pcgen.io.ExportHandler;
 import pcgen.system.ConfigurationSettings;
 import pcgen.util.Logging;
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
```diff
@@ -56,61 +60,36 @@
...
+		GREEN("preview_color_green.css"),
+		LIGHTGREEN("preview_color_light_green.css"),
+		RED("preview_color_red.css"),
+		LIGHTRED("preview_color_light_red.css"),
+		YELLOW("preview_color_yellow.css"),
+		LIGHTYELLOW("preview_color_light_yellow.css"),
+		GREY("preview_color_grey.css"),
+		LIGHTGREY("preview_color_light_grey.css");
+
+		public String cssfile;
```
thpr: This is actually risky, because it can be publicly modified.  public *final* might be doable, but even that is against our current standards enforced by checkstyle.  We should have an accessor (getter) method for cssfile based on our current code standards.  (Target is get all fields to private if we can)

code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
```diff
@@ -196,15 +174,13 @@ public void run()
 						}
 						catch (Throwable e)
 						{
-							final String errorMsg = "<html><body>Unable to process sheet<br>" +
-									e + "</body></html>";
-							ByteArrayInputStream instream = new ByteArrayInputStream(errorMsg.getBytes());
-							try
-							{
+							final String errorMsg = String.format("<html><body>Unable to process sheet<br>%s</body></html>", e);
```
thpr: I didn't count the exact length, but since we are changing code in this file, we should try to wrap at least the new lines to a good line length.  Our standard is 80, but our warnings are currently higher.  I think 90 or 100 is fine, but shouldn't be too long.

code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
thpr: @jdempsey-au Should have asked your preference here as well.  The current change can't be used (public can be modified), so either public final or have an get* method... if we allow public final rather than requiring a get* method, then we should consider what that means for our Checkstyle rules.  I much prefer the get* method for that reason (to basically drive things to be private)

LegacyKing: @jdempsey-au  I'm going to agree with @thpr here. Private data with public getter and setter methods is good coding practice. I think the term for this is called encapsulation. Yup, just did a discussion paper on it this week:  Also known as data hiding; this is where the variables of a class are set to private, which hides the variables from other classes, and then public setter and getter methods are set up to allow modification and viewing of those private variables. ;)

jdempsey-au: @thpr yeah I saw the comment and as I agreed didn't make any further note :) We use getters and setters everywhere. We should stick with that.

@LegacyKing great to see that you are digging into the theory of software development - encapsulation is very very useful in making software robust and manageable.

grimreaper: I much prefer using final public.  Using getters / setters is a violation of abstraction and can make code harder to reason about.  that said, I'll go with whatever the project style dictates

LegacyKing: Abstraction:

Abstraction in java can happen by using “abstract classes” and
“interfaces”. An abstract class or interface is common ideas that are
used by other objects. In java terms, abstract classes are ‘extended’ by
concrete classes. As an example, a dog (concrete class) can be said to
be a canine (abstract), and then furthermore an animal (abstract). The
java GUI classes are a great example of abstract classes.
Conversationally, an abstract class must pass the IS-A test for
inheritance. A canine IS-A animal, a dog IS-A canine. The other
mentioned abstract is interfaces. An interface does not need to pass the
IS-A test. I loved the example of Robot and Pet being interfaces. A dog
could be a Robot, and it can be a Pet, but not all dogs are a Pet, nor
are they all Robots. An interface is basically a set of ideas that can
be used by an object. The big point I learned is classes may only have
one abstract class above them, but may have as many interfaces as they want.


Not exactly sure how by following good coding practices of "hiding data"
that violates the principles of abstraction (common ideas). And since
this is literally the discussion for this week, I'm fairly certain I
have those basics down.

Cheers,

LegacyKing: p.s. email didn't send me the message that you flagged me on, so I did not see that till I was looking at the posts. Only saw this "I much prefer using final public. Using getters / setters is a violation of abstraction and can make code harder to reason about. that said, I'll go with whatever the project style dictates"

Cheers,

thpr: In this specific case, please do a getter (obviously no setter) for the field.

As far as the larger theory, and resolving if you are using the same definition of abstraction (hint: probably not), that's probably better for _dev than a specific PR.

code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
grimreaper: I typically wrap at 120 or so for Java due to the lengthier names, but I'll change the setting for pcgen.

grimreaper:
updated diff

LegacyKing:
@thpr anything else or is this good to be merged?

jdempsey-au:
@grimreaper Apologies for the delay!

PR #640, opened 2015-09-29
CODE-2790: Clean up IIFNode

grimreaper:
- use final where appropriate
- don't use "_" for params
- avoid issues with returning modifiable list
- use diamond operator

### DIFF
--- a/code/src/java/pcgen/io/IIFNode.java
+++ b/code/src/java/pcgen/io/IIFNode.java
@@ -21,25 +21,26 @@
 package pcgen.io;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
- * <code>IIFNode</code>.
+ * {@code IIFNode}.
  *
  * @author Mark Hulsman <mark_hulsman@users.sourceforge.net>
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/io/IIFNode.java
```diff
@@ -31,15 +32,15 @@
...
 {
-	private List<Object> falseChildren;
-	private List<Object> trueChildren;
-	private String expr;
+	private final List<Object> falseChildren;
+	private final List<Object> trueChildren;
+	private final String expr;
 
-	IIFNode(String _expr)
+	IIFNode(final String expr)
```
thpr: One of the code style discussions we have had and do not do is put final on parameters.  While technically possible and correct, there is a feeling it is harder to read, and thus we do it when having it is absolutely necessary (e.g. used in an anonymous class), but not in the general case

grimreaper: understood.  I don't know the pcgen style so I'm proposing changes and asking for feedback :-)

grimreaper:
I updated the diff.  Thanks!

PR #641, opened 2015-09-29
CODE-2790: clean up PCGIOHandler and PCGFile

grimreaper:
- remove unused functions
    - rename some variables
    - add more final

### DIFF
--- a/code/src/java/pcgen/io/PCGIOHandler.java
+++ b/code/src/java/pcgen/io/PCGIOHandler.java
@@ -257,7 +257,7 @@ public final class PCGIOHandler extends IOHandler
 		}
 	}
 
-	private boolean isPCGCersion2(List<String> lines)
+	private static boolean isPCGCersion2(List<String> lines)
 	{
 		for (String aLine : lines)
 		{
@@ -273,7 +273,7 @@ public final class PCGIOHandler extends IOHandler
 	 * @param in
 	 * @return
 	 */
-	private List<String> readPcgLines(InputStream in)
+	private static List<String> readPcgLines(InputStream in)
 	{
 		final List<String> lines = new ArrayList<String>();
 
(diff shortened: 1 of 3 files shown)


thpr:
This apparently overlaps PR#640 since it also include IIFNode

code/src/java/pcgen/io/PCGIOHandler.java
```diff
@@ -257,7 +257,7 @@ public void read(PlayerCharacter pcToBeRead, InputStream in,
 		}
 	}
 
-	private boolean isPCGCersion2(List<String> lines)
+	private static boolean isPCGCersion2(List<String> lines)
```
thpr: @jdempsey-au What do you think here?  I don't think we have a standard on this.  Honestly, I'm ambivalent to having things being static or not (don't have a strong opinion), so not sure it matters, but if things are going to be cleaned up, we should decide if this is worth the effort to be changed.

jdempsey-au: I'd leave them as instance methods unless there is a notable benefit to having them static.

thpr:
@grimreaper Please remove the two static items, and please resolve the conflicts created when the previous IIF PR was merged.  Once those are complete, we can merge this PR.

code/src/java/pcgen/io/PCGIOHandler.java
grimreaper: Why should they be instance methods?  It acts a lot more like a helper method than a property of the object itself.  Perhaps I'm wrong?

LegacyKing:
@grimreaper did you get the last note? Conflicts and corrections so we can merge this. Thanks!

LegacyKing:
This conflicted and no reply for almost a month.

LegacyKing:
@grimreaper Eitan, any update with this???

LegacyKing:
Closing this out as Eitan has already closed the other PR. I've resubmitted a fixed/non conflicted version.

PR #655, opened 2015-10-06
Code cleanup 	- make variables local from class 	variables to make it easier to follow

grimreaper:
(hopefully I'll be getting to 'real code'
 and not just cleanup shortly)


LegacyKing:
@thpr and @jdempsey-au - this look okay?

PR #678, opened 2015-11-03
CODE-2902 Greatsword allows Longbow to be equipped to the "Both Hands" slot, vice versa does not happen.

LegacyKing:


LegacyKing:
Brought in the fixes so the test can pass. Code changes are just the 2 to 1. Pretty basic.

jdempsey-au:
So this code change will mean that all two handed and double weapons will need to have hands:1 or the like. Is that what you intended?

LegacyKing:
Hi James,

Issue at hand: Today I can equip two Greatswords, a Crossbow and a 
Greatsword, or two Crossbows. All occupy the same "Phantom" slot. I.e. 
Both Hands.

After this fix:
I can only add one Greatsword, or Crossbow.

The gamemode change has HANDS at 2 still.  In fact, I added two new 
WIELD CATEGORIES to handle Ranged items, OneHandOnly, TwoHandsOnly which 
is a separate track to follow the rules.

If the concern is about HANDS, then I would say we have a major 
disconnect in the code system beyond my capability (Which I mentioned 
with the SLOTS, HANDS interactions we have all over in the tracker). Is 
HANDS in the wield category not being applied?

All I know is before the fix, inappropriate combinations are being 
allowed, after the fix, only appropriate combinations are allowed. (I 
also played with Longswords and Quarterstaves in my testing).

Everything works for me. If there is a deeper concern, then we need to 
escalate the issue to a Java monkey that understands the interactions 
better than myself.

I will test the code by changing the wieldcategory to a higher value.

Cheers,

LegacyKing:
Your comment changed from what I got in the email... Quick test, Changing the HANDS to 5 in the gamemode made the Phantom slots unusable, dropping it to 1 opened up the primary slot. As far as I can tell, this is working as intended. 

My only question then becomes, do I need to dive deeper into the java code to see why it is working?

jdempsey-au:
Thanks for checking on that Andrew - sounds like this fix is a definite improvement.

PR #691, opened 2015-11-05
Updated test

LegacyKing:
Updated the second test in the same file to pass release. This is part of the 6.05.05 release I'm uploading, but going through the proper channels here...


LegacyKing:
@jdempsey-au this is a simple change to a test. 2 becomes 1 to reflect the changes.

PR #711, opened 2015-11-22
Data 2625 code

LegacyKing:
@jdempsey-au this is both a re-implementation and update to naming standards for ease of use of the 3e unit tests. And since you asked me to add license headers to MSRD tests, I've added those to make these compliant as well.


code/src/archivedtests/pcgen/inttest/game_3e/pcGenGUIBrdJoeTest.java
```diff
@@ -1,44 +0,0 @@
-package pcgen.inttest.game_3e;
```
jdempsey-au: This would be better as a move rather than as a delete and recreate. That way the history is preserved.

code/src/test/pcgen/inttest/game_3e/pcGenGUIDrakelingSorcTest.java
```diff
@@ -46,7 +68,7 @@ public static Test suite()
 	public void testDrakelingSorc() throws Exception
 	{
 		// Commented out as it fails due to removal of dependant sources.
-		//runTest("DrakelingSorc", "3e");
+		//runTest("3e_DrakelingSorc", "3e");
```
jdempsey-au: I think this would be better left in the archivedtests folder

LegacyKing:
Closing this out and redoing the PR properly.

PR #724, opened 2015-11-24
Code 3e tests

LegacyKing:


LegacyKing:
I'm scratching my head wondering why this is working only for some files and not all of them. I used the move function in tortoisegit to achieve this. I renamed them after a commit... in theory the history should have been preserved.

LegacyKing:
Here is my log screenshot:
image

and the second commit:
image

According to the log these are renames, not delete and add. Github appears to be logging the PR incorrectly in this case.

LegacyKing:
@jdempsey-au just pending your approval.

PR #733, opened 2015-11-29
CODE-2929 CONVERTER 6.06: PREALIGN silently stripped in conversion!

LegacyKing:


jdempsey-au:
Cool, that's certainly part of the issue, but we'll need further work to
stop the PREALIGNs being written back out without the alignments.

Cheers,
James Dempsey

LegacyKing:
Apparently our build cannot find the maven jep library... odd, mine passed.

LegacyKing:
Should I remove the non-parse code you mentioned? Or is there a better method?

LegacyKing:
So this:

if (Globals.getContext().getReferenceContext().getConstructedObjectCount(PCAlignment.class) == 0)
{
	// There are no alignments for this game mode, so we
	// do not do prereqs.
	prereq.setKey("");
	prereq.setOperator(PrerequisiteOperator.EQ);
}

just comment out or remove then?

jdempsey-au:
No, I wouldn't mess with that. Its a question for @thpr now I think

Cheers,
James

LegacyKing:
Good, cause I'm trying to make heads and tails of the file, and I'm confused. Are we still supporting the number version of PREALIGN? (Is that where the formula is factoring?)

I'll re-assign the tracker to @thpr then.

Cheers,

jdempsey-au:
This code fix is an improvement to current, so I'll merge it.

PR #744, opened 2015-12-08
exclude the filename readme.md, re-include homebrew due to complaints

LegacyKing:
@jdempsey-au update the exclusion list - removed homebrew and add readme.md


jdempsey-au:
Looks good.

PR #745, opened 2015-12-08
6.05.06 release

LegacyKing:
@jdempsey-au an update cleans up missing homebrew publisher entry. Re-running test. So far so good.


jdempsey-au:
Of course, it was needed in both locations.

PR #761, opened 2015-12-15
Fix bug:  CODE-2929 - CONVERTER 6.06: PREALIGN silently stripped in c…

jdempsey-au:
…onversion!


jdempsey-au:
Travis run on the same code in my repo passed so I will still merge this despite the failure.

PR #766, opened 2015-12-17
Allow specifying config.ini directory

neonobjclash:
I made this change so pcgen would work with a system-wide installation on linux. See [here](https://aur.archlinux.org/packages/pcgen-git/).


LegacyKing:
@jdempsey-au or @thpr please review

jdempsey-au:
Looks good, thanks

PR #770, opened 2015-12-17
Pcgio handler

LegacyKing:
Update to requested items.


LegacyKing:
@jdempsey-au or @thpr please review

LegacyKing:
@jdempsey-au we able to get this in?

LegacyKing:
@jdempsey-au we have two updated PRs, will we be adding these or are we waiting for the new development cycle? Looking to begin the release process in six hours.

jdempsey-au:
As these are not bug fixes they should go into the next dev cycle. I'm looking forward to having them merged though! It is always good to make good use of the work people have put in.

PR #774, opened 2015-12-19
Revert "6.06 release"

LegacyKing:
Reverts PCGen/pcgen#773


LegacyKing:
puts us back into the three failed tests. Only other option is revert all of the 3.5e gamemode related changes.

PR #799, opened 2015-12-25
Summoner spells

LegacyKing:


LegacyKing:
@jdempsey-au in regards to CODE-2957 SPELLLEVEL only evaluates changes on a reload. This version uses SPELLLEVEL on a Class level line. Grab this from my branch and test by creating a level 1 summoner, charisma 13. Note the Spell HASTE (Level 2). When you choose to be a Summoner Unchained, Haste SHOULD be Level 3. Level up after selecting Unchained Summoner. Check the spells tab, no change. Save and re-open. Haste is now level 3 and not level 2.

My other version was doing SPELLLEVEL on ability, which really illustrates the lack of refresh, no spells period until you reload the character. This version at least populates the spell list since the class is loading the spells. Not sure if this has the same pathways, but it's a starting example.

PR #826, opened 2016-01-03
Various UI Fixes

cpmeister:
Fix for FOP awt dialog poping up during printing.
Fix for improper tree rendering in tree tables
Fix for glitched behavior for remove language buttons in the language summary table.

Minor cleanup in build.gradle


cpmeister:
Should this be merged into the 6.x branch as well?

LegacyKing:
James should make that determination for whether this should be in 6.06.x

Good to see you back in action.

Any chance you could open a jira for the issues you fixed? Gets included in the release notes.

Cheers,
Andrew

On January 3, 2016, at 1:12 AM, cpmeister <notifications@github.com> wrote:

Should this be merged into the 6.x branch as well?

build.gradle
```diff
@@ -114,20 +117,13 @@ task copyMasterSheets(type: Copy) {
...
-    into ('../../horror/htmlxml') {
-	    from 'outputsheets/d20/fantasy/htmlxml'
-    }
-    into ('../../sciencefiction/htmlxml') {
-	    from 'outputsheets/d20/fantasy/htmlxml'
-    }
-    into ('../../western/htmlxml') {
-	    from 'outputsheets/d20/fantasy/htmlxml'
-    }
+    into 'outputsheets/d20/horror/htmlxml'
```
jdempsey-au: Have these changes been tested? Does the copymastersheets still refresh the sheets in these folders?

build.gradle
```diff
@@ -218,19 +214,15 @@ task copyToLibs(type: Copy) {
 }
 
 task copyToRoot(type: Copy, dependsOn: [copyToLibs, jar, converterJar]) {
-	outputs.files.setFrom(file("$buildDir/libs/pcgen-${version}.jar"))
```
jdempsey-au: Why remove this line?

jdempsey-au:
The code bugs would be fine to push back into 6.6, but they should be registered in [jira](https://pcgenorg.atlassian.net) please. I'd prefer not to push changes to the gradle build back to 6.6 but just leave those to master.

build.gradle
cpmeister: the pcgen jar is already included in the 'from' clause in the following line so there was no need to include the file twice. Also, if you are just trying to make sure that jar is part of the task output that happens automatically for you. Or is there perhaps a more nuanced reason for this line to be there that I didn't catch?

build.gradle
cpmeister: Yes, I've tested this. It does exactly as you would expect.

jdempsey-au: This was one of those areas I had trouble with when initially implementing the build and had to use some unusual workarounds. 

I've just tested these changes out using the gradle wrapper (v2.2) and copymastersheets doesn't copy the files across. It may well work in later versions of gradle, but I get a failure in distTar when I try the build in gradle 2.6 or 2.10.

build.gradle
jdempsey-au: I've tried removing this line and I see it fail in 2.2 and 2.10. Again I think it is one of those unusual workarounds I needed to do to get it working correctly.

jdempsey-au: This change causes the copyToRoot task to fail in gradle 2.2 and 2.10. Again it is one of those odd workarounds I needed to do.

build.gradle
cpmeister: Alright, I'll rip out the build.gradle changes and put the bug fixes in another pull request

jdempsey-au:
PR has now been superseded

PR #827, opened 2016-01-03
Fix bug: CODE-2963 - Converter output of CLASSES:TYPE=Psionic=1 is …

jdempsey-au:
…not accepted as valid syntax.


jdempsey-au:
@LegacyKing If you are happy with this syntax revision and the checks pass please feel free to merge it in.

PR #830, opened 2016-01-05
Partial Fix: CODE-2965 - Print preview does not accurately reflect printe…

cpmeister:
…d page

Enhanced performance of print preview zooming and page selection.
Grab and drag functionality is now available on previewed pages.


LegacyKing:
Quick Glance looks good. Passes the travis tests. I'll leave this for other code monkeys to chime in for the moment, since I have seven data-ish bugs to investigate and/or fix.

@cpmeister  Just for giggles, what do I do to reproduce the issue this patch fixes so I can give it a test spin?

@thpr you able to review this?

cpmeister:
Well, in the old print screen you could zoom in on the bottom and you could see that the pdf footer was getting cut off. That has been fixed but it has revealed another vexing issue. The pdfs that are exported from the character are not in the A4 format, even though it is my default, which gives a bit more insight into CODE-411 and CODE-2537.

LegacyKing:
@cpmeister  feel free to fix additional vexing issues. I think the next release will be delayed so we can get the errors getting reported resolved.

cpmeister:
I can keep committing fixes and various other code changes, but I don't want to get too ahead of myself because other code monkeys might start to disagree with some of it. Would you prefer giant pull requests or more bite sized pull requests with regards to issue fixes?

LegacyKing:
I would suggest a branch per code issue so any disagreements won't hold up the others patches. (Otherwise, each PR will just reflect a cascade of previous issues and defeats the purpose entirely)

Our best practice is one issue to a PR. Of course there are exceptions to that, for example James and I will break that rule from time to time, but we do adhere to one commit per issue so any problems can be dealt with quickly.

cpmeister:
Works for me.

code/src/java/pcgen/system/BatchExporter.java
```diff
@@ -471,6 +472,16 @@ public boolean accept(File aFile, String aString)
 		});
 	}
 
+	public static void printToXmlStream(CharacterFacade character, OutputStream outputStream)
```
jdempsey-au: It would be great to have some javadoc for this method

code/src/java/pcgen/util/fop/FopTask.java
```diff
@@ -0,0 +1,310 @@
...
+	{
+		this.inputSource = inputXml;
+		this.xsltSource = xsltSource;
+		this.renderer = renderer;
+		this.outputStream = outputStream;
+	}
+
+	public static FopTask newFopTask(InputStream inputXml, File xsltFile, OutputStream outputPdf)
+	{
+		return new FopTask(new StreamSource(inputXml), new StreamSource(xsltFile), null, outputPdf);
```
jdempsey-au: Should this be checking xsltFile like the ones below?

code/src/java/pcgen/system/BatchExporter.java
cpmeister: I actually have a bunch of javadoc changes in the commit "Deleted FOPHandler in preference to FopTask" but I rolled it back because I wanted to check that commit in a separate pull request.

code/src/java/pcgen/util/fop/FopTask.java
cpmeister: xsltFile can be null since it would mean that the identity transformer should be used.

code/src/java/pcgen/system/BatchExporter.java
jdempsey-au: Sounds good.

PR #840, opened 2016-01-08
Patch Unit Test "Jango". Appears the load error fixes commit fixed so…

LegacyKing:
…mething in Jango's code. I see it is now granting Jango a full complement which was not present before. "Test63b (one, one, one, one", and "Test63c (2x)".

Some odd changes, pushing this up for a full review

### DIFF
--- a/code/testsuite/csheets/35e_Jango.xml
+++ b/code/testsuite/csheets/35e_Jango.xml
@@ -1013,16 +1013,6 @@ BIO
 	<feats>
 		<!-- Visible standard feats (not including the auto feats) -->
 		<feat>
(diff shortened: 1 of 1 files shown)


code/testsuite/csheets/35e_Jango.xml
```diff
@@ -1196,7 +1216,7 @@ BIO
 			<chamod>0</chamod>
 			<cr>0</cr>
 			<dr></dr>
-			<feat>Test63a(One)</feat>
+			<feat></feat>
```
LegacyKing: @jdempsey-au I don't understand this section, not crucial but wanted to make sure this was expected.

PR #845, opened 2016-01-09
DATA-2695 Rogue Archetypes not removing Uncanny Dodge 

oiml:
Also fixed Improved Uncanny Dodge levels not stacking


data/pathfinder/paizo/roleplaying_game/core_rulebook/cr_abilities_class.lst
```diff
@@ -83,7 +83,7 @@ Fighter			CATEGORY:Class		DEFINE:Fighter_CFP_Level|0														ABILITY:Intern
...
+Rogue				CATEGORY:Class		DEFINE:Rogue_CFP_Level|0	DEFINE:Pool_Rogue_Class_Selection|0								ABILITY:Internal|AUTOMATIC|Class Skills ~ Rogue|PREVAREQ:Rogue_CF_ClassSkills,0	ABILITY:Class|AUTOMATIC|Rogue ~ Standard Class|!PREABILITY:1,CATEGORY=CLASS,TYPE.AltRogueChoice	BONUS:VAR|Rogue_CFP_Level|RogueLVL|TYPE=Base	BONUS:VAR|RogueSneakAttackLVL|RogueLVL	BONUS:VAR|TrapfindingLVL|RogueLVL	BONUS:VAR|RogueTalentLVL|RogueLVL|PREVARGTEQ:RogueLVL,2	BONUS:VAR|RogueTrapSenseLVL|RogueLVL|PREVARGTEQ:RogueLVL,3	BONUS:VAR|MasterStrikeLVL|RogueLVL|PREVARGTEQ:RogueLVL,20
 Sorcerer			CATEGORY:Class		DEFINE:Sorcerer_CFP_Level|0														ABILITY:Internal|AUTOMATIC|Class Skills ~ Sorcerer|PREVAREQ:Sorcerer_CF_ClassSkills,0																BONUS:VAR|Sorcerer_CFP_Level|SorcererLVL|TYPE=Base
```
LegacyKing: If referencing a class feature like this, the naming standard is "Class ~ Class Feature Name". Each class should have its own class features, and then the shared tracker ability can be granted directly from that ability. Make sense?

LegacyKing:
I might have a redesign to do on this. Each class should have a class feature call, however, with archetypes replacing and delaying things this becomes a very tangled ball of string

I won't be able to give this a proper review until much later tonight.

Can you quickly run down what the exact bug is on this?

data/pathfinder/paizo/roleplaying_game/core_rulebook/cr_abilities_class.lst
oiml: The "Uncanny Dodge Tracker" approach is currently being used with the Barbarian, I more or less just copied that. There are class abilities that are named "Barbarian ~ Uncanny Dodge" and "Barbarian ~ Improved Uncanny Dodge" (same for rogue), but they are never granted to the classes. The two CRB prestige classes that grant uncanny dodge do the same - they just increase the UncannyDodgeLevel variable by 1, which grants the generic Uncanny Dodge. Increasing it by 1 again grants Improved Uncanny Dodge. 
The rogue had the problem that one line just granted uncanny dodge, no check for archetypes whatsoever, so it could never be replaced again.
The not stacking bug came from the fact that the stacking counter was put on the unreferenced class features and never got used. 
The Uncanny Dodges are generally kind of messy, I could certainly do a complete cleanup with properly named class abilities (this would include changing the currently existing and working code for the Barbarian (plus unchained), the Assassin, the Rogue (plus unchained) and another few classes that I don't recall right now.

LegacyKing:
I will point out this is why each feature should be in its own branch. Avoids the multi-issue one thing holds up the rest.

PR #853, opened 2016-01-10
BatchExporter cleanup.

cpmeister:
Removed use of temporary files for pdf exports.
Deleted FopHandler and other related classes and instead use the more straight-forward FopTask.
Changed methods to use Java-7 try-with-resources functionality to handle stream cleanup after export.


cpmeister:
The aspect of this that *might* cause problems is the fact that temporary files are no longer used for pdf export. I think in earlier versions of fop this caused might have caused problems, but with this newer version of fop I haven't noticed any such issues with my (regrettably) limited testing.

LegacyKing:
@jdempsey-au will need to review and let me know if this is destined for 6.06. However, what is the reasoning behind removing the temp files? I like being able to review those xml files to see what values are being processed. If they no longer exist solving some problems may be more difficult. Unless I'm mistaking the temp character files for something else here.

cpmeister:
The temp files in question would be of an xml output sheet that is then fed through the xslt files to generate the pdfs.

LegacyKing:
Yeah, those are necessary for debugging issues - James and myself have used those many times to solve problems. Isn't there an option to save those in the UI?

Though dumping them in an easier to find location would be nice. However, that would be a feature request.

cpmeister:
The reason I removed the temp files was to perhaps speed up pdf generation of slower systems by generating the file into memory than onto disk. Then again the performance gain from this might be negligible so I'm not against removing it entirely to get back debugging features. But you are right, from the looks of it there wouldn't be another way to reasonably view those intermediate generated xml files.

cpmeister:
Alright, I'll make it so the generation of that xml file can be configured in the pcgen preferences.

LegacyKing:
@cpmeister make sure the feature is distinctly in its own branch. I don't think James will want features going to 6.06.x we're focusing on just bug fixing.

jdempsey-au:
I think this change should stay in master as it is not a direct bug fix and
we are in final stabilisation stage on the 6.6 branch.

Cheers,
James Dempsey

cpmeister:
Fine with me

PR #856, opened 2016-01-11
CODE-1353 - Add Zoom Controls for Thumbnail viewer for those without …

cpmeister:
…Scroller/Zoom Wheel enabled mice


LegacyKing:
Neat!

PR #857, opened 2016-01-11
DATA-2699 Expert Archer (Fighter Archer Archetype) not granting the correct bonus

oiml:
Crossbow Expert also had the wrong formula


LegacyKing:
@Distant-Scholar review this please. :)

PR #858, opened 2016-01-12
DATA-2700 Bloodrager Primalist Archetype replaces wrong Bloodline Powers

oiml:


LegacyKing:
@Distant-Scholar so this one is interesting, more files, and a false positive on Travis. My SOP is to re-run travis to see if that clears up the issue. Take a look at this as well. Does it look good?

LegacyKing:
@Distant-Scholar also, bonus points if you catch the mistake in our coding as well. ;)

PR #859, opened 2016-01-12
Bug Fix: CODE-2969 - Character sheet doesn't refresh

cpmeister:
CODE-2939 - Character Sheet page does not update when change character


LegacyKing:
@jdempsey-au this good for 6.06?

jdempsey-au:
This one would be good to have in 6.6

PR #872, opened 2016-01-14
DATA-2708 Arcanist Familiar Exploit not granting a familiar

oiml:
Tested stacking with Rogue Familiar Talent, Sorcerer Arcane Bloodline
and Wizard, all worked fine.


LegacyKing:
Passes, clear for back port to 6.6

@oiml this is above and beyond the call of duty, but can you do me a favor, all these BENEFIT tags need to be changed to "DESC: " yes, including the added space at the beginning. We do not support BENEFIT outside of feats. So this text does not appear anywhere. This can be a separate branch. Thanks,

oiml:
Sure, I can do that.

PR #878, opened 2016-01-16
BENEFIT: to DESC: in ACG

oiml:
Excluding Warpriest Bonus Feats.


LegacyKing:
Why excluding Warprist bonus feats?

oiml:
The warpriest bonus feats are counted as feats, so the BENEFIT part does show up on the outputsheet. I believe changing this to DESC: would remove the line break, which might not be wanted.

LegacyKing:
If you look, the "Bonus Feats" in the file only attach the real feat to the PC. They are essentially placeholders. You can tell what an actual Feat is by CATEGORY:Feat. Those bonus feat benefits aren't being displayed in the UI.

oiml:
Yeah, you're right. I also noticed that at least one class ability output looks weird now (bloodrager claws). I'll close this for now and reopen it when I'm done (probably tonight).

oiml:
All BENEFIT: tags were converted and I hope that I got all of the double text entries.

PR #885, opened 2016-01-19
DATA-2723 Bloodrager BloodlineLVL not being calculated properly

oiml:


LegacyKing:
As a maintainer, it helps if the issue being addressed accurately reflects the described problem. As a developer monkey, if you see the problem, you should EDIT the issue to more accurately describe the problem. Remember, we're the "experts" and deal with the underlying cause, the end users only describe the symptoms.

In this case, I'm scratching my head wondering how adding new variables corrects a CALCULATION error. "Bloodrager BloodlineLVL not being calculated properly". That indicates a var isn't formatted correctly. Instead, this should be named "Bloodline Powers are granted at first level, instead of the correct levels". 

Hope that helps.

oiml:
Heh, yeah. I actually didn't create this bug report and just copied the title without paying too much attention to it. I'll watch out for this in the future!

PR #905, opened 2016-01-24
DATA-2734 and DATA-2735, Dragon Disciple not granting bonus feats/granting bonus spells too early

oiml:
I just realized this when I was done, but should I do this in two separate pull requests in the future?


data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_abilities_class.lst
```diff
@@ -311,7 +316,7 @@ Bloodrager Bloodline ~ Abyssal Bonus Spells	CATEGORY:Special Ability	TYPE:Intern
...
-Bloodrager Bloodline ~ Draconic Bonus Spells	CATEGORY:Special Ability	TYPE:Internal	VISIBLE:NO	SPELLKNOWN:CLASS|Bloodrager=1|Shield|PREVARGTEQ:BloodlineCasterLVL,7	SPELLKNOWN:CLASS|Bloodrager=2|Resist Energy|PREVARGTEQ:BloodlineCasterLVL,10	SPELLKNOWN:CLASS|Bloodrager=3|Fly|PREVARGTEQ:BloodlineCasterLVL,13	SPELLKNOWN:CLASS|Bloodrager=4|Fear|PREVARGTEQ:BloodlineCasterLVL,16
+Bloodrager Bloodline ~ Draconic Bonus Spells	CATEGORY:Special Ability	TYPE:Internal	VISIBLE:NO	SPELLKNOWN:CLASS|Bloodrager=1|Shield|PREVARGTEQ:BloodlineCasterLVL,7|PRESPELLTYPE:1,Arcane=1	SPELLKNOWN:CLASS|Bloodrager=2|Resist Energy|PREVARGTEQ:BloodlineCasterLVL,10|PRESPELLTYPE:1,Arcane=2	SPELLKNOWN:CLASS|Bloodrager=3|Fly|PREVARGTEQ:BloodlineCasterLVL,13|PRESPELLTYPE:1,Arcane=3	SPELLKNOWN:CLASS|Bloodrager=4|Fear|PREVARGTEQ:BloodlineCasterLVL,16|PRESPELLTYPE:1,Arcane=4
```
LegacyKing: Why the appended PRExxx?

LegacyKing:
Okay, walk me through these changes, cause I'm not following this easily in the diff. I see !PRECLASS for a class that doesn't exist in CR. This should toss an error if I load just the core rule book.

data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_abilities_class.lst
```diff
@@ -370,7 +375,7 @@ Draconic Resistance		KEY:Draconic Resistance ~ Acid				CATEGORY:Special Ability
...
...TEGORY=Special Ability,Bloodrager Bloodline ~ Draconic (Green)	ABILITY:Special Ability|AUTOMATIC|Breath Weapon ~ Acid Line|PREABILITY:1,CATEGORY=Special Ability,Bloodrager Bloodline ~ Draconic (Black),Bloodrager Bloodline ~ Draconic (Copper)	ABILITY:Special Ability|AUTOMATIC|Breath Weapon ~ Cold|PREABILITY:1,CATEGORY=Special Ability,Bloodrager Bloodline ~ Draconic (White),Bloodrager Bloodline ~ Draconic (Silver)	ABILITY:Special Ability|AUTOMATIC|Breath Weapon ~ Electricity|PREABILITY:1,CATEGORY=Special Ability,Bloodrager Bloodline ~ Draconic (Blue),Bloodrager Bloodline ~ Draconic (Bronze)	ABILITY:Special Ability|AUTOMATIC|Breath Weapon ~ Fire Cone|PREABILITY:1,CATEGORY=Special Ability,Bloodrager Bloodline ~ Draconic (Red),Bloodrager Bloodline ~ Draconic (Gold)	ABILITY:Special Ability|AUTOMATIC|Breath Weapon ~ Fire Line|PREABILITY:1,CATEGORY=Special Ability,Bloodrager Bloodline ~ Draconic (Brass)																																		BONUS:VAR|BreathWeaponTimes|1|PREVARGTEQ:BloodlineLVL,8	BONUS:VAR|BreathWeaponTimes|1|PREVARGTEQ:BloodlineLVL,16	BONUS:VAR|BreathWeaponTimes|1|PREVARGTEQ:BloodlineLVL,20	BONUS:VAR|BreathWeaponDice|BloodlineLVL	BONUS:VAR|BreathWeaponDC|10+(BloodlineLVL/2)+CON
```
LegacyKing: What was wrong with this bonus statement?
max(0,(BloodlineLVL-12)/4)+1


Which leads me to the next head scratcher, why are we making three different statement bonuses instead of combining into a formula?

BONUS:VAR|BreathWeaponTimes|1|PREVARGTEQ:BloodlineLVL,8	BONUS:VAR|BreathWeaponTimes|1|PREVARGTEQ:BloodlineLVL,16	BONUS:VAR|BreathWeaponTimes|1|PREVARGTEQ:BloodlineLVL,20

Should be
BONUS:VAR|BreathWeaponTimes|1+(BloodlineLVL>=16)+(BloodlineLVL>=20)

I don't see why the first variable requires a PRExxx, the feature is added at 8th level correct? If you can't use it at least once per day, then you shouldn't have access to it, or am I missing something?

LegacyKing: Ah, I see

max(0,(BloodlineLVL-12)/4)+1
vs.
max(0,(((BloodlineLVL-12)/4)+1)

data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_abilities_class.lst
oiml: I did this both for the sorcerer and the bloodrager. If you have a sorcerer5/dragon disciple6, you have only level 4 spell slots, but your bloodline is advanced enough to grant you the 5th level bonus spell, so it shows up on your character sheet even if you can't cast it. The Dragon Disciple class description even mentions this case. With this you only get the bonus spells added to your spells known list if you have a spellslot of the appropriate level.

data/pathfinder/paizo/roleplaying_game/advanced_class_guide/acg_abilities_class.lst
oiml: Dragon Disciple can give you access to it earlier than level 8, so if you get it at 7 the formula gives you an extra use. Before you had really weird behavior with this where you'd get the bloodrager breath weapon and the sorcerer breath weapon (and doubled up on the damage on both), but the sorcerer breath weapon had uninitialized variables and incomplete text descriptions. The PRECLASS in the core rulebook is one of the strangest things I encountered so far, because it doesn't give you any errors at all. Loading Core Rulebook only, gaining the ability that has the tag, everything works fine, I don't even get a warning (this might actually be a bug). I don't know if there is another way to block bloodragers from getting this without prohibiting it in the core rulebook, other than adding the ability again in the ACG with bloodrager in mind.

PR #920, opened 2016-01-30
CODE-2990: UI performance enhancements

cpmeister:
This really helps on the feats tab where there might be thousands of different abilities to choose from. Makes the tab switch pretty instantaneous. :)


jdempsey-au:
Looks safe to me

PR #942, opened 2016-02-06
Bug Fix - CODE-2993: Thumbnail exports poorly to pdf

cpmeister:


LegacyKing:
Does the input matter if it's a jpg or png or anything else? I see we now export to png. Just curious.

cpmeister:
It doesn't really matter. But JPEG has limitations when it comes to transparency, an issue that png doesn't have. The images may look slightly different (due to them using different types of compression) but the difference should not be very noticeable or even matter that much to the end user.

cpmeister:
As far as the input types, we still support jpeg during thumbnail selection we just change it to png during export.

LegacyKing:
Sounds good. The change is trivial. I'll give it a good test. Thanks!

PR #955, opened 2016-02-14
Fix bug: OS-440 PDF Export and Spellbook display

LegacyKing:


LegacyKing:
@Zaister please give this a review. I do not believe the IF tests were required for correct output.

PR #958, opened 2016-02-14
corrected skill name in underlying principles trait

oiml:


LegacyKing:
Normally this should be attached to a jira for proper release note updates.

PR #964, opened 2016-02-16
NEWTAG-530-B

thpr:
Rebuild source output in Freemarker


thpr:
Neither is preferred, and I don't think effort to swap between the two is time well spent.  We should be transitioning to using actual FreeMarker items not the legacy export tokens.

${deity.alignment}

If we want to address other issues like file consistency, then open a separate FREQ.  Let's keep this to addressing issues around source change.

LegacyKing:
Ah, then transitioning to the new format away from legacy is what we will do. Was there something you needed from data for this pr?

thpr:
Any output sheet using pc.source in the old style will break... so not data, but the output sheet folks need to look at the current OS.

PR #986, opened 2016-02-22
DATA-2771: Inner Sea Gods list incomplete

Zaister:
The list was not just incomplete, it was completely missing.

I added some fluff as FACTSETs, such as the deities' symbols, sacred animals and sacred colors as listed the book. However I am getting this warning on load:

SEVERE Thread-3 FactSetDefinition:75 Deity output symbol already exists, ignoring Visibility to EXPORT for FACTSET: Symbol

Maybe because SYMBOL used to be its own tag? I though that was deprecated in favor if using FACTs?


data/pathfinder/paizo/campaign_setting/inner_sea_gods/isg__datacontrols.lst
```diff
@@ -2,3 +2,11 @@
 
 # Name			Dataformat		Visible	Explanation
 FACTSETDEF:SPELL|Deity	DATAFORMAT:String	VISIBLE:YES	EXPLANATION:Just for Listing deities that are predisposed to certain spells.
+#
+FACTSETDEF:DEITY|Symbol	DATAFORMAT:String	VISIBLE:YES	EXPLANATION:The deity's holy symbol
```
LegacyKing: Uh, Symbol is already made. And it is FACTDEF:Symbol. Hence your error. You cannot have a FACT and FACTSET share the same namespace. A Deity only has one symbol. If the item will be single it should be a FACT, if you can have multiple, then it is a Set of Facts FACTSET.

Cheers,

PR #996, opened 2016-02-27
NEWTAG-582 CRITMULT code control

thpr:


code/src/java/pcgen/core/Equipment.java
```diff
@@ -5510,6 +5470,7 @@ public int getCritMultiplier()
 	 * Gets the altCritMultiplier attribute of the Equipment object
 	 * 
 	 * @return The altCritMultiplier value
+	 * @deprecated due to CRITMULT code control
 	 */
 	public int getAltCritMultiplier()
```
cpmeister: If you are going to have @deprecation in the method javadoc, wouldn't you also want to add 
@Deprecated to the method as well?

thpr: the javadoc version does both in Eclipse (and should do so elsewhere), since the javadoc versio predates the ability to annotate code, so no need to do the annotation as well.

PR #1005, opened 2016-03-01
Tree tables now use the regular table headers instead of buttons

cpmeister:
CODE-3009: Use standard table header columns instead of buttons
CODE-3010: Use java 6 TableRowSorter for table sorting rather than a custom JTableHeader
CODE-3011: Make tree view selection dropdown more intuitive.


code/src/java/pcgen/gui2/coreview/CoreViewFrame.java
```diff
@@ -240,8 +238,8 @@ public void setPerspective(CorePerspective corePerspective)
 		@Override
 		public ListFacade<? extends TreeView<CoreViewNodeFacade>> getTreeViews()
 		{
-			DefaultListFacade<TreeView<CoreViewNodeFacade>> views =
-					new DefaultListFacade<TreeView<CoreViewNodeFacade>>();
+			DefaultListFacade<TreeView<CoreViewNodeFacade>> views
+					= new DefaultListFacade<TreeView<CoreViewNodeFacade>>();
```
thpr: This doesn't look like our formatting guidelines.  Are you using the settings in standards?

cpmeister: I use Netbeans for my pcgen work. I've adjusted its formatting settings to adhere as closely as possible the pcgen formatting standards. That said, I've looked through the Eclipse code format xml files and I cannot find (it is possible I just missed it) any settings regarding wrapping the line after the assignment operator. Luckily Netbeans has such a formatting option so I can keep to our existing format. But you do bring up a good point that there have been some formatting rules I've been breaking :)

Should this formatting rule be added to the formatting standard if it isn't already there?

Zaister:
It seems some dialogs may have lost their sorting. I noticed that on the chooser that pops up when selecting a weapon proficiency for the Weapon Focus feat.

Zaister:
The new class JTableMenuButton.java is wreaking havoc on my my repo. It's always showing up as modified and I can't get rid of that with "git checkout --". I think it's the old line ending problem, as I am getting a message about line endings in a git diff. Can you fix that, please?

LegacyKing:
+1

Ditto, my last commit is also showing this file as modified.

cpmeister:
Well shoot. I'll try to fix it asap.

cpmeister:
Alright, I just changed the line endings. Let me know if you still have problems.

PR #1009, opened 2016-03-02
Eliminate LoadableLegalScope

thpr:
Basically, this wasn't useful - we need to check the scope names are the same, as it may not be the same object class (think SubClass) and also needs to be able to handle non-CDOMObject (think Dynamic)


cpmeister:
Isn't this going to conflict with the PR for NEWTAG-580?

PR #1023, opened 2016-03-05
DATA-2789 Bracers of Falcon's Aim don't improve crit range

oiml:


LegacyKing:
was this tested to see if it worked?

oiml:
I did test this with the 6.06 version and the 6.07 autobuild, worked just fine in both versions.

PR #1028, opened 2016-03-08
Feature: DATA-2683 There is no "Young" Age set

LegacyKing:
* Remove Default Biosetting - it interferes with any pcc bioset that does not match exactly
* Place bioset in one file to avoid issues
* Add Young category
* Implement Young Stat changes


LegacyKing:
@cpmeister Travis is failing, but personal tests and these errors do not exist. Anyways, this the request you made.

PR #1071, opened 2016-03-21
NEWTAG-588 LEGS CodeControl

thpr:


code/src/java/pcgen/core/display/CharacterDisplay.java
```diff
@@ -1627,7 +1627,7 @@ public boolean containsRacialSubType(RaceSubType st)
 	 * 
 	 * @return The number of legs.
 	 */
-	public int getLegs()
+	public int getPreFormulaLegs()
```
cpmeister: why did this change?

thpr: It was changed to precisely describe what it is doing.  It is only accurate "pre formula"  - meaning without the LEGS code control enabled.  This is because it uses the facets.  It is therefore for use only by things that actually can operate in the "old mode" - meaning things like PRELEGS or a LEGS term from the old formula system.  We intentionally do not want the "new legs" to pass through to those items, to prevent mixing of the old and new systems.

PR #1080, opened 2016-03-22
NEWTAG-592 ACCHECK Code Controls

thpr:


code/src/java/pcgen/core/Equipment.java
```diff
@@ -2349,19 +2349,6 @@ public String getWieldName()
...
-	 * Description of the Method
-	 * 
-	 * @param aPC The PC that has this Equipment
-	 * 
-	 * @return Description of the Return Value
-	 */
-	public Integer acCheck(final PlayerCharacter aPC)
-	{
-		return Math.min(getSafe(IntegerKey.AC_CHECK)
-			+ (int) bonusTo(aPC, "EQMARMOR", "ACCHECK", true), 0);
```
cpmeister: There seems to be a lot of code that used this. Why get rid of it?

thpr: I'm operating with a bias to delete methods from Equipment and PlayerCharacter when possible to simplify those interfaces, and prevent usage of the old methods elsewhere... force as much to the (old, to be deprecated) output system and terms as possible

cpmeister: IMO this method is sophisticated enough to warrant being refactored the way it was. So having it "unrefactored" seems kind of backwards to me from a code management standpoint. I don't mind it being removed from Equipment but at the very least I think it should be moved to a utility class.

PR #1081, opened 2016-03-22
Fpr/pf/size

LegacyKing:
@thpr not sure if the array approach will work. It seems the freemarker template is doing regular slice with the bracket approach. I need to do ${pc.val.face[0]}${pc.val.face[1]} to display a space/face of 15.

Anyways, if the hashmap change will handle freemarker, then this will be ready to be incorporated in.

Only thing left to fix is the unit tests.


LegacyKing:
Cleaned up single commit version PR'd. Closing this one.

PR #1088, opened 2016-03-25
Eliminate x.class version of getFormatManager

thpr:


thpr:
This is preparation for some more advanced formats that will mean all formats should be referenced by name

PR #1092, opened 2016-03-26
NEWTAG-576-B

thpr:
Armor (AC) Code Controls


cpmeister:
I wonder if at some point we should starting making static fields out of all the string constants that are used like "AC" and such.

thpr:
Probably.  Let's merge this and then I'll create a PR to extract the code controls

PR #1102, opened 2016-03-27
Code control constants

thpr:


cpmeister:
Some of the constants don't have the '*' character that strings they replace have. Is this a problem or intentional?

thpr:
Intentional - some were missing it (bug fixes) the rest now have the "*" inserted by the method in ControlUtilities or PlayerCharacter

PR #1103, opened 2016-03-27
Introduce Variable Channels

thpr:
Bring up to Formula Lib v59


cpmeister:
PCGen-base was updated in the formula project, is this new PCGen-fomula version going to use this new PCGen-base or is it ok with the existing one?

LegacyKing:
Failing tests here.

thpr:
Connor, Good catch.  Updated (and tests fixed)

PR #1104, opened 2016-03-28
Size Implementation using the New Formula Parser

LegacyKing:


code/testsuite/base-xml.ftl
```diff
@@ -102,7 +102,7 @@
 		</skin>
 		<cr><@pcstring tag="CR"/></cr>
 		<face>
-		<face><@pcstring tag="FACE"/></face>
+			<face><#if pc.val.os_size!false == true><#if pc.val.face[0] == pc.val.face[1]>${pc.val.face[0]} ft.<#else>${pc.val.face[0]} ft. by test ${pc.val.face[1]} ft.</#if><#else>${pcstring('FACE')}</#if></face>
```
thpr: It strikes me that instead of forcing you to have pc.val.os_size, which really replicates "do I have the FACE code control?" we should give you real access to the code control... wouldn't that make more sense (and save you a variable?)

LegacyKing: Yes, that would help. Especially since I have discovered freemarker is extremely strict about values being absent.

data/pathfinder/alluria_publishing/remarkable_races/rr_races.lst
```diff
@@ -25,7 +25,7 @@ Zif											STARTFEATS:1	SIZE:M	MOVE:Walk,20																LANGBONUS:TYPE=Spo
...
+Rift Swarm										STARTFEATS:1			MOVE:Walk,20				MODIFY:Face|SET|10,10|PRIORITY=200																																																	BONUS:VAR|HasLowlightVision|1																																						ABILITY:Internal|AUTOMATIC|Race Traits ~ Rift Swarm												NATURALATTACKS:Swarm,Weapon.Melee,*1,4d6,SPROP=plus Poison (Fort DC 26)															RACETYPE:Outsider				RACESUBTYPE:Swarm								TYPE:Outsider		CR:14			SOURCEPAGE:p.140
 Gumpawomp										STARTFEATS:1	SIZE:L	MOVE:Walk,20																															BONUS:COMBAT|AC|7|TYPE=NaturalArmor																				BONUS:VAR|RaceSizeIsLong|1																				BONUS:WEAPONPROF=Tail|REACH|5														ABILITY:Internal|AUTOMATIC|Race Traits ~ Gumpawomp								LEGS:4	HANDS:2	NATURALATTACKS:Tail,Weapon.Melee.Bludgeoning,*1,2d6																		RACETYPE:Animal				RACESUBTYPE:Gumpawomp							TYPE:Animal			CR:1			SOURCEPAGE:p.141
```
thpr: Why does this line (and some others in various places) still have an old "RaceSizeIsLong" variable?

LegacyKing: Missed them, should be converted to MODIFY:SizeIsLong|SET|True

data/pathfinder/alluria_publishing/remarkable_races/rr_races.lst
```diff
@@ -34,7 +34,7 @@ Purple Peeper									STARTFEATS:1	SIZE:T
...
-Coin Drake										STARTFEATS:1	SIZE:L	MOVE:Walk,30,Burrow,30,Fly,100	FACE:10	REACH:10																								BONUS:COMBAT|AC|20|TYPE=NaturalArmor																			BONUS:VAR|OverrideRaceReach,RaceSizeIsLong|1	BONUS:VAR|Maneuverability|2																														ABILITY:Internal|AUTOMATIC|Race Traits ~ Coin Drake																										SR:21					MONSTERCLASS:Dragon:13				RACETYPE:Dragon														TYPE:Dragon			CR:10			SOURCEPAGE:p.146									KIT:1|TN
+Coin Drake										STARTFEATS:1	SIZE:L	MOVE:Walk,30,Burrow,30,Fly,100																									BONUS:COMBAT|AC|20|TYPE=NaturalArmor																			BONUS:VAR|OverrideRaceReach,RaceSizeIsLong|1	BONUS:VAR|Maneuverability|2																														ABILITY:Internal|AUTOMATIC|Race Traits ~ Coin Drake																										SR:21					MONSTERCLASS:Dragon:13				RACETYPE:Dragon														TYPE:Dragon			CR:10			SOURCEPAGE:p.146									KIT:1|TN
```
thpr: Still "RaceSizeIsLong" in the file.  You should do a search in eclipse for the string in the data files.  While you're at it, look for "VAR|SizeIsLong" since there also seems to be a misnamed variable in the old system.

thpr:
This commit should also be rebased from the current master, since the history looks to be a branch from a very old master.  Merging as was done here is not the best way to keep current.

PR #1117, opened 2016-03-30
CODE-2994 Channel updates for Stat

thpr:
Creates the actual "compatibility" interface for the UI to be able to attach to for stats, regardless of whether the STATSCORE code control is enabled.  (Note it can't be enabled at all because it doesn't actually exist, but this will at least return the "old style" object so that the UI code can be converted to using this subsystem.


cpmeister:
Conflict

PR #1129, opened 2016-04-02
Linked UI into stat channels.

cpmeister:
There is a lot of cleanup work that still needs to be done, but this should be enough to get input support for stats.

### DIFF
--- a/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
+++ b/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
@@ -170,6 +170,7 @@ import pcgen.facade.util.DefaultReferenceFacade;
 import pcgen.facade.util.ListFacade;
 import pcgen.facade.util.ListFacades;
 import pcgen.facade.util.ReferenceFacade;
+import pcgen.facade.util.WriteableReferenceFacade;
 import pcgen.facade.util.event.ChangeListener;
 import pcgen.facade.util.event.ListEvent;
 import pcgen.facade.util.event.ListListener;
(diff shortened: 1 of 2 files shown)


cpmeister:
@thpr Can you take a look?

LegacyKing:
Looks interesting. While we wait for @thpr I assume this will compile? How would I grab the values?

input(Stat,"STR",Score)?

code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
```diff
@@ -373,7 +375,18 @@ private void initForCharacter()
...
+		statScoreMap = new HashMap<StatFacade, WriteableReferenceFacade<Integer>>();
+		for (StatFacade stat : dataSet.getStats())
+		{
+			if (stat instanceof PCStat)
+			{
+				statScoreMap.put(stat, StatAdapter.generate(theCharacter.getCharID(), (PCStat) stat));
+			}
+			else
+			{
+				statScoreMap.put(stat, new DefaultReferenceFacade<Integer>());
```
thpr: Tactically, I think the checkin is okay, but strategically, this is the one part that stands out to me.  My personal preference would be some form of message here since we should assert that StatFacade == PCStat.  If we don't make that assumption, then we really should be putting StatAdapter.generate into DataSetFacade so that it's hidden from CharacterFacadeImpl? ... or what's the appropriate place so that the assumption of StatFacade == PCStat is always in one place?

thpr:
@LegacyKing  More than one step to do here.   You will need a code control, and that will be explained when it's complete.

LegacyKing:
@thpr and @cpmeister and @Zaister -- I believe this has caused a Blocker Issue in the main development line. We now have endless cyclic granting on Racial bonuses.

thpr:
This is a merged issue, if there is a bug in the code please open a bug
report.

Thanks

Tom

Zaister:
The culprit seems to be the for block in code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
There is a jira issue open for this: CODE-3045.

PR #1143, opened 2016-04-08
fixed type of kitsune lowlightvision racial trait

guzompi:
See JIRA DATA-2862


LegacyKing:
Looks good. In the future please use the branch system, keeps separate issues out of the same PR, also makes the pr process smoother. Cheers,

PR #1172, opened 2016-04-25
Freq/code 3044a

LegacyKing:
@cpmeister or @thpr


LegacyKing:
Not sure what I'm missing. Loads without error, but Goldielocks still can't find the two spells. (Running a test instance branch with the spell changes and migration file updated)

LegacyKing:
Disregard last comment, setting the version correctly helps. Works for me. Just need @thpr or @cpmeister to approve and merge.

LegacyKing:
@cpmeister 
I think functionally this might be useful to have for KEY name changes in spells.

cpmeister:
Could you elaborate on where this would be used? I don't see how this PR interacts with the code base.

LegacyKing:
We have a file 'migration' that is supposed to assist in the changing of KEYs to a new assigned key. When a PC is loaded, any Old Key would normally generate a warning. With the migration in place, that warning is suppressed, and the new key inserted onto the character instead. It is my hope that the SPELL code (Which seems to be getting more complex with newer releases) might be beneficial. There have been some corner case naming irregularities and having this function to bring 3pp sources into current standards would be handy.

Whether the code is fully complete, I cannot tell. I was following the existing examples of Equipment and Abilities when I set this up. It is possible I missed a link somewhere.

cpmeister:
I think you need to add some code on line 4127 of pcgen.io.PCGVer2Parser. Look at line 3678 as an example for how it is done for race. Other than that it looks good.

LegacyKing:
Yeah, looks like a bit of trial and error to do that file. Might be a bit.

code/src/java/pcgen/io/migration/SpellMigration.java
```diff
@@ -0,0 +1,73 @@
...
+ * SpellMigration translates old spell keys into their current values. This is 
+ * used to allow clean loading of older characters which were saved with spell 
+ * keys that have now been changed in the data.
+ * 
+ * @author Andrew Maitland <drew0500@yahoo.com>
+ * @version $Revision$
+ */
+public class SpellMigration
+{
+	private static Map<int[], List<MigrationRule>> spellChangesForVer = new HashMap<int[], List<MigrationRule>>();
```
grimreaper: Map<> ?

code/src/java/pcgen/io/migration/SpellMigration.java
```diff
@@ -0,0 +1,73 @@
...
+	/**
+	 * Find the new spell key to replace the provided one.
+	 * 
+	 * @param spellKey The original spell key as found in the character file.
+	 * @param pcgVer The version of PCGen in which the character was created.
+	 * @return The new spell key, or the passed in one if it has not changed.
+	 */
+	public static String getNewSpellKey(String spellKey, int pcgVer[], String gameModeName)
+	{
+		List<MigrationRule> spellChangeList = spellChangesForVer.get(pcgVer);
```
grimreaper: `List<>` ?

LegacyKing:
Hi @grimreaper 

I know it is hard to believe but I am a complete novice when it comes to Java. Worse, the pcgen code base is quite large and I don't even have a complete grasp on it. This code is similar to that submitted and actively used today by @jdempsey-au , the former Code Silverback. If nothing else, the code is meant to be consistent with whatever already exists. I had java and C++ at the same time in my classes, so a little overlap course work. I don't normally submit java stuff, but this was a relatively minor change, and something I need to bring 3rd party sources in line with standards. 

I would be leery of altering specs without running it by @thpr or @cpmeister. If they agree it's a good change, then I'll be happy to alter it.

grimreaper:
@LegacyKing fwiw I did not intend for my comments to come off as requirements or being annoying. Sometimes code review can feel like that, particularly when terse like mine. Additionally, I have a (bad?) habit of picking up on issues that are pre-existing in the code bases during review and will sometimes comment on them just to bring awareness.

Overall this change is fine by me and I'd "accept" were I a reviewer.

LegacyKing:
I don't mind the review at all. If my reply came off as that, then I apologize. I merely wanted to explain the fact my java experience is very very limited, and the PCGen base is the most exposure to an existing program beyond my own classwork. Making your own limited scope program is far easier than tinkering on a multifaceted behemoth such as PCGen.

If your comments and feedback prompt the reviewers to take it into consideration then great. I know Tom and Connor would appreciate refactoring, clean up and anything to make PCGen easier to transition to Java 8. Apparently the java docs are a concern. However, toss me an email and we'll get you on the team hipchat, where we can chat at leisure without alerting everyone who watches these issues with notification emails. (help@pcgen.org will reach me).

grimreaper:
sent you an email. Future correspondence will be off-PR

PR #1173, opened 2016-04-25
CODE-2646: Make Spell MODs case insensitive

Zaister:


cpmeister:
Did you format that by hand or is there a tool that did that for you?
Also, did you notice any performance hit from switching from HashMap to TreeMap?

Zaister:
I formatted it by hand, as the lines looked strangely ragged. I have not seen any noticeable performance change

PR #1183, opened 2016-04-30
Implement: DATA-2898 Please implement "Fates Favored" to alter Luck b…

LegacyKing:
…onuses on character


LegacyKing:
@Distant-Scholar or @Zaister can you verify the correct bonuses are applied here.

Zaister:
I can't really say if you them all.

PR #1197, opened 2016-05-02
Bug fix CODE-3045: Racial bonus are reapplied on each level up (and o…

cpmeister:
…n other changes)

### DIFF
--- a/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
+++ b/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
@@ -1833,14 +1833,14 @@ public class CharacterFacadeImpl implements CharacterFacade, EquipmentListListen
 
 	private void refreshStatScores()
 	{
-		for (StatFacade stat : statScoreMap.keySet())
-		{
-			WriteableReferenceFacade<Integer> score = statScoreMap.get(stat);
-			if (stat instanceof PCStat)
-			{
-				score.set(theCharacter.getTotalStatFor((PCStat) stat));
-			}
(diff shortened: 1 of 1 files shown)


Zaister:
Connor, as far as I can see this fix doesn't work as intended. It removed the racial bonus being applied over and over again, but it also removes it being applied at in the first place. If a create a new character now, and then set the race, to elf, for example, that stat bonuses are ignored.

I played around earlier with the code and I believe this works: in code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java, remove/comment lines 379-389 (and undo this commit). Can you take a look at this again, please.

cpmeister:
Strange, when I test it locally and the racial bonus was indeed added only once. That said the racial bonus doesn't display immediately which might lead to some confusion. Can you give me the steps you took to replicate your issue?

Zaister:
I created a new character and then chose elf from the race dropdown, and the bonuses were not applied when I moved the focus away from the menu, as they used to be.

cpmeister:
Well there certainly appears to be a repainting issue for the stats. But just to confirm, you don't see the race stats display after clicking the ability scores table? That does the trick for me. Granted I can hardly call it a permanent solution, kinda revealed another issue though...
But I just want to make sure we are seeing the same result.
Commenting out 379-389 would remove the "input" token support for ability scores that Tom is working on. I would prefer to not undo that if I can.

Zaister:
You're right, the bonuses are applied, but only once I change the tab, but not when I just click anywhere on the summary tab. Not really a permanent solution indeed.

PR #1212, opened 2016-05-05
Bug fix data 2909 champions of balance

evilpixie87:
Inquisitor subdomains were added, but "FACT:Domain|(name of main domain)" gives errors. Since inquisitors get only one domain, I removed this tag. Also removed spells, since inquisitors don't get domain spells.


LegacyKing:
LST diff is not fun...

PR #1216, opened 2016-05-10
Update to pcgen.sh to fix Unix/DOS file issue

bradfonseca:
The issue here is with the way Windows/DOS systems create text files versus how Unix/Linux systems do it. Basically, "DOS uses carriage return and line feed ("\r\n") as a line ending, which Unix uses just line feed ("\n"). You need to be careful about transferring files between Windows machines and Unix machines to make sure the line endings are translated properly." (http://www.cs.toronto.edu/~krueger/csc209h/tut/line-endings.html).

There are a number of possible fixes, including a Perl program called dos2unix that can do the conversion. More information can be found at https://kb.iu.edu/d/acux and at the StackOverflow article I posted (http://stackoverflow.com/questions/2920416/configure-bin-shm-bad-interpreter).

The fact is that the Shell file is fine except that it was created or edited at some point in a DOS environment so it's picked up the carriage return ("\r") characters in the line endings.

### DIFF
--- a/pcgen.sh
+++ b/pcgen.sh
@@ -1,0 +1,98 @@
+#!/bin/sh
+cd `dirname $0`
+
+# java.awt.Desktop.browse should be available and setting BROWSER is not needed anymore
+if [ "x$BROWSER" = x ]
+then
+    case "$WINDOWMANAGER" in
+        *kde ) BROWSER=kde-open ;;
+        *gdm ) BROWSER=gnome-open ;;
+        * ) BROWSER=netscape ;;
+    esac
+fi
+
+available_memory="unknown"
+default_min_memory=256
+default_max_memory=512
+
+# Linux /proc/meminfo
+if [ -e "/proc/meminfo" ]; then
+	available_memory=$(grep MemFree: /proc/meminfo | awk '{ print $2; }')
+	echo "Available memory: $available_memory kB"
+
+# BSD (thus MacOSX) memory command line should be in /usr/bin/vm_stat
+elif [ -x /usr/bin/vm_stat ]; then
+	# Mach Virtual Memory Statistics: (page size of 4096 bytes)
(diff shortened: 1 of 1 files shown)


LegacyKing:
You're fast. I just sent you the instructions a minute after this notification appeared. I cannot see line endings in the file itself. I'll let @thpr take a look, but this looks good to me. Not sure why the entire file is showing as changed though...

bradfonseca:
I got fork and pull request to work through the web interface which is cheating but expedient. I uploaded my version of the file which is why the entire file is showing as changed, I guess

LegacyKing:
We might need to have the person doing to release change the git check out settings. I fear the common default is check out to a format that functions for the local OS, i.e. (windows line endings, put back in repo the way the file is) may be the culprit here. If I pull this update to do the release locally, then the file will end up with Window style line endings again. There has to be a method to correct this behavior.

Nylanfs:
Can it be excluded from checkout, like i can tell gut ignores the temp files that jedit makes?

Sent from Yahoo Mail on Android 
 
  On Thu, May 12, 2016 at 12:45 PM, Andrew Maitland<notifications@github.com> wrote:   
We might need to have the person doing to release change the git check out settings. I fear the common default is check out to a format that functions for the local OS, i.e. (windows line endings, put back in repo the way the file is) may be the culprit here. If I pull this update to do the release locally, then the file will end up with Window style line endings again. There has to be a method to correct this behavior.

LegacyKing:
Then it defeats the purpose of having the file. If we exclude it, then
it will not exist in the release. Release happens locally. Which means
on windows when I do it.

LegacyKing:
#1250 should prevent this issue from cropping up any more.

cpmeister:
What is the status of this?

LegacyKing:
Should be good to go...

PR #1217, opened 2016-05-11
CODE-3051: Simplify quite a bit of code in NameGenPanel

grimreaper:
Summary:
This simplifies the class NameGenPanel quite a bit and makes it easier
to process.
It also introduces a temporary class called Predicates which is intended
to be used until the switch to Java 1.8. Once that occurs this code can
be simplified even more by converting a couple of one-function-classes
to lambdas.

This also weakens many types to their higher level form (and then uses
<> to avoid redundant information)

This change also avoids the use Vector and replaces it with
List/ArrayList where possible.

(I know there should be a bug ID here for code cleanup but not sure what
the current one is)

Test Plan:
Opened GMGen NameGen and tried some functions
Built pcgen

### DIFF
--- a/code/src/java/pcgen/gui2/doomsdaybook/NameGenPanel.java
+++ b/code/src/java/pcgen/gui2/doomsdaybook/NameGenPanel.java
@@ -19,8 +19,6 @@
  */
  package pcgen.gui2.doomsdaybook;
 
-import gmgen.util.LogUtilities;
-
 import java.awt.BorderLayout;
 import java.awt.FlowLayout;
 import java.awt.Insets;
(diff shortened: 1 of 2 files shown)


grimreaper:
I'll try to be more responsive with this change than my last one.

LegacyKing:
pcgenorg.atlassian.net > You can create one for code clean up. I'll be happy to promote you to a developer if you already have access.

Also, the code peer reviewers have been rather slow to respond with other daily activities taking their time. Code refactoring looks good to me.

grimreaper:
CODE-3051: also note that I may have incomplete in my testing and I believe there are some cases where this may result in an NPE. Looking into if I can repro.

grimreaper:
I'll resend this. Think I found the issues. Sorry for the spam. Also going to split this up a bit into separate PRs :/

PR #1230, opened 2016-05-14
Merge pull request #3 from PCGen/master

trent-shipley:
pcgen20160513


LegacyKing:
Empty PR. I suggest you use the git Pull from PCGen in the future to avoid this mistake, and it reduces needless PRs since you shouldn't need to PR to update from an upstream.
Cheers,

PR #1231, opened 2016-05-14
Pcgen606x

trent-shipley:
20160513


LegacyKing:
Uh, conflicted, and all PRs should be directed to MASTER only. Only the CODE, ARCH and CONTENT Silverbacks are permitted to backport from the master into the production branches.
Plus it would be an EXTREMELY Bad mistake to merge MASTER into 6.06.x...

PR #1254, opened 2016-05-21
Acg archetypes wizard and witch

evilpixie87:
Wizard was already done, so witch and Fact system all I really had to add.


LegacyKing:
Fix by "try to patch issues #1258"

PR #1255, opened 2016-05-21
Acg archetypes warpriest witch and wizard

evilpixie87:
Sending again....


LegacyKing:
Fix by "try to patch issues #1258"

PR #1264, opened 2016-05-23
Fix various installer issues.

javydreamercsw:
The issue for the missing exe file was due to it being generated prior to it being created. This changes make sure it is generated before the copy is attempted.


javydreamercsw:
The build failure seems to be related tho this issue: https://sourceforge.net/p/launch4j/feature-requests/74/

Not related with the code itself.

LegacyKing:
Are we able to resolve this, or do we need to revert the work till the problem is resolved?

javydreamercsw:
The problem is on the server running the build. Not sure how much you guys have control over that.

This is due to the build now generating the exe file each time vs. having it as part of the controlled files. I guess another option would be to just have the exe there, but I would need to remove and change lots of stuff from what I did.

LegacyKing:
https://github.com/travis-ci/travis-ci/issues/5176

This is what you need to add to your .travis.yml to install the missing 32-bit libs:

sudo: required
before_script:
  - sudo apt-get update
  - sudo apt-get install lib32z1 lib32ncurses5

LegacyKing:
I'm testing it on my end, and doing a build test right now. So far so good.

javydreamercsw:
Got a green by Travis :D

javydreamercsw:
Pending your test and it should be ready. Thanks for the help!

LegacyKing:
I got a message:
  File: "C:\Projects\pcgen\pcgen\build/nsis/jre/jre_64/*.*" -> no files found. (C:\Projects\pcgen\pcgen/installers/win-installer/pcgen.nsis:167)
  File: "C:\Projects\pcgen\pcgen\build/nsis/jre/jre_32/*.*" -> no files found. (C:\Projects\pcgen\pcgen/installers/win-installer/pcgen.nsis:169)

Beyond that, the installer is working. The versioning looks cleaner. Only test left is if the exe shortcut works! After that, I'll be happy to merge in this "Hot fix" in... since I consider this an extension of work that was already approved by Connor.

javydreamercsw:
Hmmm. Those are the files used for embedding the jre. I'll check that file in a bit.

LegacyKing:
Okay, main issue still remains. I do not have any exe file being created.

LegacyKing:
I'm left with the install folder having no pcgen.exe file, and an empty jre folder as well.

javydreamercsw:
Checking...

javydreamercsw:
How do you call the installer build? Want to know what are you running to troubleshoot.

javydreamercsw:
I'm cloning the repository as I didn't have it at home.

javydreamercsw:
Are you running all task?

javydreamercsw:
I think I found the issue, if it was called from all->build task. Testing it now. I see it downloading the JRE and all so it seems to be working. Won't commit until I check the installer output.

LegacyKing:
Sent email - join me in hipchat to discuss.

LegacyKing:
And the tasks for testing the build are:
fullZip
buildNsis

LegacyKing:
My test branch is up to date with your master, and even just doing the buildNsis only, my shortcut can't find the pcgen.exe file.

javydreamercsw:
Try running create Executive alone

LegacyKing:
As long as task fullZip + buildNsis is used together, this works. Task buildNsis will not work on its own, as it failed to execute createExe. I can update the release notes to note this limitation.

PR #1267, opened 2016-05-24
Data 2924

LegacyKing:
@draconswan you requested this, so you get the honor of doing the deep dive testing. Standard and Unchained Versions work. You need to verify that the archetypes still work correctly.

This was some major refactoring work.


draconswan:
I'll begin my testing then

PR #1303, opened 2016-05-31
Add enhancements for the Jenkins release build

javydreamercsw:
1) Fix deployAutobuild to depend on createExe so the exe file is generated if needed.
2) Add task to correct line endings of the pcgen.sh file to be linux.

### DIFF
--- a/installers/win-installer/gendatalist.pl
+++ b/installers/win-installer/gendatalist.pl
@@ -86,6 +86,7 @@ $pub{'reality_deviant'} = 'Reality Deviant Publications';
 $pub{'rite'} = 'Rite Publishing';
 $pub{'rite_publishing'} = 'Rite Publishing';
 $pub{'rpg_objects'} = 'RPG Objects';
+$pub{'sagaborn'} = 'Sagaborn';
 $pub{'secular_games'} = 'Secular Games';
 $pub{'silven_publishing'} = 'Silven Publishing';
 $pub{'silverthorne_games'} = 'Silverthorne Games';


cpmeister:
This has conflicts.

javydreamercsw:
This is due to https://pcgenorg.atlassian.net/browse/ADMIN-49. Back when the pull request was done there was no issues.

LegacyKing:
@javydreamercsw  you able to update the PR to resolve conflicts? I'm rerunning travis.

LegacyKing:
Execution failed for task ':unixLineEndings'.
> A problem occurred starting process 'command 'fromdos''

LegacyKing:
Per Anestis:
I've install dos2unix/unix2dos.
unix2dos and dos2unix are the actual commands.

javydreamercsw:
I'll take a look.

javydreamercsw:
Well, it's using the right command now:

> A problem occurred starting process 'command 'dos2unix''

I tried adding it to Travis, hopefully that does it.

javydreamercsw:
Great, that did the trick. Up to you now!

LegacyKing:
@cpmeister Since you're back, I'll let you handle the merge.

build.gradle
```diff
@@ -220,53 +221,65 @@ artifacts {
...
+    into "$projectDir"
+    rename "(.+)-$version(.+)", '$1$2'
+}
+
+//Install the tools from: http://www.thefreecountry.com/tofrodos/
+task unixLineEndings( type: Exec) {
+    if(Os.isFamily(Os.FAMILY_WINDOWS)){
+        workingDir "$projectDir/code/sbin"
+        executable "bash"
+        args "-c", "fromdos $projectDir/code/pcgen.sh"
```
cpmeister: Windows does not come with "bash" so I'm unsure of how this would execute on a windows machine...

javydreamercsw: Thanks for pointing that out.

build.gradle
```diff
@@ -220,53 +221,63 @@ artifacts {
...
+    from "$buildDir/release/pcgen.exe"
+    from "$projectDir/code/pcgen.sh"
+    into "$projectDir"
+    rename "(.+)-$version(.+)", '$1$2'
+}
+
+//Install the tools from: http://www.thefreecountry.com/tofrodos/
+task unixLineEndings( type: Exec) {
+    if(Os.isFamily(Os.FAMILY_WINDOWS)){
+        commandLine "cmd", "/c", "fromdos $projectDir/code/pcgen.sh"
```
cpmeister: This still isn't quite right, 'fromdos' is not a native executable on windows. I think you need to specify the file path of fromdos.exe for this to work. I don't see you including fromdox.exe in this commit so that likely needs to be added too.

javydreamercsw:
The same is true with the Linux tools. They are not installed. I guess I could bundle those if approved by the rest of the team.

javydreamercsw:
Added the exe for windows. Linux doesn't work the same as it needs to be compiled on the target. It can easily be fixed for travis adding the dependency.

.nb-gradle-properties
```diff
@@ -0,0 +1,17 @@
...
+      <task-names>
+        <name must-exist="no">installFrodos</name>
+        <name must-exist="no">--stacktrace</name>
+        <name must-exist="no">--info</name>
+      </task-names>
+      <task-args/>
+      <task-jvm-args/>
+    </task>
+  </common-tasks>
+</gradle-project-properties>
```
cpmeister: This file needs to be excluded from the PR, actually you can add this file to the .gitignore file as well to be safe.

cpmeister:
Why are all the c source files being included as well? Do these need to be compiled on the machine before the gradle script can use it? Or are you expecting gradle to compile them before use?

javydreamercsw:
Well, for Linux it needs to be compiled on target or user prompted to
install some stuff.

Not sure what's the preference.

grimreaper:
Please use sed instead of checking in C files. You can also just set .gitattributes correctly.

javydreamercsw:
What you mean by sed?

javydreamercsw:
Now need to merge again...

grimreaper:
dos2unix: sed -i '' -e 's/^M$//'
unix2dos: sed -i '' -e 's/^M$//' -e 's/$/\r/'

javydreamercsw:
Still not sure what you mean. Can you show me what would be the real command to convert pcgen.sh?

I'm clueless.

thpr:
This needs a rebase and squash to reduce complexity.

LegacyKing:
@javydreamercsw Are you able to move this into a new branch, rebase and squash?

javydreamercsw:
I guess I could.

PR #1337, opened 2016-06-19
Modernize the OSX GUI code

grimreaper:
The old OSX gui2 code relies on a deprecated and poorly designed
interface.  Switch to using a modern variant of these methods.

While here
 - fix mode on pcgen.sh
- add missing newline to gradle.properties.


grimreaper:
The build failure appears to be because its running the OSX code on a debian machine. Is there a way to avoid this?

javydreamercsw:
To me it looks like a version mismatch as the methods in the code are no longer there. But you might be right. 

I guess my question is why it didn't fail prior to this change.

grimreaper:
> I guess my question is why it didn't fail prior to this change.
I have the same question. I tried a few variations and they all fail. This looks like some kind of gradle or dependency issue to me, but I don't full understand what's going on here :(

Is there a way to skip this on non-osx?

javydreamercsw:
Skip what? If you meant the includes, this should help:

http://stackoverflow.com/questions/8796615/modify-gradles-runtime-dependencies-according-to-operation-system

cpmeister:
It looks like a dependency issue to me. It doesn't matter if the jars included are fully implemented or not, the necessary classes and methods must be present. We previously used a osx stub jar for compilation and when running on osx we used the real jar provided by the os. When not running on osx none of the stub methods were called so there was never a problem. In summary, I can't imagine this being anything other than a dependency issue.

cpmeister:
I think I know what is going on. I think the old apple jar is interfering with the compilation. Remove that jar from the dependencies and think it will solve this problem.

cpmeister:
I was referring to the MRJ14 jar...

PR #1343, opened 2016-06-25
Fix a couple of different issues

grimreaper:
- remove some methods that are not needed
        - convert and remove some deprecated methods
        - Fix the documentation
        - Don't compare a possible NaN to NaN since this will always be
          false


LegacyKing:
Looks good to me. Waiting for @cpmeister to review.

PR #1347, opened 2016-06-29
Stat output token fix

Zaister:
This fixes DATA-2974 with the help of the underlying CODE-3068


Zaister:
@cpmeister can you please check this?

LegacyKing:
Data looks good. @cpmeister needs Code review.

cpmeister:
The addition of a hard coded ability type in the code is very concerning. Is there a way to fix this that doesn't require the code to define a "Equipment" type for abilities?

Zaister:
Unfortunately no, I think. I have found no way to programmatically deterime what object has granted an ability. If that were available, the hard-coding could be avoided. However, the STAT output token is destined for depreciation anyway, so this hard-coding should go away then anyway.

PR #1357, opened 2016-07-07
DATA-2979 bloatmage base saves have wrong formula

oiml:


LegacyKing:
Awaiting proper formula - see DATA-2979 Jira for correct formula. https://pcgenorg.atlassian.net/projects/DATA/issues/DATA-2979?filter=allopenissues

PR #1364, opened 2016-07-12
Implement DATA-2987: Allow inubrix for ammunition

Zaister:
Also some improvements on the PRETYPE tags in this file.


LegacyKing:
Merge checker is frozen... looks like I'll have to notify the github devs.

PR #1394, opened 2016-07-28
Clean up output sheet to avoid FOP warnings

Zaister:
This updates our default output sheets to current FOP syntax, so that they do not generate excessive warnings and can even be loaded with strict validation active.

These four syntax issues were occurring widely throughout the files:
* Empty `<fo:table-cell>` nodes are no longer allowed in FOP 1.0+. We had a lot of `<fo:table-cell/>`, or cells that can be empty if a loop or template called inside returns an empty result. I have added `<fo:block/>` to the empty cells or wrapped the loops and templates in a `<fo:block>` node.
* The `padding` attributes cannot be used with tables that use the collapsed-border table model and are ignored, but generate a warning. I have removed these attributes.
* The fixed table-layout assumes `width="auto"` when the `width` attribute is not present. However, auto layout is not supported by FOP. FOP then assumes `width="100%"` and generates a warning. I have explicitly added the `width="100%"` attribute.
* The attributes `border-top` and `border-bottom` should actually be `border-top-width` and `border-bottom-width`. I have fixed these appropriately.

### DIFF
--- a/outputsheets/d20/fantasy/pdf/common_sheet/block_attack.xslt
+++ b/outputsheets/d20/fantasy/pdf/common_sheet/block_attack.xslt
@@ -18,7 +18,7 @@
 ====================================-->
 	<!-- Begin Conditional Combat Modifiers -->
 	<xsl:template match="attack" mode="conditional">
-		<fo:table table-layout="fixed" space-before="2mm">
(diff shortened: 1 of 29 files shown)


Zaister:
For now, only the sheets in the `d20/fantasy` folder have been cleaned. I will fix the rest later.

PR #1396, opened 2016-07-28
Fix OS-292: Support UTF-8

Zaister:
This solves the UTF-8 problem by using features of the updated FOP library to imbed a custom font. I have chosen Noto Sans (https://fonts.google.com/specimen/Noto+Sans) which is free and available under the Apache License, Version 2.0 (http://www.apache.org/licenses/LICENSE-2.0). This font is designed to reduce the so-called "tofu" (i.e. the boxes produced by missing glyphs) and supports most international letters (but not Chinese or Japanese). @Nylanfs has commented that the license should be compatible with PCGen, so we should have no problems using and redistributing these fonts.

I have added a new `fonts` directory under `outputsheets` where I have added the files for this font and  adapted the FOP configuration file to load them. Furthermore I have updated the fantasy output sheets (so far) to use these fonts.

Advanced users can even add their own fonts by copying them into the `fonts` directory and modifying the FOP configuration file and the output sheets accordingly.

I have taken the opportunity to update the footer templates of the output sheets that contain the credits for the sheets to match the current state of the project.

I'd like to have this change reviewed by @LegacyKing and @cpmeister as it touches various areas of PCGen and even includes a code modification and adds a new library to be able to read the font configuration from the FOP configuration file. Please comment.


outputsheets/d20/fantasy/pdf/common_sheet/block_bio_condensed.xslt
```diff
@@ -50,6 +50,7 @@
 <!-->		<xsl:if test="string-length(translate(normalize-space(concat(description,bio)), ' ', '')) &gt; 0">	-->
 			<fo:page-sequence>
 				<xsl:attribute name="master-reference">Portrait</xsl:attribute>
+				<xsl:attribute name="font-family">NotoSans</xsl:attribute>
```
LegacyKing: Just a comment, not a fix required, but annotations about what is happening would be handy for the uninitiated. Like "Using notoSans here gives us control over the export fonts used".

LegacyKing: I'm seeing this repeated, is there no way to place this into a global var and allow it to be set in one location? Otherwise, a lot of room for missing the font family.

LegacyKing:
Other than I'd like to see a central var be used to reduce the number of files changed if someone wishes to use another font family, this looks good to me.

@cpmeister this is merge worthy on the content side.

Zaister:
Good idea about changing the font in only one place. So I will change fop.xconf to give the configured font the alias "PCGen Sans". This way can use that alias in the various output sheets and then you can simply edit the fop.xcof to use different font files.

LegacyKing:
COMMENT: Awesome! Thanks for that. Can we add a note to the WIKI about it, and that way during the doc update project, we can get that included in the Advanced Homebrewer section. :)  +1

Zaister:
OK, changed the files as described above. now only  fop.xconf need to be changed to use another font family. I have added a comment to the file about how to do that.

build.gradle
```diff
@@ -148,6 +148,7 @@ dependencies {
     compile group: 'mrj', name: 'MRJ141Stubs', version:'0'
     compile group: 'net.sf.squirrel-sql.thirdparty-non-maven', name: 'skinlf', version:'6.7'
     compile group: 'org.apache.xmlgraphics', name: 'fop', version:'2.1'
+    compile group: 'org.apache.pdfbox', name: 'pdfbox', version:'1.8.9'
```
cpmeister: Why is pdfbox needed?

Zaister: pdfbox, or rather fontbox which is a part of pdfbox, is needed by fop to embed custom fonts.

Zaister: I noticed a problem with the approach I'm using. The PDF renderer is not the only renderer we're using, as the print preview in the GUI uses the AWT renderer. The AWT renderer, unfortunately, is not configured by the fop.xconf file, and for fonts can only use installed system fonts. This means, as of now in this branch, it tries to use a system font called "PCGenSans" which, of course, it canot find, and then falls back to default, which is the Times font included in the FOP library that looks a lot unlike the actual print output.

A solution would be to use "Arial" as the alias instead of "PCGenSans". Arial can reasonably be expected to be installed on a user's system, and while rendering the print preview in Arial does not exactly conform to the actal print output, is looks close, and Arial was what we've been using there until now.

Thoughts?

Zaister:
Note: there is an even newer version 2.0.2 of fontbox, but version 2 or higher seems to be incompatible with current FOP.

cpmeister:
Would [this](http://stackoverflow.com/questions/5652344/how-can-i-use-a-custom-font-in-java) help?

Zaister:
Thanks, I will check that out.

In the meantime I noticed that fontbox 2.0.2 does actually work for us. My first using it test failed, but then I was inclunding the whole pdfbox package.

Zaister:
@cpmeister your tip checks out, I can use the custom font with that in the AWT renderer. However, I can't use the alias then, so I'll have to move the font family name in the XSLT sheets into a centrailzed XSL variable.

Zaister:
OK, here's the new status of this branch
* The font alias in `fop.xconf` has been eliminated and replaced by an XSL variable defined in `outputsheets/d20/fantasy/pdf/fantasy_common`.
* Hardcoded loading the Noto Sans fonts into the current GraphicsEnvironment in the `PrintPreviewDialog` class, so they can be used by the AWT renderer.
* Updated fontbox library to most current version 2.0.2

I think this is now ready to be merged. More character sheets (e.g. other game modes) can then be adapted gradually.

Zaister:
The Travis CI build failure seems unrelated. Looks like `http://librepo.pcgen.org` is unavailable and this causes the failure.

cpmeister:
In order to prevent the fonts to be reloaded everytime the print preview dialog is shown I think that the font registration logic should be put inside Main.java so that it occurs only once on startup.
Other than that everything looks good. Nice work.

Zaister:
Actually it happens inside a static method, so it's only perfomed once, and only if the print print preview is actually called up. But I can move it into Main.java if you prefer.

cpmeister:
Yes, please do.

Zaister:
OK, I will, then

Zaister:
I've moved the font initialization o code to a new method in `Main.java` that is called from `startupWithGUI()`.

PR #1399, opened 2016-07-30
Data 3013

Llisandur:
Fixed the White Mage archetype's Fast Healing greater exploit again. I didn't notice last time that it was requiring itself, or that it was looking for its archetype requirement in the wrong CATEGORY. It's all fixed and works now.


LegacyKing:
You should rebase your branches to remove all that extra merging, or use direct pulls. I should only see actual commits in your PR, not merges.

Llisandur:
Ok. I've been looking for how to do that, but I didn't know what to call it until now. I should have it fixed for next time.

PR #1408, opened 2016-08-13
Added type, benefit, and desc support to freemarker output sheets

cpmeister:


cpmeister:
@thpr Can you take a look at this? I made some changes to ResultFacet that I just want to you look over to make sure I'm not doing something wrong.

code/src/java/pcgen/output/model/InfoModel.java
```diff
@@ -78,21 +78,22 @@ public TemplateModel get(String key) throws TemplateModelException
 	{
 		CaseInsensitiveString cis = new CaseInsensitiveString(key);
 		MessageFormat info = cdo.get(MapKey.INFO, cis);
-		if (info == null)
+		
+		StringBuffer sb = new StringBuffer(100);
+		if (info != null)
```
thpr: I'm not a fan here.  If info is null, we need to produce some form of error... not suppress it and leave it in the output sheet.

code/src/java/pcgen/cdom/facet/analysis/ResultFacet.java
```diff
@@ -57,14 +57,19 @@ public Object getGlobalVariable(CharID id, String varName)
 
 	public Object getLocalVariable(CharID id, CDOMObject cdo, String varName)
 	{
+		String localScopeName = cdo.getLocalScopeName();
+		if(localScopeName == null){
```
thpr: Well, first, this is not our code standards :) ... space after if, open curly brace on a new line.
Other than that, yes, it looks correct

code/src/java/pcgen/cdom/facet/analysis/ResultFacet.java
```diff
@@ -57,14 +57,19 @@ public Object getGlobalVariable(CharID id, String varName)
...
+			return getGlobalVariable(id, varName);
+		}
+		
 		/*
 		 * TODO Given who this serves, this is an interesting situation. This
```
thpr: Please delete the TODO, since you basically addressed it by dropping to Global when necessary.

PR #1422, opened 2016-08-21
Style Change: use diamond operator

grimreaper:
Across the board prefer using the <> operator instead of duplicating
type information. This reduces the duplication of information and makes
it easier to make future changes and cleanups


Zaister:
I'm confused, when did we change our coding standards?

PR #1428, opened 2016-08-22
Remove dead code; modernize where possible

grimreaper:
- remove code which is unused
- modernize some code which check for null to pass through Objects.requireNonNull
- make some modifiers private instead of public
- prefer {@code} instead of &lt;code&gt;
- shift of 32 is equal to shift of 0 (which is equal to no shift at all)
- remove some references to subversion


code/src/java/pcgen/util/Comparators.java
```diff
@@ -211,30 +186,26 @@ public int compare(Object o1, Object o2)
...
 			}
-			else if (obj1 instanceof CDOMObject)
-			{
-				CDOMObject co = (CDOMObject) obj1;
-				key = co.get(StringKey.SORT_KEY);
-				if (key == null)
-				{
-					key = co.getDisplayName();
+			else {
+				if (obj1 instanceof CDOMObject) {
```
LegacyKing: Will point this out. The code standard is curly braces are on their own lines...

if(some such thing)
{
    some code
}

I personally think it adds more lines, but it does make readability easier.

grimreaper: Thanks! Changed the style in my IDE to match (I hope?)

cpmeister:
I think this PR simply does too many different things. There are some that I approve of and some things that I disapprove of. I would prefer if PRs would have a single type of change in mind rather than a multiple types of changes under the vague and subjective goal of "modernization".

remove code which is unused - not all code that is unused should be removed. Some code simply has yet to be used.
modernize some code which check for null to pass through Objects.requireNonNull - I'm ok with this
make some modifiers private instead of public - some of these are intionally public, even if they are unused publicly.
prefer {@code} instead of <code> - approved
shift of 32 is equal to shift of 0 (which is equal to no shift at all) - I think this is a code bug rather than an intentional operation
remove some references to subversion - approved

PR #1438, opened 2016-08-24
modernize and simplify

grimreaper:
modernize these files

code/src/java/pcgen/core/SettingsHandler.java
code/src/java/pcgen/gui2/facade/Gui2CampaignInfoFactory.java
code/src/java/pcgen/persistence/PersistenceManager.java
code/src/java/pcgen/persistence/SourceFileLoader.java
code/src/java/pcgen/persistence/lst/LstSystemLoader.java
code/src/java/pcgen/persistence/lst/TraitLoader.java
code/src/java/pcgen/persistence/lst/URIEntry.java
code/src/java/pcgen/persistence/lst/URIFactory.java
code/src/java/pcgen/system/FacadeFactory.java
code/src/java/pcgen/util/chooser/ChoiceHandler.java
code/src/java/pcgen/util/chooser/ChooserFactory.java
code/src/java/pcgen/util/chooser/ChooserInterface.java
code/src/java/pcgen/util/chooser/RandomChooser.java
code/src/java/plugin/pretokens/test/PreCampaignTester.java
code/src/test/pcgen/core/prereq/PreCampaignTest.java

and rebuild wrapper to include source

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ apply plugin: 'build-dashboard'
 
 group = 'net.sourceforge.pcgen'
 
-description = """PCGen"""
+description = 'PCGen'
 
 defaultTasks 'clean', 'build'
 
@@ -166,6 +166,7 @@ dependencies {
     compile group: 'net.sourceforge.argparse4j', name: 'argparse4j', version: '0.7.0'
     compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.55'
     compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.64'
+    compile group: 'org.jetbrains', name: 'annotations', version:'15.0'
 
 
     //testCompile group: 'org.easymock', name: 'easymock', version:'2.5.2'
--- a/code/src/java/pcgen/util/chooser/RandomChooser.java
+++ b/code/src/java/pcgen/util/chooser/RandomChooser.java
@@ -20,7 +20,6 @@ package pcgen.util.chooser;
 
 import java.util.ArrayList;
 import java.util.List;
-
 import pcgen.base.util.RandomUtil;
 import pcgen.facade.core.ChooserFacade;
 import pcgen.facade.core.InfoFacade;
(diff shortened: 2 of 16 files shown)


build.gradle
```diff
@@ -166,6 +166,7 @@ dependencies {
     compile group: 'net.sourceforge.argparse4j', name: 'argparse4j', version: '0.7.0'
     compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.55'
     compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.64'
+    compile group: 'org.jetbrains', name: 'annotations', version:'15.0'
```
cpmeister: What is the benefit of the annotations in this library?

code/src/java/pcgen/util/chooser/RandomChooser.java
```diff
@@ -52,33 +51,14 @@
...
-	/**
-	 * Sets the AllowsDups attribute of the Chooser object
-	 *
-	 * @param aBool  The new AllowsDups value
-	 */
-    @Override
-	public void setAllowsDups(boolean aBool)
-	{
-		theAllowDuplicatesFlag = aBool;
-	}
```
cpmeister: I'm pretty sure that this setter has a purpose, even if it isn't used. Please don't remove it.

build.gradle
grimreaper: It serves as a hinter for FindBugs, Eclipse/IntelliJ, and various other tools.
See for example http://help.eclipse.org/kepler/index.jsp?topic=/org.eclipse.jdt.doc.user/tasks/task-using_null_annotations.htm

cpmeister:
I think that adding final to parameters that don't need it shouldn't be encouraged to final nor enforced. I can understand from a coding standpoint for new code to have final parameters at the coder's discretion but I don't think it is something that existing code should be changed because of.

build.gradle
cpmeister: Is it possible to use JSR 305 annotations instead? Netbeans (the IDE I use) doesn't support the jetbrains annotations as far as I know.

grimreaper: Its the same thing AFAIK. You just need a place to get the annotations from.
You can also get the same annotations from com.google.code.findbugs

grimreaper:
moving the annotations into a separate PR.  It is going to be a base commit for future work on PlayerCharacter.

PR #1447, opened 2016-08-26
Complete dice work

grimreaper:
This finishes the work to modularize and abstract out the dice work. It also kills DiceEx which isn't used and should be rebuilt as a DiceConfiguration.

There is more than can be done: making toFormula be generated from the collector + modifiers. The Dice, SystemDie, and related classes could be pulled apart and moved into the user classes, but this is good enough for now.


cpmeister:
This has  conflicts

cpmeister:
The new files need the license headers

code/src/java/gmgen/plugin/dice/Die.java
```diff
@@ -20,45 +20,34 @@
...
  *  Created on January 24, 2002, 11:15 AM
  */
-package gmgen.plugin;
+package gmgen.plugin.dice;
 
 import java.util.Random;
 
-/** Abstract class describing a die of any kind
- * @author Soulcatcher
- * @since May 24, 2003
```
cpmeister: Why are you removing the author tag?

grimreaper: This particular file has been basically stripped to the bare minimum and no longer directly reflects the work of the author. I can restore of required, but its of questionable value (vcs stores this history)

cpmeister:
Add license headers to the new files and resolve the conflicts and I'll merge it.

grimreaper:
license headers added

PR #1458, opened 2016-08-26
autobranch: part 9

grimreaper:


code/src/java/pcgen/rules/persistence/token/PreCompatibilityToken.java
```diff
@@ -91,17 +91,17 @@ public String getTokenName()
 		return tokenName;
 	}
 
-	public int compatibilityLevel()
+	public static int compatibilityLevel()
```
thpr: This method (and the two below) are used in the CompatibilityToken interface.
This is a red flag to me that something funny is going on and this check-in needs to be reviewed in more detail (there may be a code bug somewhere?)

PR #1473, opened 2016-08-26
autobranch: part 21

grimreaper:


code/src/java/pcgen/util/PJEP.java
```diff
@@ -301,7 +301,7 @@ else if (parent instanceof PlayerCharacter)
 	 * @return Returns the variableSource.
 	 */
 	@Deprecated
-	protected String getVariableSource()
+	private String getVariableSource()
```
cpmeister: If you are going to make this private might as well delete this method. It seems deprecated anyway.

cpmeister:
Fix the build failure and I'll merge it.

code/src/java/pcgen/util/PJEP.java
thpr: Unfortunately it's used, and since this is the old formula system, I think it's better off not being messed with, as we're phasing it out anyway.

grimreaper:
build failure fixed. Also fixed a local setting so that my default tests include this :)

code/src/java/pcgen/util/PJEP.java
cpmeister: @grimreaper I guess you should change this back to protected then and if Tom doesn't have any more objections I'll merge it.

grimreaper: done

PR #1483, opened 2016-08-26
autobranch: part 31

grimreaper:


cpmeister:
Objects.equals(a,b) is not equivalent to (a != null && a.equals(b)) and as such it cannot be used as a replacement.

PR #1485, opened 2016-08-26
autobranch: part 33

grimreaper:


code/src/java/pcgen/cdom/facet/DamageReductionFacet.java
```diff
@@ -112,61 +113,67 @@ public void dataRemoved(DataFacetChangeEvent<CharID, CDOMObject> dfce)
 			return andMap;
 		}
 		CaseInsensitiveMap<Integer> orMap = new CaseInsensitiveMap<>();
-		for (Map.Entry<DamageReduction, Set<Object>> me : componentMap
-				.entrySet())
+		/*
+		 * TODO Shouldn't this expansion be done in the DR
+		 * token? (since it's static?)
+		 *//*
+		  * TODO Shouldn't this expansion be done in the DR
```
thpr: Kind of concerned with these things showing up multiple times.    These will need review, and thus I think we need to do these in smaller chunks (800 files way too many)

grimreaper:
This was the original branch that I split into other PRs. Did not mean to make this one a real diff.

PR #1491, opened 2016-08-28
move attribute fields outside of PlayerCharacter

grimreaper:
Rather than keep PlayerCharacter attributes hardcoded with specific getters and setters for each, move them into appropriate enum types.

### DIFF
--- a/code/src/java/pcgen/core/BioSet.java
+++ b/code/src/java/pcgen/core/BioSet.java
@@ -41,6 +41,7 @@ import pcgen.base.util.DoubleKeyMap;
 import pcgen.base.util.TripleKeyMapToList;
 import pcgen.cdom.base.Constants;
 import pcgen.cdom.base.NonInteractive;
+import pcgen.cdom.enumeration.PCAttribute;
 import pcgen.cdom.enumeration.Region;
 import pcgen.util.Logging;
 
@@ -229,36 +230,13 @@ public final class BioSet extends PObject implements NonInteractive
 
 		if (ranList.contains("HAIR"))
 		{
-			pc.setHairColor(generateBioValue("HAIR", pc));
+			pc.setPCAttribute(PCAttribute.HAIRCOLOR, generateBioValue("HAIR", pc));
 		}
 
 		if (ranList.contains("SKIN"))
 		{
-			pc.setSkinColor(generateBioValue("SKINTONE", pc));
+			pc.setPCAttribute(PCAttribute.SKINCOLOR, generateBioValue("SKINTONE", pc));
 		}
-	}
-
-	/**
-	 * Remove the user from the map
-	 * @param region
-	 * @param race
-	 * @param tag
-	 */
-	public void removeFromUserMap(final String region, final String race, final String tag)
-	{
-		final String key;
-		final int x = tag.indexOf(':');
-
-		if (x < 0)
-		{
-			key = tag;
-		}
-		else
-		{
-			key = tag.substring(0, x);
-		}
-
-		userMap.removeListFor(Region.getConstant(region), race, key);
 	}
 
 	@Override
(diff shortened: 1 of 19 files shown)


code/src/java/pcgen/core/GenericPCAttribute.java
```diff
@@ -0,0 +1,24 @@
+/*
+ *  Initiative - A role playing utility to track turns
```
LegacyKing: duplicate title?

code/src/java/pcgen/cdom/enumeration/PCStringKey.java
```diff
@@ -51,8 +51,8 @@
 	PERSONALITY2,
 	PHOBIAS,
 	PLAYERSNAME,
-	RESIDENCE,
 	SKINCOLOR,
+	RESIDENCE,
```
thpr: Why is this being taken out of alphabetical order?

code/src/java/pcgen/core/BioSet.java
```diff
@@ -250,14 +251,7 @@ public void removeFromUserMap(final String region, final String race, final Stri
...
 
-		if (x < 0)
-		{
-			key = tag;
-		}
-		else
-		{
-			key = tag.substring(0, x);
-		}
+		key = x < 0 ? tag : tag.substring(0, x);
```
thpr: This type of in-lining is controversial of whether it is a best practice.  Before we make these types of changes, we need a discussion on the code list as to best practices.  I happen to dislike this usage because the "x < 0" is not in parenthesis and thus the line is harder to interpret.

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -405,15 +397,15 @@
 	private MasterFacet masterFacet = FacetLibrary.getFacet(MasterFacet.class);
 	private AutoEquipmentListFacet autoListEquipmentFacet = FacetLibrary.getFacet(AutoEquipmentListFacet.class);
 	private FollowerFacet followerFacet = FacetLibrary.getFacet(FollowerFacet.class);
-	
+
```
thpr: I'm really not a fan of changing spacing.  In the past this has created a lot of thrashing between different committers as different IDEs have different spacing and formatting rules.  We should try to focus on functional changes as much as we can and minimize these types of changes unless something is wildly out of whack.

thpr:
Also note that before this is merged, our code standard is a large number of commits (more than single digits) should be combined (Squashed via Rebase) to keep a cleaner history...

code/src/java/pcgen/core/PlayerCharacter.java
grimreaper: agreed. This is entirely my IDE doing this change. We should also pick a standard and stick to it.

grimreaper:
can't you get the clean history by doing the squash merge?
history is roughly identical when using merge commits?

code/src/java/pcgen/cdom/enumeration/PCStringKey.java
grimreaper: fixed

code/src/java/pcgen/core/PlayerCharacter.java
grimreaper: also, I plan on making more changes to PlayerCharacter so I'd prefer to just set my IDE correctly for pcgen and let it make the fixes

cpmeister:
Find a way to squash some of these commits together and I'll merge it.

grimreaper:
squashed into a single commit.

code/src/java/pcgen/core/BioSet.java
thpr: Withdrawn based on discussion on IRC

PR #1493, opened 2016-08-28
slight cleanup of bioset

grimreaper:


code/src/java/pcgen/core/BioSet.java
```diff
@@ -411,19 +409,16 @@ private String appendAgesetInfo(Region region,
...
-				final String aRaceName = raceIt.next();
-
-				if (!"AGESET".equals(aRaceName))
+				if (!"AGESET".equals(stringSortedMapEntry.getKey()))
 				{
-					final SortedMap<String, String> tags = races.get(aRaceName);
+					final SortedMap<String, String> tags = stringSortedMapEntry.getValue();
 
-					for (Iterator<String> tagIt = tags.keySet().iterator(); tagIt.hasNext();)
+					for (final Map.Entry<String, String> stringStringEntry : tags.entrySet())
```
thpr: I think before we merge any more "final" items we need to have a code team discussion.  This has been hashed out before, and putting them in does not match the output of our last discussion, so I can't really support these being merged.

PR #1494, opened 2016-08-28
a mixture of automated and manual cleanups to PlayerCharacter

grimreaper:
Since I plan on spending some time with PlayerCharacter this is a 'base cleanup' commit to simplify future reviews.


cpmeister:
I think these changes would likely serve to give Tom a headache.

PR #1495, opened 2016-08-28
remove 'load' from PlayerCharacter ctor

grimreaper:
this isn't used and harms my plan to separate out PlayerCharacter into composable chunks.


thpr:
This is a red flag to me and I'd like to do a deep review of this before it is merged

grimreaper:
I presume there is history here. Semantically this code is identical so either
(a) load used to be used, and isn't needed anymore
(b) load used to be used, is still needed, and there is a bug
(c/d) load was added but it isn't used yet (either bug or not a bug)

thpr:
Yes, nasty bugs in the past that made it necessary.  I have reviewed the history and this can be safely removed, thanks for the cleanup work.

PR #1496, opened 2016-08-28
just add annotations for @Nullable, @NotNull, and @Contract

grimreaper:
This is like JSR308 but lets it be used

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -166,7 +166,7 @@ dependencies {
     compile group: 'net.sourceforge.argparse4j', name: 'argparse4j', version: '0.7.0'
     compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.55'
     compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.64'
-
+    compile group: 'org.jetbrains', name: 'annotations', version:'15.0'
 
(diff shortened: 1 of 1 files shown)


cpmeister:
I would prefer to use the JSR 305 annotations instead since it is more of a standard than the jetbrains annotations which are more or less targeted for the Intellij IDE. A standard would be more appropriate since PCGen developers use a variety of IDEs for development.

grimreaper:
JSR 305 is not implemented. You need to pull them from somewhere. These will work with any IDE.

cpmeister:
Not sure we can include this library due to licensing. PCGen releases under LGPL 2.1 which isn't compatible with the Apache 2.0 license that the jetbrains annotations are released under.

grimreaper:
Good point. Would com.google.code.findbugs be okay? It is also LGPL.

cpmeister:
As long as it is LGPLv2.1. LGPLv3 is not compatible.

cpmeister:
Yeah, I think that google's findbugs is under LGPLv3. But it might be ok since we are using the code to build PCGen not run PCGen. I'll have to get back to you on that.

grimreaper:
it seems the version of gradle we use it too old for compileOnly dependencies. It might be possible to exclude the jar for distribution some other way though? I'm not a gradle expert.

grimreaper:
reopened with change to compileOnly since the annotations library does not need to be distributed

PR #1499, opened 2016-08-28
Remove unused code from PlayerCharacter

grimreaper:
While generally some of this stuff may be 'for future use' it would be better to re-implement in the future after I've had a chance to refactor most of this code out of PlayerCharacter.
Removing code from PC now reduced the overhead and chance for conflicts as I make it composable.


thpr:
In general I'm in favor of this, what I would like to do however is sync on your plans for PlayerCharacter.  I have two concerns: (1) I have 40 or so local branches many which impact Player Character.  Major changes cause a lot of rebase/merge work for me. (2) There are changes going on that will deprecate parts of Player Character and I'd hate for you to invest in rebuilding things I just remove soon after.

grimreaper:
I'd be happy to chat more - it might be easier to chat in the HC or IRC channel rather than over github.
Broadly speaking what I'm trying to do is decompose PlayerCharacter into a series of interfaces. That itself is a stepping stone to fully modularizing PlayerCharacter such that it doesn't hold much state.
For the first step in the cleanup what I'd like to see is
PlayerCharacter implements PCWithEquipment, PCWithLevel, PCWithClasses
and then migrate existing types of PlayerCharacter into interfaces.

PR #1501, opened 2016-08-28
main can't be contrusted

grimreaper:


cpmeister:
Technically this doesn't fit the PCGen formatting standards, even thought it does look neat and concise. Might consider changing the coding standards at some point to allow something like this but for now please format the brackets to separate lines and I'll merge it.

PR #1502, opened 2016-08-28
move PCU into PC

grimreaper:
-additional prep work for PC


cpmeister:
I think @thpr still has a plan regarding PlayerCharacterUtilities. So I would need him to weigh in on this before it can be deleted.

thpr:
I'll work with his plan for the moment.  Longer term arch discussion on why this existed depending on where things go over time

PR #1503, opened 2016-08-28
clean up Globals a bit

grimreaper:
Remove tons of stuff from Globals which isn't really used


grimreaper:
can't repro, remove for now

PR #1508, opened 2016-08-29
clean up Globals a bit (attempt #2)

grimreaper:


code/src/java/plugin/pretokens/test/PreCampaignTester.java
```diff
@@ -159,7 +159,7 @@ private int countCampaignByName(final String key, CDOMObject source,
 			List<URI> selCampaigns = pMan.getChosenCampaignSourcefiles();
 			for (URI element : selCampaigns)
 			{
-				final Campaign aCampaign = Globals.getCampaignByURI(element);
+				final Campaign aCampaign = Globals.getCampaignByURI(element, true);
```
cpmeister: Is this intentional?

grimreaper: Yes. This is the only caller that 'complains on error'. Really, that should be pulled out and moved into this function instead of being a boolean flag on getCampaignByURI but this is good enough for now.

PR #1509, opened 2016-08-29
DATA-3043 swashbuckler favored class panache bonus

Zaister:
I have no idea why Github add the previously merged output sheet commit to this PR.


LegacyKing:
Probably branched after you made the commit. quick fix is a rebase on the branch from master and then force the update. PR looks good otherwise, I'll merge it after you fix the duplicate commit issue.

Zaister:
I did as instructed, that did indeed remove the double commit. Weird.

PR #1510, opened 2016-08-30
begin process of genericifying facets, PlayerCharacter, and related code even further

grimreaper:
≥This isn't by any means optimal. What I'd love is for the facet to provide the GenericPCAttribute that it defines, rather than having the map live in PlayerCharacter. This is just a stepping stone to get there.

### DIFF
--- a/code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
+++ b/code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
@@ -120,7 +120,7 @@ public abstract class AbstractItemFacet<IDT extends PCGenIdentifier, T> extends
 	 * When the value is is null, what to return
 	 */
 
-	protected T valueWhenNull()
+	public T valueWhenNull()
 	{
 		return null;
 	}
--- a/code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
+++ b/code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
@@ -120,7 +120,8 @@ public abstract class AbstractItemFacet<IDT extends PCGenIdentifier, T> extends
 	 * When the value is is null, what to return
 	 */
 
-	protected T valueWhenNull() {
+	protected T valueWhenNull()
+	{
 		return null;
 	}
 
--- a/code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
+++ b/code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
@@ -61,7 +61,7 @@ public abstract class AbstractItemFacet<IDT extends PCGenIdentifier, T> extends
 			Logging.errorPrint(getClass() + " received null item: ignoring");
 			return false;
 		}
-		final T old = getRaw(id);
+		T old = getRaw(id);
 		if (old == obj)
 		{
 			return false;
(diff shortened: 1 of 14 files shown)


grimreaper:
tests fail with 
    junit.framework.AssertionFailedError: listener remove event count expected:<0> but was:<1>
I'm not 100% sure why. I'll investigate more tomorrow. Would still like to know if the approach is appreciated.

grimreaper:
FWIW I'd like to move away from Enums and into a proper OO design. This is a step to removing the need for PlayerCharacter to be directly aware of the facets and have facets 'declare themselves'

cpmeister:
I'll have to wait until @thpr weighs in on this. I'd hate for there to be conflicts.

code/src/utest/pcgen/cdom/facet/fact/WeightFacetTest.java
```diff
@@ -40,41 +40,33 @@ public void setUp() throws Exception
...
-
 	@Override
 	protected AbstractItemFacet<CharID, Integer> getFacet()
 	{
 		return facet;
 	}
 
-	private int n = 1;
+	// random value > 0
+	private int n = 123242;
```
cpmeister: What is this for?

grimreaper: the real reason for this is that in AgeFacet it was originally 0 which caused confusion with the newly implemented valueWhenNull().
the secondary reason is that test values should typically be somewhat arbitrary.

cpmeister: Ah, noted

code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
```diff
@@ -110,9 +110,23 @@ public T remove(IDT id)
 	 * @return the item value for this AbstractItemFacet and the Player
 	 *         Character represented by the given PCGenIdentifier.
 	 */
-	public T get(IDT id)
+	public T get(final IDT id)
+	{
+		final T tmp = (T)getCache(id);
```
thpr: I'm sorry, but these final entries aren't valuable.  There are plenty of automated tools that want to do this, but this destroys code clarity.
Final should be used where it actually provides a hint to the user, not everywhere it might be possible to use it under any circumstances.
These usages as well as on parameters I think are actually bad usages of final

code/src/utest/pcgen/cdom/facet/fact/AgeFacetTest.java
```diff
@@ -48,13 +49,14 @@ protected Integer getItem()
 	@Test
 	public void testItemSetMultGetAge()
 	{
-		assertEquals(0, facet.getAge(id));
+
+		assertEquals(0, (int)facet.get(id));
 		getFacet().set(id, 4);
-		assertEquals(4, facet.getAge(id));
+		assertEquals(4, (int)facet.get(id));
```
thpr: I'm not convinced this is a valuable change, to force a cast vs having type safety beforehand.
This is an architectural decision of type safety, and I'm not convinced this "moves things forward".  This needs further discussion on a list

grimreaper: Do you have a suggestion for how to avoid this? The issue is that assertEquals works both Object and int and thus generates ambiguity.

code/src/java/pcgen/cdom/facet/base/AbstractItemFacet.java
grimreaper: I can easily remove the final style change. What do you think of the architectural approach though?

grimreaper:
how is the code now? Removed the casts, removed the finals.

grimreaper:
This needs to be reworked since the core code here caused integration test failures

PR #1514, opened 2016-08-31
make use of Stream API

grimreaper:
- hasControl() only needs a .contains call instead of a .get() call
- simplify some logic by using the streams API
- reduce visibility for things which aren't used outside of PC (makes it easier to spot errors in the future)

### DIFF
--- a/code/src/java/pcgen/core/PlayerCharacter.java
+++ b/code/src/java/pcgen/core/PlayerCharacter.java
@@ -1131,17 +1131,10 @@ public class PlayerCharacter implements Cloneable, VariableContainer
 	 */
 	private static Equipment getEquipmentNamed(final String aString, final Collection<Equipment> aList)
 	{
-		Equipment match = null;
-
-		for (Equipment eq : aList)
-		{
-			if (aString.equalsIgnoreCase(eq.getName()))
-			{
-				match = eq;
-			}
-		}
-
-		return match;
+		return aList.stream()
+				.filter(v -> aString.equalsIgnoreCase(v.getName()))
+				.reduce((a, b) -> b) // get last item
+				.orElse(null);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -10085,6 +10057,6 @@ public String getControl(String string)
 
 	public boolean hasControl(String string)
 	{
-		return controller.get(ObjectKey.getKeyFor(String.class, "*" + string)) != null;
+		return controller.containsKey(ObjectKey.getKeyFor(String.class, "*" + string));
```
thpr: IF the value can be set to null, these are not equivalent.  Have you checked for that case?

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -8109,14 +8092,7 @@ private boolean isCrossClassSkill(Skill sk, PCClass pcc)
...
 	{
-		for (PCClass cl : getClassSet())
-		{
-			if (isCrossClassSkill(sk, cl))
-			{
-				return true;
-			}
-		}
-		return false;
+		return getClassSet().parallelStream().anyMatch(v -> isCrossClassSkill(sk, v));
```
thpr: thought we weren't doing parallel?

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -1158,17 +1158,7 @@ public int getEquipSetNumber()
...
-		for (Equipment eq : aList)
-		{
-			if (aString.equalsIgnoreCase(eq.getName()))
-			{
-				match = eq;
-			}
-		}
-
-		return match;
+		return aList.stream().filter(v -> aString.equalsIgnoreCase(v.getName())).findFirst().get();
```
thpr: Logic change.  the FOR loop does LAST, the stream does first

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -8105,14 +8098,7 @@ public void setTemplatesAdded(CDOMObject po, PCTemplate pct)
...
 	{
-		for (PCClass cl : getClassSet())
-		{
-			if (isClassSkill(cl, sk))
-			{
-				return true;
-			}
-		}
-		return false;
+		return getClassSet().parallelStream().anyMatch(v -> isClassSkill(v, sk));
```
thpr: parallel here too

grimreaper: fixed

PR #1515, opened 2016-08-31
resize does not depend on the pcc

grimreaper:


code/src/java/pcgen/core/Globals.java
```diff
@@ -681,35 +681,20 @@ else if (baseSize > finalSize)
...
-		for (final String rType : resizeTypeList)
-		{
-			if (upperTypeList.contains(rType.toUpperCase()))
-			{
-				return true;
-			}
-		}
-
-		return false;
+		return typeList.parallelStream().map(String::toUpperCase).anyMatch(resizeTypeList::contains);
```
cpmeister: I would prefer if you didn't using the streaming api for this, or at the very least don't parallelize it. Parallelization isn't useful in situations where the list size is small, in fact it is often slower

grimreaper: From an original reading of the documentation I thought that parallelStream was not always parallel (it merely allowed one to configure it elsewhere). Reading it again and I was wrong.

PR #1517, opened 2016-08-31
further decouple PlayerCharacter from the rest of the system (part 2)

grimreaper:
Depends on #1516 - this continues the work to find PlayerCharacter and remove it from various parts.

### DIFF
--- a/code/src/test/pcgen/AbstractCharacterTestCase.java
+++ b/code/src/test/pcgen/AbstractCharacterTestCase.java
@@ -43,7 +43,7 @@ import plugin.lsttokens.testsupport.BuildUtilities;
  * @author frugal@purplewombat.co.uk
  */
 @SuppressWarnings("nls")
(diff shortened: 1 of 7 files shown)


code/src/test/pcgen/AbstractCharacterTestCase.java
```diff
@@ -52,21 +52,21 @@
...
 	protected SizeAdjustment gargantuan;
-	protected SizeAdjustment huge;
+	private SizeAdjustment huge;
 	protected SizeAdjustment large;
 	protected SizeAdjustment medium;
 	protected SizeAdjustment small;
-	protected SizeAdjustment tiny;
+	private SizeAdjustment tiny;
 	protected SizeAdjustment diminutive;
 	protected SizeAdjustment fine;
```
cpmeister: I don't think these should be changed

PR #1518, opened 2016-09-01
Clean up initiative code

grimreaper:
While this may overlap with the GUI rewrite a bit this code intersects with the PlayerCharacter code in somewhat indirect ways. I ended up here, so clean it up.

### DIFF
--- a/code/src/java/plugin/initiative/gui/Initiative.java
+++ b/code/src/java/plugin/initiative/gui/Initiative.java
@@ -46,6 +46,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Vector;
+import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


code/src/java/plugin/initiative/gui/Initiative.java
```diff
@@ -2389,7 +2249,7 @@ else if (returnVal == SavingThrowDialog.FAIL_OPTION)
 			combatantUpdated(cbt);
 			State newStatus = cbt.getStatus();
 
-			if (!oldStatus.equals(newStatus) && newStatus == State.Dead)
+			if ((oldStatus != newStatus) && (newStatus == State.Dead))
```
thpr: I think this is debatable whether this is a best practice to go away from object equality.

PR #1526, opened 2016-09-01
update gradle to latest 2.x series

grimreaper:


cpmeister:
why are you deleting pcgen.sh?

grimreaper:
uuuuh, I thought I was recreating it. One sec

cpmeister:
wait, nvm I think it should be deleted

Zaister:
pcgen.sh is being deleted? why?

LegacyKing:
Seconded. pcgen.sh is used by our Linux users to start the program. It
should be built as part of the compile/build process.

PR #1529, opened 2016-09-01
clean up some code

grimreaper:
- use diamond operator where possible
- avoid casting where possible
- use array initializer where possible

(three separate commits so that they can be reverted)

### DIFF
--- a/code/src/java/pcgen/cdom/base/CDOMObject.java
+++ b/code/src/java/pcgen/cdom/base/CDOMObject.java
@@ -148,7 +148,7 @@ public abstract class CDOMObject extends ConcretePrereqObject implements
 
 	public final Set<IntegerKey> getIntegerKeys()
 	{
-		return integerChar == null ? Collections.<IntegerKey> emptySet()
+		return integerChar == null ? Collections.emptySet()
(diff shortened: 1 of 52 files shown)


code/src/java/pcgen/cdom/base/CDOMObject.java
```diff
@@ -230,7 +230,7 @@ public final Formula remove(FormulaKey key)
 
 	public final Set<FormulaKey> getFormulaKeys()
 	{
-		return formulaChar == null ? Collections.<FormulaKey> emptySet()
+		return formulaChar == null ? Collections.emptySet()
```
thpr: Why destroy the generics hint here?  This will introduce warnings back into the code...
I don't like this set of changes

grimreaper: AFAICT this won't generate a new warning since the type is inferred but I might not understand Java / noticed the warning?

thpr: I retract my concern.  Java 8 is smarter than previous versions and can correctly derive the type without a warning.

PR #1535, opened 2016-09-01
remove useless boxing

grimreaper:


code/src/java/pcgen/gui2/facade/Gui2InfoFactory.java
```diff
@@ -859,15 +859,15 @@ private HtmlInfoBuilder getEquipmentHtmlInfo(Equipment equip)
 
 		Integer a = EqToken.getMaxDexTokenInt(pc, equip);
 
-		if (a.intValue() != Constants.MAX_MAXDEX)
```
thpr: So the question on these (where it's == or !=) is do compilers unconditionally error on this situation of Integer == Integer?  They haven't in the past, and this is where automated boxing assumptions get dangerous... if the Constant is changed, this code is vulnerable.  Therefore removing the .intValue() increases the fragility of this code.  I think this is another point for discussion.

grimreaper: They do not unconditionally error. I'm confused about maintaining the .intValue() helps anything though? It has identical semantics.

thpr: new Integer(3) == new Integer(3) returns false.
So if a is an object (Integer) and you remove the .intValue() call then in the future if Constants.MAX_MAXDEX is changed an object, then the behavior of this method changes, as the two values are tested for instance equality, not value equality.  So there is a case with automated unboxing that actually is higher risk of producing problems, which is why I point is out.  The point here is that the .intValue() in this case is not actually useless.  It is, in fact, defensive programming, and in my opinion should be left alone.

grimreaper: should we replace it with Objects.equals() or AssertEquals()? IMHO relying on explicit boxing for this kind of subtle effect is asking for issues.

PR #1537, opened 2016-09-01
Replace dedicated SpellLoader class with GenericLoader

Zaister:


grimreaper:
LGTM

Zaister:
This changes spells from being stored in a global map to being stored in the reference context, like the other objects.

PR #1541, opened 2016-09-01
Skill Focus => affect Constitution based skills

milcondoin:
The Skill Focus feat can now affect skills based on CON.
There is no rule in Pathfinder which disallows that.


LegacyKing:
1. The red x means this fix breaks tests i.e. 

    org.junit.ComparisonFailure: Errors encountered while loading Pathfinder RPG Core Supplements expected:<[]> but was:<[Error: No Skill objects of TYPE=Constitution were loaded but were referred to in the data]>
        at pcgen.persistence.lst.DataLoadTest.testLoadSources(DataLoadTest.java:169)

2. Even if I accepted this, your skills would need more work to even function properly.

The correct fix on your end would be to change the CHOOSE to just have 'TYPE=Base', and that's it. No errors, and no fuss. 

The normal method is to include a BONUS:SKILLS|%LIST|3|TYPE=SkillFocus. I would make each skill have an ability and you'd select that from a list... and the bonuses would be granted from the choice. It would be expandable and homebrew friendly.

LegacyKing:
Improved version merged in... closing.

PR #1542, opened 2016-09-02
bump spring to latest release (4.3.2)

grimreaper:
depends on #1538

it seems that this Just Works acc' to to my resting, but merge after #1538 to make it easier to bisect errors I may have missed


thpr:
Needs conflict resolution

PR #1545, opened 2016-09-02
removed a bunch of unused code from meta cdom

grimreaper:


code/src/java/pcgen/cdom/meta/ConvertingFacetView.java
```diff
@@ -65,25 +64,18 @@ public String getDescription()
 	}
 
 	@Override
-	public boolean represents(Object src)
```
thpr: Unfortunately, you are now altering code that has private branches trying to use such code.  Please don't remove this from the interface.

thpr:
This one in my mind should just be closed.
pcgen.meta is relatively new code and I have things in private branches that depend on these items... so I don't think it deserves a deep cleanup only for me to have to reimport this code into a handful of branches.

PR #1546, opened 2016-09-02
Simplify dice even further

grimreaper:
This eliminates a level of abstraction that is no longer needed


thpr:
This is dramatic enough, and if you are intending to use it strategically, it should be checked in with unit tests for this subsystem.

code/src/java/gmgen/plugin/dice/NSidedModifiedDieConfig.java
```diff
@@ -20,34 +20,25 @@
...
-		};
 	}
 
 	@Override
 	public int roll()
 	{
-		return counter.totalCount(
-				ResultModifier.modify(modifiers)
-		);
+		return IntStream.generate(() -> Die.rand.nextInt(sides) + 1).map(v -> v + bias).limit(n).sum();
```
thpr: Format to separate rows, please

grimreaper:
This primarily /undoes/ a leve of abstraction I introduced. Agred about unit tests. I'll add some.

build.gradle
```diff
@@ -165,6 +165,8 @@ dependencies {
 
     testCompile group: 'junit', name: 'junit', version:'4.12'
     testCompile group: 'xmlunit', name: 'xmlunit', version:'1.3'
+    testCompile group: 'org.hamcrest', name: 'hamcrest-all', version: '1.3'
```
thpr: I have to admit, I'm not a fan of things just being added to build.gradle without a discussion.  I think we need more broad buy-in that things will be widely used and are value add to the project before they are just adopted

grimreaper:
this test error confuses me, especially since it doesn't seem to be flaky

code/src/java/gmgen/plugin/dice/NSidedModifiedDieConfig.java
grimreaper: done

thpr: Still seeing this as one line...

PR #1547, opened 2016-09-02
getComparator is never used

grimreaper:


code/src/java/pcgen/cdom/enumeration/SkillsOutputOrder.java
```diff
@@ -33,49 +30,9 @@
...
-		}
-	};
-
-	public abstract SkillComparator getComparator(PlayerCharacter pc);
+	NAME_ASC,
+	NAME_DSC,
+	TRAINED_ASC,
+	TRAINED_DSC,
+	MANUAL;
 }
```
thpr: So this is a huge red flag that we have an enumeration that now has no behavior associated with different output orders.  This should be left alone and flagged as a UI bug I think... or it needs a deep UI code review to understand what is going on elsewhere to do this work

grimreaper: Agreed on the red flag issue. See SkillDisplay.java line 141 public static void resortSelected(PlayerCharacter pc, SkillsOutputOrder sortSelection)
a cursory look seems to show that this is doing the work, but I might be wrong.

That said, it might be better to keep the comparator with the enum and removed from resortSelected.

thpr: I would prefer the enums have the behavior, even if it means two methods.  Eliminating the switch in resortSelected would be a good thing.

grimreaper: redid this PR to eliminate the switch 🌀

PR #1549, opened 2016-09-02
simplify getInstance

grimreaper:


cpmeister:
I don't think this PR is beneficial.

PR #1551, opened 2016-09-02
convert from Junit 3 -> Junit 4 properly

grimreaper:
I got really confused when I saw the @Test annotations along with an extension from TestCase.
Begin converting entirely to junit 4


thpr:
No concerns from me - this should be fixed by now... but yes, there is history here - a bug in one of the IDEs back a few years preventing running the tests from within the IDE unless they were Junit 3 (whereas we swapped the distro to Junit4)

PR #1552, opened 2016-09-02
Fix Integration Tests - Revert "Simplified test dependencies"

grimreaper:
This reverts commit 3bc19a5cb47f68eaaebf6283a67e512202f03c81.


grimreaper:
note: this only makes the tests compile - it doesn't actually fix the failures. I'm working on that now

PR #1554, opened 2016-09-03
make compilation faser

grimreaper:
see https://docs.gradle.org/current/dsl/org.gradle.api.tasks.compile.JavaCompile.html for details

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -131,6 +131,7 @@ task copyMasterSheets(type: Copy) {
 
 compileJava {
     it.dependsOn 'copyMasterSheets'
+    options.incremental = true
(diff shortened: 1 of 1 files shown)


thpr:
The challenge is I'm not sure this is purely safe... what specifically is this impacting? all builds? We need to ensure Jenkins stays a full build, since interface changes cause errors on incremental builds, and with the state of some sections of PCGen that can be an issue.

grimreaper:
This only affects compileJava, not compiling tests. That said, I don't know how to tell jenkins compared to the regular env in gradle. Is there some flag to check?

PR #1558, opened 2016-09-03
simplify some test code

grimreaper:


thpr:
Needs import cleanup

code/src/utest/pcgen/base/formula/AddingFormulaTest.java
```diff
@@ -55,30 +59,34 @@ public void testEquality()
...
 	{
 		AddingFormula f = new AddingFormula(-2);
-		assertEquals(3, f.resolve(Integer.valueOf(5)).intValue());
-		assertEquals(-8, f.resolve(Double.valueOf(-6.7)).intValue());
+		assertEquals(3, f.resolve(5).intValue());
+		assertEquals(-8, f.resolve(-6.7).intValue());
 		testBrokenCalls(f);
 	}
 
+	@Test
```
thpr: I don't think this should be a test... it's a utility function and probably should be renamed to not start with "test"

code/src/utest/pcgen/base/formula/SubtractingFormulaTest.java
```diff
@@ -56,30 +60,34 @@ public void testEquality()
...
 	{
 		SubtractingFormula f = new SubtractingFormula(-2);
-		assertEquals(7, f.resolve(Integer.valueOf(5)).intValue());
-		assertEquals(-4, f.resolve(Double.valueOf(-6.7)).intValue());
+		assertEquals(7, f.resolve(5).intValue());
+		assertEquals(-4, f.resolve(-6.7).intValue());
 		testBrokenCalls(f);
 	}
 
+	@Test
```
thpr: I don't think this should be a test... it's a utility function and probably should be renamed to not start with "test"

PR #1560, opened 2016-09-03
add a few functional interface annotations

grimreaper:
this doesn't add it globally - only where it seems to make sense


thpr:
I'm not sure of the value of this?

grimreaper:
Its a marker interface: "a functional interface has exactly one abstract method." Its used in some static analysis and stops compilation if a second non-abstract method is added.

PR #1565, opened 2016-09-03
simplify WriteDirectoryPanel

grimreaper:


code/src/java/pcgen/gui2/converter/panel/WriteDirectoryPanel.java
```diff
@@ -52,28 +51,14 @@
...
-        @Override
-		public boolean accept(File parentDir, String fileName)
+		if (StringUtils.endsWithIgnoreCase(fileName, ".pcc"))
 		{
-			/*
-			 * This is a specific "hack" in order to speed loading when
-			 * in a development (Subversion-based) environment - Tom
-			 * Parker 1/17/07
-			 */
-			if (".svn".equals(fileName))
```
thpr: You're removing logic here, why?

thpr: Oh, never mind, ha, not using svn anymore

code/src/java/pcgen/gui2/converter/panel/WriteDirectoryPanel.java
```diff
@@ -133,9 +113,9 @@ public boolean returnAllowed()
 	public void setupDisplay(JPanel panel, final CDOMObject pc)
 	{
 		panel.setLayout(layout);
-		JLabel label = new JLabel(
+		Component label = new JLabel(
 				"Please select the Directory where Converted files should be written: ");
-		JButton button = new JButton("Browse...");
+		AbstractButton button = new JButton("Browse...");
```
thpr: I think this is another point for discussion.  I get why we use List and Set rather than ArrayList, et al.  But I'm not sure I'm fond of always jumping down to Component and AbstractButton...

grimreaper: in general I think the left side of the declaration should have the weakest type possible (excluding Object).

PR #1568, opened 2016-09-03
some simple RandomNamePlugin cleanup

grimreaper:


code/src/java/plugin/doomsdaybook/RandomNamePlugin.java
```diff
@@ -22,15 +22,12 @@
 import gmgen.GMGenSystemView;
 import gmgen.pluginmgr.messages.AddMenuItemToGMGenToolsMenuMessage;
 import gmgen.pluginmgr.messages.RequestAddTabToGMGenMessage;
-
 import java.awt.Component;
 import java.awt.event.ActionEvent;
 import java.io.File;
-
```
thpr: This is not consistent with the rest of our files, and is not consistent with the eclipse formatting we use... so these deleted spaces will be reverted by eclipse users.

grimreaper: Can you point me to a style guide so I can configure my IDE ?

code/src/java/plugin/dicebag/DiceBagPlugin.java
```diff
@@ -141,13 +122,13 @@ public void stop()
     @Override
 	public int getPriority()
 	{
-		return SettingsHandler.getGMGenOption(LOG_NAME + ".LoadOrder", 20);
+		return SettingsHandler.getGMGenOption(DiceBagPlugin.LOG_NAME + ".LoadOrder", 20);
```
thpr: This is an interesting change.  The constant is in this file - why have the fully qualified name?

We also seem to have a mix of tab and space indenting (Override in spaces, public int in tabs) - what is your formatter using?

grimreaper: This change is because it is referencing a static field from a non-static context. I like this stylistically and it makes it easier to move code around.
My formatter is detecting whatever the current file is using I think. I didn't touch the @overide so it remains a weird mix.

PR #1576, opened 2016-09-04
make the PMD rules better, more explicit and more opinionated

grimreaper:
Lets argue about all the rules :)


thpr:
Having now tested this, I'm okay with it, but will let Connor check as well

code/standards/pmdruleset.xml
```diff
@@ -15,23 +15,60 @@
...
+  <rule ref="rulesets/java/basic.xml">
+    <exclude name="AvoidUsingHardCodedIP" />
+  </rule>
+  <rule ref="rulesets/java/braces.xml" />
+  <rule ref="rulesets/java/clone.xml">
+    <exclude name="ProperCloneImplementation" />
+    <exclude name="CloneThrowsCloneNotSupportedException" />
+  </rule>
+  <rule ref="rulesets/java/codesize.xml" >
+    <exclude name="TooManyFields" />
```
thpr: I actually disagree that we should suppress TooManyFields and TooManyMethods

grimreaper: Does this provide value? I've found that the number of fields is rarely an issue, but the complexity and breakdown is. I'll remove the exclusions though (this PR is still strictly better than where we are today)

code/standards/pmdruleset.xml
```diff
@@ -15,23 +15,60 @@
...
+    <exclude name="ConsecutiveLiteralAppends" />
+    <exclude name="AvoidStringBufferField" />
+  </rule>
+  <!--<rule ref="rulesets/java/sunsecure.xml">-->
+  <rule ref="rulesets/java/typeresolution.xml" />
+  <rule ref="rulesets/java/unnecessary.xml">
+    <exclude name="UselessParentheses" />
+  </rule>
+  <rule ref="rulesets/java/unusedcode.xml">
+    <exclude name="UnusedModifier" />
```
grimreaper: that is included by default. Note these are *exclude* rules.

code/standards/pmdruleset.xml
grimreaper: exclusion removed

PR #1577, opened 2016-09-04
Add basic Log4J Logging

grimreaper:
This is really here to add logback.xml and to show how it gets called.

example logs:
image

### DIFF
--- a/code/src/java/pcgen/system/Main.java
+++ b/code/src/java/pcgen/system/Main.java
@@ -20,9 +20,7 @@
  */
 package pcgen.system;
 
-import java.awt.Font;
-import java.awt.FontFormatException;
-import java.awt.GraphicsEnvironment;
+import java.awt.*;
 import java.io.File;
 import java.io.IOException;
 import java.io.PrintWriter;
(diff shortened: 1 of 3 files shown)


code/src/java/pcgen/system/Main.java
```diff
@@ -129,6 +134,19 @@ private static void logSystemProps()
 	}
 
 	/**
+	 * Don't delete. Used for testing logging.
```
thpr: Why have a test in main line code rather than out in test code?

grimreaper: this is mainly there so I could add a line to main() and run. It was very useful for testing that the classpath was set up correctly. This differs from main to test. I don't /really/ need this function. It was just nice to have.

grimreaper:
I'm mainly concerned about if you like the style of how to get the logger. If that's okay the rest is easy to change.

cpmeister:
I don't think that slf4j can be used in PCGen because PCGen uses some custom logging levels that slf4j cannot provide. I think that using Log4j directly is a better approach.

grimreaper:
What kind of custom logging levels? They are typically better suited as adding contextual information? I've seen CONFIG? anything else?

cpmeister:
Take a look at pcgen.util.PCGenLogLevel

PR #1578, opened 2016-09-04
migrate more code from junit 3 -> junit 4

grimreaper:


grimreaper:
this is too much of a sweeping change right now to be worth it. It will happen over time, but changing deep core tests to junit 4 has a number of ramifications and I'd rather deal with them by nibbling at the edges for now.

PR #1582, opened 2016-09-04
combined version bump

grimreaper:
this is the remainder of the version bumps


thpr:
This is already obsolete - let's leave this until we get through the detailed ones

PR #1586, opened 2016-09-04
NEWTAG-607 Data Tables

thpr:
NEWTAG-607 Lookup() function
NEWTAG-608 DATATABLE Campaign Token
NEWTAG-609 TABLE[x,y] format
NEWTAG-610 COLUMN[x] format

### DIFF
--- a/code/src/java/pcgen/cdom/format/table/TableFormatManager.java
+++ b/code/src/java/pcgen/cdom/format/table/TableFormatManager.java
@@ -32,7 +32,7 @@ import pcgen.base.util.ObjectDatabase;
  * that such a column exists in the DataTable, not that it has a specific name
  * or that it is the only column.
  */
-public class TableFormatManager implements FormatManager<DataTable>
+public final class TableFormatManager implements FormatManager<DataTable>
 {
 
 	/**
--- a/code/src/java/pcgen/cdom/format/table/TableFormatManager.java
+++ b/code/src/java/pcgen/cdom/format/table/TableFormatManager.java
@@ -1,0 +1,191 @@
+/*
+ * Copyright 2016 (C) Tom Parker <thpr@users.sourceforge.net>
+ * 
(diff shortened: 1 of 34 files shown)


code/src/java/pcgen/cdom/format/table/TableFormatManager.java
```diff
@@ -0,0 +1,191 @@
...
+	 * {@inheritDoc}
+	 */
+	@Override
+	public boolean equals(Object o)
+	{
+		if (o instanceof TableFormatManager)
+		{
+			TableFormatManager other = (TableFormatManager) o;
+			return lookupFormat.equals(other.lookupFormat)
+				&& resultFormat.equals(other.resultFormat);
```
cpmeister: Unless TableFormatManager is final, this is a potentially erroneous equals() implementation

code/src/utest/pcgen/rules/persistence/TableLoaderTest.java
```diff
@@ -0,0 +1,447 @@
...
+			//Yes
+		}
+		catch (IllegalArgumentException e)
+		{
+			//Yes
+		}
+		catch (IndexOutOfBoundsException e)
+		{
+			//Yes
+		}
```
cpmeister: You expect ALL of these errors?

thpr: Any one of is valid - all it needs to do is fail...

code/src/java/pcgen/cdom/format/table/TableFormatManager.java
thpr: I will change it to final, but I think the question is how much we want to worry about it... that is true almost universally across the entire java.* design, in how they originally implemented .equals()... they didn't use .getClass(), they just tested the local variables.  (The same is basically universally true across our code base.)

PR #1559, opened 2016-09-03
remove javadocs that only make sense with svn keywords

grimreaper:

PR #1587, opened 2016-09-05
Partial removal of @version

grimreaper:
partial application of #1559


thpr:
489 also too many to see at once in github (limit is 300)

grimreaper:
ya, I'm working on splitting this further

PR #1591, opened 2016-09-05
Revert "make tests run faster"

thpr:
This reverts commit bd29db9ef1348547cc4061ad105246192dc502fe.


thpr:
Reverted as this basically locks up my computer when used... making me unable to actually perform any other actions in parallel with a build.  That makes the build basically unusable for me.

PR #1594, opened 2016-09-05
don't be strict about constant names in classes

grimreaper:


thpr:
I don't agree with this - if we want to change this, it needs to end up on the developers list for discussion

grimreaper:
alright - I'll bring this up on the list

PR #1604, opened 2016-09-05
Add visible for testing annotation

grimreaper:
While the VisibleForAnnotation can be added for a few different places I chose to get it from Guava since there are other annotations that may be useful
N.B. this only adds guava as a compile time dependency and does not affect distribution


grimreaper:
actually, I have a better version of this

PR #1607, opened 2016-09-05
Add visible for testing annotation and enable use of findbugs

grimreaper:


thpr:
No, no guava due to size

grimreaper:
Why close this?

grimreaper:
It isn't a distribution dependency. Are you concerned about the compilation dep size as well?

grimreaper:
also, have we considered running proguard on the jar post-build? It might drastically reduce the size.

PR #1613, opened 2016-09-05
Add findbugs to gradle

grimreaper:
This also makes a few other changes: 
- enable some reporting by default. This moves on a path towards *erroring* when the reporting fails which is the only real way to ensure we don't regress.


grimreaper:
FWIW I updated this diff

PR #1626, opened 2016-09-07
clean up purchase mode frame

grimreaper:


grimreaper:
I know this being replaced, but I was looking for a bug here and ended up just making this

PR #1627, opened 2016-09-07
convert many more tests to junit 4

grimreaper:
and clean up some tests I didn't convert in prep for the next round

### DIFF
--- a/code/src/testcommon/compare/WeakReferenceInequality.java
+++ b/code/src/testcommon/compare/WeakReferenceInequality.java
@@ -24,9 +24,9 @@ public class WeakReferenceInequality implements InequalityTest<WeakReference<?>>
 {
 
     @Override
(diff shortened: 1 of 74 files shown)


code/src/testcommon/compare/WeakReferenceInequality.java
```diff
@@ -24,9 +24,9 @@
 {
 
     @Override
-	public String testInequality(WeakReference<?> wr1, WeakReference<?> wr2, InequalityTester t, String location)
+	public String testInequality(WeakReference<?> t1, WeakReference<?> t2, InequalityTester t, String location)
```
thpr: Why this change?

grimreaper: make it match the function it is overriding

PR #1633, opened 2016-09-08
move some code around and in doing so, remove two files

grimreaper:


thpr:
Nope, sorry

thpr:
So the issue here is that this change is very short sighted.  There needs to be a longer discussion here about why some of these things exist and what is really long term plans around the involved classes

PR #1634, opened 2016-09-08
more test cleanup: junit and cleanup

grimreaper:


code/src/java/pcgen/persistence/PersistenceManager.java
```diff
@@ -65,7 +65,7 @@ public static PersistenceManager getInstance()
 	 * 
 	 * CODE-1889 to remove use of this method
 	 */
-	public void setChosenCampaignSourcefiles(List<URI> l)
+	public static void setChosenCampaignSourcefiles(List<URI> l)
```
thpr: This is now a completely confusing class design.  You still have a getInstance() method, but it returns an instance with no non-static methods?  I know the original design was to have different managers, but that design isn't being pursued, but if you are going to get rid of non-static methods, then getInstance() makes no sense.

grimreaper: I want to eliminate that, but in a separate PR. (I should have actually separated this one into two). In general I'm trying to keep my PRs smaller.

PR #1637, opened 2016-09-08
partial cleanup

grimreaper:


code/src/java/pcgen/gui2/SplashScreen.java
```diff
@@ -44,6 +46,7 @@
 public class SplashScreen extends JWindow implements PCGenTaskListener
 {
 
+	private static final long serialVersionUID = 5186500840685769783L;
```
thpr: Why?  What value does this REALLY provide?

code/src/java/pcgen/persistence/SourceFileLoader.java
```diff
@@ -1112,7 +1114,7 @@ private void logError(String message, Throwable e)
 	 * @param sources
 	 *            The list of potential sources to be added.
 	 */
-	private void addQualifiedSources(Campaign c, ListKey<CampaignSourceEntry> lk)
+	private void addQualifiedSources(CDOMObject c, ListKey<CampaignSourceEntry> lk)
```
thpr: This is *designed for* Campaigns, not CDOMObjects in general.  So this is actually now communicating incorrectly to a user of the method.  It is communicating false information by being made more generic.

code/src/java/pcgen/gui2/SplashScreen.java
grimreaper: mostly it shuts up some warnings from PMD. Otherwise its not useful. I guess I could go and try and remove the warning generator instead.
It might also be correct to make the window not serializable.

PR #1638, opened 2016-09-09
Updated Champions Of Purity / APG

ProtoJazz:
Fixed a few issues I noticed while building an Oracle. Unless Im understanding things wrong Oracles should have access to any spell not marked "Clerics (only)". I hope so, because burst of radiance is the best spell my oracle has. 

I also fixed up the damage calculation for Death’s Touch(Bones Oracle Mystery)

` Death’s Touch (Su): You can cause terrible wounds to appear on a creature with a melee touch attack. This attack deals 1d6 points of negative energy damage +1 point for every two oracle levels you possess. If used against an undead creature, it heals damage and grants a +2 channel resistance for 1 minute. You can use this ability a number of times per day equal to 3 + your Charisma modifier.`

It was calculating the damage as Oracle level, not Oracle /2. Which at first was great, until I realized it wasn't accurate. Though being half as strong explains why its not a must have spell.


LegacyKing:
Standard practice is to open a Jira so we have accurate release notes - want you to get credit after all.

Zaister:
I don't think it makes sense to add "Oracle" to the CLASSES tag in spells, as oracles use the cleric list, and so should be covered by "Cleric" in that tag. Note that, for examples, the Core Rulebook cleric spells are never modded to have Oracle added here.

I propose that this change to the CLASS tags is reverted.

Zaister:
This makes me wonder, actually, why we always have both wizard _and_ sorcerer in the CLASS tags. One of them would be enough. Probably just because it has always been like that.

Distant-Scholar:
I know there were 3.5 spells that were sorcerer only or wizard only.  I
can't remember if Pathfinder has any.

DS/DL


On Sun, Sep 11, 2016, at 05:47 AM, Stefan Radermacher wrote:
> This makes me wonder, actually, why we always have both wizard *and*
> sorcerer in the CLASS tags. One of them would be enough. Probably just
> because it has always been like that.
> — You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub[1], or mute the
> thread[2].
>


Links:

  1. https://github.com/PCGen/pcgen/pull/1638#issuecomment-246173916

ProtoJazz:
@Zaister In this case the spells were not select-able on an oracle without adding Oracle to the Classes tag.

PR #1641, opened 2016-09-10
fix bug in TabbedPane

grimreaper:


code/src/java/pcgen/gui2/tools/SpinningTabbedPane.java
```diff
@@ -835,16 +836,18 @@ private int previous(int current, int[] indices)
                     secundum = penultimatum = indices[1];
                     ultimatum = indices[2];
 
+                    break;
```
grimreaper: this is the bug

code/src/java/pcgen/gui2/tools/SpinningTabbedPane.java
```diff
@@ -38,17 +33,36 @@
...
+
 /**
  * <code>SpinningTabbedPane</code>.
  *
  * @author &lt;a href="binkley@alumni.rice.edu"&gt;B. K. Oxley (binkley)&lt;/a&gt;
  *
  * @see JTabbedPane
  */
-public class SpinningTabbedPane extends JTabbedPane
+class SpinningTabbedPane extends JTabbedPane
```
cpmeister: This counts as a utility class, and as such it should remain public

grimreaper: fixed

PR #1642, opened 2016-09-10
split PlayerChar and PlayerCharImpl

grimreaper:
This is primarily to make it easier to understand the data flow around PlayerCharacter. It splits it into an interface and an implementation. While they should both die, it is a lot easier to find opportunities if the interface is used separately from the Implementation.


grimreaper:
nevermind: I posted the wrong version of this diff

PR #1644, opened 2016-09-10
dec -> dec

grimreaper:
avoid .equals() on BigDecimal

### DIFF
--- a/code/src/java/pcgen/core/PlayerCharacter.java
+++ b/code/src/java/pcgen/core/PlayerCharacter.java
@@ -7764,7 +7764,7 @@ public class PlayerCharacter implements Cloneable, VariableContainer
 
 	public void adjustAbilities(final Category<Ability> aCategory, final BigDecimal arg)
 	{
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -7765,7 +7765,7 @@ public boolean doLevelAbilities()
 
 	public void adjustAbilities(final Category<Ability> aCategory, final BigDecimal arg)
 	{
-		if (arg.equals(BigDecimal.ZERO))
+		if (arg.compareTo(BigDecimal.ZERO) == 0)
```
cpmeister: Why? I think this makes the code less easy to read.

grimreaper: The method BigDecimal.equals() takes scale into consideration. For example BigDecimal("0.00").equals(BigDecimal.ZERO) == false. I don't know all the callsites here, but seeing .equals() on a BigDecimal in code smell

PR #1647, opened 2016-09-10
useless javadoc -> no javadoc

grimreaper:
remove @see where it is useless


code/src/java/pcgen/cdom/facet/LevelInfoFacet.java
```diff
@@ -39,8 +39,7 @@
 	 * Overrides the default behavior of AbstractListFacet, since we need to
 	 * ensure we are storing the PCLevelInfo objects in an ordered list (since
 	 * we are implicitly storing the level of the PCLevelInfo by its location)
-	 * 
-	 * @see pcgen.cdom.facet.base.AbstractListFacet#getComponentSet()
```
cpmeister: I think some of these are not useless @see in the docs. In the case that @see is the only thing I can agree that it is useless, but not in situations where there is other documentation alongside it.

PR #1657, opened 2016-09-10
explicit conversion where loss might exist

grimreaper:


thpr:
The big concern I have here is that these changes are functional changes to BONUS.  This may (or may not) be correct, as the BONUS system has some weird (truly weird) rounding scenarios.  I touch the subsystem as little as possible for that reason

grimreaper:
I expect that none of these are functional changes and merely making explicit what is currently implicit. That said, if you feel this introduces risk without value feel free to close

PR #1660, opened 2016-09-10
public -> protected in abstract classes

grimreaper:


cpmeister:
What is the point of this?

grimreaper:
stylistic

PR #1663, opened 2016-09-10
cast -> weak cast

grimreaper:


cpmeister:
While in principle yes, you shouldn't cast more than you need to, at the same time you shouldn't turn
`if(obj instanceof SomeObj) return ((SomeObj)obj).foo()`
into
`if(obj instanceof SomeObj) return ((ParentObj)obj).foo())`
This inherently makes the code buggy when the original intention was clear.
Also, sometimes you want to do specific casts so you don't have to import more classes.

PR #1665, opened 2016-09-10
avoid using sringbuffer where its not helpful

grimreaper:


cpmeister:
I want to keep StringBuilder for stylistic purposes

Zaister:
I think compiler uses StringBuilder internally anyway if you write code like

    String s = "foo" + "bar";

PR #1686, opened 2016-09-10
fix minor validation bug in compatability

grimreaper:
don't think this is anything other than a newly working validation, but maybe I missed where else this would get triggered.

### DIFF
--- a/code/src/java/pcgen/io/Compatibility.java
+++ b/code/src/java/pcgen/io/Compatibility.java
@@ -12,18 +12,26 @@ import pcgen.rules.context.LoadContext;
 import pcgen.rules.persistence.token.ParseResult;
 import pcgen.util.Logging;
 
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
 public final class Compatibility
 {
 
-	public static PCTemplate getTemplateFor(PCTemplate template,
-			String templateKey, String feat)
+	private Compatibility()
+	{
+	}
+
+	@Nullable
+	static PCTemplate getTemplateFor(PCTemplate template,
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/io/Compatibility.java
```diff
@@ -112,15 +122,15 @@ else if (!max.equals(min))
 			return null;
 		}
 		int closeParenLoc = first.lastIndexOf(')');
-		if (openParenLoc == -1)
+		if (closeParenLoc == -1)
```
grimreaper: this is the actual bug

thpr: Interesting that was missed originally.  This was originally coded for the 5.12->5.14 conversion, so probably not really used anymore

grimreaper: should it just be removed? I'm not sure how much of this is still useful

code/src/java/pcgen/io/Compatibility.java
```diff
@@ -132,8 +142,7 @@ else if (!max.equals(min))
 			return null;
 		}
 		context.commit();
-		PersistentTransitionChoice<?> ptc = applied.getListFor(ListKey.ADD).get(0);
-		return ptc;
+		return applied.getListFor(ListKey.ADD).get(0);
```
thpr: I'm concerned this won't compile.  There have been similar changes made in other places that when inlined cause various compilers to complain... At what level has this been tested?

code/src/java/pcgen/io/Compatibility.java
thpr: It's still needed

PR #1687, opened 2016-09-10
fix a minor doc bug ("1" -> "2")

grimreaper:


code/src/java/pcgen/io/PCGIOHandler.java
```diff
@@ -597,7 +598,7 @@ private void resolveDuplicateEquipmentSets(PlayerCharacter currentPC)
 	 * Pick one of two equipment sets sharing a path to be moved to a new path. 
 	 * Only non containers will be moved to avoid issues with contents.   
 	 * @param equipSet1 The first equipment set at a path.
-	 * @param equipSet1 The second equipment set at a path.
```
grimreaper: this is the bug

PR #1688, opened 2016-09-10
update our checkstyle config

grimreaper:
This add checkstyle to the build path. 
- update checkstyle to the latest version so that documentation applies
- be stricter about a few things
- begin to error rather than warn on things for which we have no current problems
- add checkstyle to the build path to help reduce the chance of adding regressions
- relax a few conventions to more accurately match what seems to be project style.

FWIW I ask that you avoid holding this review on the choice of specific style rules. Those can easily be changed. I want to make sure that we're on a path to making use of the tools that we can.

### DIFF
--- a/code/standards/checkstyle.xml
+++ b/code/standards/checkstyle.xml
@@ -9,6 +9,7 @@
     Description: none
 -->
 <module name="Checker">
+  <property name="charset" value="UTF-8"/>
   <property name="severity" value="warning"/>
   <module name="TreeWalker">
     <property name="cacheFile" value="checkstyle-cachefile"/>
@@ -17,13 +18,17 @@
       <property name="allowUndeclaredRTE" value="true"/>
       <property name="allowMissingParamTags" value="true"/>
       <property name="allowMissingReturnTag" value="true"/>
+      <property name="allowMissingThrowsTags" value="true" />
       <property name="allowMissingPropertyJavadoc" value="true"/>
     </module>
     <module name="JavadocType"/>
     <module name="JavadocVariable">
       <property name="scope" value="public"/>
     </module>
-    <module name="ConstantName"/>
+    <module name="ConstantName">
+      <!--off for now to direct focus, but apply later-->
+      <property name="applyToPrivate" value="false" />
+    </module>
     <module name="LocalFinalVariableName"/>
     <module name="LocalVariableName"/>
     <module name="MemberName"/>
@@ -32,7 +37,9 @@
     <module name="ParameterName"/>
     <module name="StaticVariableName"/>
     <module name="TypeName"/>
-    <module name="AvoidStarImport"/>
+    <module name="AvoidStarImport">
+      <property name="allowStaticMemberImports" value="true"/>
+    </module>
     <module name="IllegalImport"/>
     <module name="RedundantImport"/>
     <module name="UnusedImports"/>
@@ -45,7 +52,6 @@
     <module name="MethodParamPad"/>
     <module name="NoWhitespaceAfter"/>
     <module name="NoWhitespaceBefore"/>
-    <module name="OperatorWrap"/>
     <module name="ParenPad"/>
     <module name="TypecastParenPad"/>
     <module name="WhitespaceAfter"/>
@@ -53,7 +59,7 @@
       <property name="tokens" value="ASSIGN"/>
     </module>
     <module name="ModifierOrder"/>
-    <module name="AvoidNestedBlocks"/>
+    <module name="AvoidNestedBlocks" />
     <module name="EmptyBlock"/>
     <module name="LeftCurly">
       <property name="option" value="nl"/>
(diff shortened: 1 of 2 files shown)


grimreaper:
bah I ran this against main but not the others. Fixing now

thpr:
Needs list discussion as this is changing standards

grimreaper:
> Needs list discussion as this is changing standards
I was literally writing the list message as you were writing this :-p

code/standards/checkstyle.xml
```diff
@@ -13,13 +14,16 @@
...
+      <property name="allowMissingThrowsTags" value="true" />
       <property name="allowMissingPropertyJavadoc" value="true"/>
     </module>
     <module name="JavadocType"/>
     <module name="JavadocVariable">
       <property name="scope" value="public"/>
     </module>
-    <module name="ConstantName"/>
+    <module name="ConstantName">
+      <property name="applyToPrivate" value="false" />
```
thpr: Why not enforce this?

code/standards/checkstyle.xml
```diff
@@ -75,13 +83,21 @@
     <module name="FinalClass"/>
     <module name="HideUtilityClassConstructor"/>
     <module name="InterfaceIsType"/>
-    <module name="VisibilityModifier"/>
+    <module name="VisibilityModifier">
+      <property name="protectedAllowed" value="true" />
+      <property name="allowPublicFinalFields" value="true" />
```
thpr: Why?  I know I have a few classes where I've bent the rules, but that doesn't make them good design :)
the danger with public final is if that item is mutable, OO was destroyed in the process.  Shouldn't this be suppressed on an individual basis when it's reviewed by a person, rather than globally not enforced?

code/standards/checkstyle.xml
```diff
@@ -41,22 +47,24 @@
...
     <module name="ParenPad"/>
     <module name="TypecastParenPad"/>
     <module name="WhitespaceAfter"/>
     <module name="WhitespaceAround">
       <property name="tokens" value="ASSIGN"/>
     </module>
     <module name="ModifierOrder"/>
-    <module name="AvoidNestedBlocks"/>
+    <module name="AvoidNestedBlocks">
+      <property name="allowInSwitchCase" value="true" />
```
thpr: Why?

code/standards/checkstyle.xml
grimreaper: A number of classes already violate this and I don't want to introduce a motive for churn. In addition we currently have tens of thousands of warnings: Ideally we could cut that number into one that can be meaningfully audited.

code/standards/checkstyle.xml
grimreaper: I'd rather have a separate check for "mutable public final" fields. If that check can't exist then it may make sense to turn this off.

code/standards/checkstyle.xml
grimreaper: personal style. Mostly want to allow code like 
case 2:
{
  code here;
}

as the meta comment above notes: if the only concern here are the particular style rules just let me know what to change them to. In the general sense I tried to capture existing pcgen style.

code/standards/checkstyle.xml
thpr: Okay, so should we put this with a comment that says "off for now to focus on stuff later, but apply later"?

code/standards/checkstyle.xml
thpr: My personal opinion is switch is generally not what we should be using - those should strongly be considered to be situations where we should ask if something should actually have a method.  A really complicated switch is a code smell, IMHO.  So I personally would not want to encourage large switch statements by having blocks.

code/standards/checkstyle.xml
grimreaper: done

code/standards/checkstyle.xml
grimreaper: removed

code/standards/checkstyle.xml
thpr: Setting this to needs review - as we need a specific plan for mutable public final fields

grimreaper: Is there something I can do to avoid this for now? Perhaps leave the exception out, and add supressions when we're ready to actually error out? I don't want to hold off on the entire diff due to a single rule :/

thpr: Leave the exception out for now

grimreaper: done.

grimreaper:
I think is ready for merge now. Removed the exception and cleaned up the merge conflicts

code/standards/checkstyle.xml
```diff
@@ -80,13 +87,20 @@
     <module name="FinalClass"/>
     <module name="HideUtilityClassConstructor"/>
     <module name="InterfaceIsType"/>
-    <module name="VisibilityModifier"/>
+    <module name="VisibilityModifier">
+      <property name="protectedAllowed" value="true" />
+      <property name="allowPublicImmutableFields" value="true" />
```
thpr: This should be removed - too much risk as per previous comment

PR #1689, opened 2016-09-10
remove jdepend

grimreaper:
jdepend provides nearly zero value to pcpgen. Its useful for doing certain kinds of academic analysis; however since we're still not at the point of erroring out for true bugs or serious style errors don't even bother producing jdepend reports

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,6 @@ plugins {
     id 'findbugs'
     id 'java'
     id 'maven'
-    id 'jdepend'
     id 'pmd'
 
     id "de.undercouch.download" version "3.1.1"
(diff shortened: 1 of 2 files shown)


thpr:
Academic?  Not really.  This is an architecture discussion for the list.  (This isn't to say we can error on it today, but there are places in the code that are very much enforced as a DAG as best I can)

grimreaper:
If you're just looking to enforce DAG like properties on specific hierarchies other tools can do this (for example PMD).
My reference to academic use has to do with the various coupling metrics (Ca, Ce) and abstraction metrics (A, I, and D) which have minimal predictive value of bugs on their own but combined with other sources of data can be useful.

Do you routinely look at the output of jdepend? Can we instead move the logic into pmd rules that we can error on?

thpr:
Where specifically in PMD?  The stuff in "Coupling" is effectively useless due to false positives - it hates many instances of good design.

thpr:
Please fix the conflicts - I will let this merge as I can do this by hand and we can reintroduce later when we want it automated

PR #1691, opened 2016-09-10
explicit conversion

grimreaper:
add explicit conversion where we're losing precision


thpr:
The big concern I have here is that these changes are functional changes to BONUS.  This may (or may not) be correct, as the BONUS system has some weird (truly weird) rounding scenarios.  I touch the subsystem as little as possible for that reason

grimreaper:
this is a duplicate PR - closing in favour of the original

PR #1703, opened 2016-09-10
safe new instance

grimreaper:


cpmeister:
How is this safer?

grimreaper:
I was trying to write this up and realized I can't do it better than the quote from my IDE:

> The newInstance method propagates any exception thrown by the no-arg constructor, including checked exceptions. Use of this method effectively bypasses the compile-time exception checking that would otherwise be performed by the compiler. Replacing such a method call with a call to the java.lang.reflect.Constructor.newInstance() method avoids this problem by wrapping any exception thrown by the constructor in a java.lang.reflect.InvocationTargetException.

PR #1709, opened 2016-09-10
prefer lf line endings

grimreaper:
this makes the default line ending lf (and leaves .bat files as crlf)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,6 @@
 # Handle line endings automatically for files detected as text
 # and leave all files detected as binary untouched.
-* text=auto
+* text=auto eol=lf
 
 #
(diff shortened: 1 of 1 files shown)


thpr:
This is likely to be a problem - and should not be merged without significant discussion
We have had huge line-ending problems in the past since we are on many platforms with many different methods of checking out of git.  Such changes in the past have taken us weeks (if not months) to sort out.  I don't think this is worth the risk.

grimreaper:
I'll bring this up on the list. I suspect that while this introduces some risk for some changes, it reduces risk by making explicit the decision about which line ending to use.

grimreaper:
per list discussion, close.

PR #1711, opened 2016-09-10
Convert some stat items to Number, add Code Controls for Stat

thpr:


thpr:
I would specifically like Connor to look at CharacterFacade and CharacterFacadeImpl, since it changes items from Integer to Number in order to have the old and new forms consistent.  This is a side effect of the new formula system all being Number.class, not specifically to Double or Integer.  Happy to go revisit if we think there needs to be an Integer type to force this to be clean... posted for review not necessarily merge

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -9515,9 +9631,9 @@ public void setStat(PCStat stat, int value)
 		statValueFacet.set(id, stat, value);
 	}
 
-	public Integer getStat(PCStat stat)
```
grimreaper: why this change?

code/src/java/pcgen/core/display/CharacterDisplay.java
```diff
@@ -1536,7 +1536,7 @@ public String getSubstitutionClassName(PCClassLevel lvl)
 
 	public Integer getStat(PCStat stat)
```
grimreaper: this does not mirror the type change above

code/src/java/pcgen/core/PlayerCharacter.java
thpr: To avoid a double-boxing operation.  The old version was guaranteed to return an Integer from the get, so the return of Integer guaranteed no boxing operation.  The new version returns a Number and calls intValue, so it is guaranteed to have an int, so we return int to avoid boxing.  In both cases, we return natively what we know we have, so that if someone else wants exactly what we have, we don't box and then unbox.

code/src/java/pcgen/core/display/CharacterDisplay.java
thpr: Good catch, fixed

code/src/java/pcgen/core/PlayerCharacter.java
thpr: Which, I might add, is one of the reasons I don't turn off automated boxing warnings on my system, and why I explicitly write boxing/unboxing.  Automation would hide the warning and allow you to write methods that would basically be the equivalent of: Integer foo = Integer.valueOf(getStuff().intValue());
  ... where the valueof is implied in the foo assignment and the .intValue is implied in the return from getStuff

cpmeister:
Looks good from what I can tell. But any real issue will be clear very quickly in the UI with this change. If you don't see errors and everything displays properly on the summary table and in the preview sheets then I think everything is good.

thpr:
Closed in favor of new PR

PR #1716, opened 2016-09-11
clean up networking code

grimreaper:


thpr:
Approved once the if/switch question on the list is resolved

code/src/java/plugin/network/NetworkClient.java
```diff
@@ -19,40 +19,40 @@
...
 import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.net.Socket;
 import java.util.StringTokenizer;
 
-public class NetworkClient
+import pcgen.core.SettingsHandler;
+import pcgen.util.Logging;
+
+class NetworkClient
```
thpr: This needs testing.  I think this breaks the plugins so they can no longer be loaded.
I've already had to revert a similar change elsewhere in the plugin system.

grimreaper:
replacing with something else

PR #1717, opened 2016-09-11
more code cleanup

grimreaper:


code/src/java/plugin/modifier/set/AddModifierFactory.java
```diff
@@ -121,16 +122,16 @@ public String getInstructions()
...
 		/**
 		 * The objects to be added to the active set when this AddModifier is
 		 * processed
 		 */
 		private Indirect<T[]> toAdd;
 
-		public AddIndirectArrayModifier(FormatManager<T[]> formatManager,
-			int userPriority, Indirect<T[]> toAdd)
+		private AddIndirectArrayModifier(FormatManager<T[]> formatManager,
+		                                 int userPriority, Indirect<T[]> toAdd)
```
thpr: I'm concerned this indent is a formatting mismatch that may trigger churn

grimreaper: likely is. What should I set my IDE to?

thpr: I have to admit I don't know the exact knob, which is why we have the standards file for formatting.

grimreaper:
killed in favor of other cleanups

PR #1724, opened 2016-09-12
reduce debug information produced by logback

grimreaper:


grimreaper:
see https://github.com/logstash/logstash-logback-encoder/issues/160 for details

PR #1741, opened 2016-09-23
clean up RollingMethods

grimreaper:


code/src/java/pcgen/core/RollingMethods.java
```diff
@@ -436,28 +433,13 @@ else if ((param instanceof Vector) && (curNumberOfParameters == 3))
...
-                }
-                else
+                if ((dRolls > Integer.MAX_VALUE) || (dFaces > Integer.MAX_VALUE))
                 {
-                    result = roll(iRolls, iFaces, keep);
+                    throw new ParseException("Values greater than " + Integer.MAX_VALUE + " not allowed.");
                 }
+                int iRolls = Math.round(((Double) numberOfRolls).intValue());
+                int iFaces = Math.round(((Double) faces).intValue());
+                double result = (keep == null) ? RollingMethods.roll(iRolls, iFaces, numToKeep, reroll, 0) : RollingMethods.roll(iRolls, iFaces, keep);
```
thpr: This has to be bigger than our line length limit...

grimreaper: fixed

thpr:
Still think this file needs reformatting to 90 line length limit

grimreaper:
as discussed the line length was different among different config files. Reformatted the /entire/ file. Leme know if anything else in my IDE should be changed.

thpr:
Now we seem to have line ending differences.... :/

grimreaper:
ugh, I'll fix this later. This is one of the many reasons I'd really like to have a proper entry in gitattributes.

thpr:
how is text=auto not "proper"?  It's there, you may need to ensure your local tools respect it...

PR #1742, opened 2016-09-24
simplify KnownSpellIdentifier

grimreaper:


code/src/java/pcgen/cdom/content/KnownSpellIdentifier.java
```diff
@@ -146,28 +147,24 @@ public Integer getSpellLevel()
...
-					{
-						spellList.add(sp);
-					}
-				}
-			}
-		}
+			spellList.addAll(hml.getKeySet().stream()
+			                    .filter(classSpellLists::contains)
+			                    .filter(cdomList -> (spellLevel == null) || hml
+					                                                                .getListFor(cdomList)
```
thpr: This seems really complicated for this construct.

grimreaper: IMHO this isn't that complicated; its just two filters...

grimreaper:
per external convo

PR #1758, opened 2016-10-02
Replace 'sync libs to root' step with a symlink.

grimreaper:
This avoids copying files around, ensures build artifacts are always in
the same place, and moves us in the direction of avoiding the need for
anything to modify the root.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -243,16 +243,6 @@ task copyToRoot(type: Copy, dependsOn: [copyToLibs, jar, converterJar]) {
     rename "(.+)-$version(.+)", '$1$2'
 }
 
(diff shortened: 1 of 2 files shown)


Zaister:
I'm not sure it's a good idea having the execution environment rely on anything insider a volatile space like the build directory.

grimreaper:
rebased

I'm not sure why we copy stuff around in any case.

cpmeister:
I need to confirm, but I don't think the symlink will behave well on a windows system.

grimreaper:
can someone with windows please check? I suspect this isn't an issue and this would clean up the build scripts a bit.

As for @Zaister's concerns: we shouldn't be copying the final artifact to the root of the repo. If anything we should copy them to a dedicated "output" directory.

PR #1762, opened 2016-10-02
migrate running of checkstyle, findbugs, and pmd into their own target

grimreaper:


grimreaper:
The report can be generated by running the tool. At the moment I silence the output since we have way too many errors for it to be meaningful.
The related deps are defined as plugins in build.gradle.

PR #1773, opened 2016-10-05
reformat files to eliminate checkstyle warnings

grimreaper:


code/src/java/pcgen/core/EquipmentChoice.java
```diff
@@ -376,9 +381,7 @@ public void addSelectableEquipment(
 	{
 		for (Equipment aEquip : Globals.getContext().getReferenceContext().getConstructedCDOMObjects(Equipment.class))
 		{
-			if (
-				aEquip.isType(typeString) &&
-				!this.getAvailableList().contains(aEquip.getName()))
+			if (aEquip.isType(typeString) && !availableList.contains(aEquip.getName()))
 			{
 				this.getAvailableList().add(aEquip.getName());
```
thpr: If you're removing this.getAvailableList() this is another instance

code/src/java/pcgen/core/EquipmentChoice.java
```diff
@@ -623,8 +621,7 @@ else if ("STAT".equals(kind))
 				{
 					this.addStats();
 				}
-				else if ("SKILL".equals(kind) || originalkind.equals("SKILL")
-						&& "ANY".equals("SKILL"))
+				else if (originalkind.equals("SKILL") && "ANY".equals("SKILL"))
```
thpr: This is a logic change - what if it's not anything predicted by the IFs... we can't assume it's SKILL can we?

code/src/java/pcgen/io/exporttoken/AbilityToken.java
```diff
@@ -254,50 +253,44 @@ protected String getTokenForCategory(String tokenSource,
...
-		{
-			aList = AbilityToken.buildAbilityList(types, negate, abilityType,
-					view, aspect, abilityList);
-		}
-		else
-		{
-			aList = AbilityToken.buildAbilityList(key, view, abilityList);
-		}
+		MapToList<Ability, CNAbility> abilityCNAbilityMapToList =
+			(key == null)
```
thpr: I think this is too complicated for an inline IF.  They should be short, not wrapping lines.

code/src/java/pcgen/io/exporttoken/AbilityToken.java
```diff
@@ -420,7 +411,8 @@ protected String getTokenForCategory(String tokenSource,
 	 * @return True if it matches one of the types else false
 	 */
 	static boolean abilityMatchesType(String abilityType,
-		Ability aAbility, List<String> types, List<String> negate)
+	                                  Ability aAbility, Collection<String> types,
```
thpr: Similar to another PR, we need to get consistent on these wrapping rules.  This doesn't match how the standards file wraps

PR #1779, opened 2016-10-15
[build] Remove brackets after trouble with libs not copied

vincent-lhote:


LegacyKing:
@cpmeister or @thpr can you verify this fixes the build issue and merge?

grimreaper:
LGTM

cpmeister:
I can't fathom how this fixes anything. I think this change might be the result of a bug witch hunt. But given that is it harmless and provides a bit more consistency I'll allow it.

PR #1783, opened 2016-10-16
[DATA][PFRPG][UM] Fix typo in Ear-Piercing Scream spell description text

romen:
There was a minor typo in the description text of Ear-Piercing Scream:

```
[...] takes (CASTERLEVEL/2,5)d6 points [...]
```

instead of

```
[...] takes (min(CASTERLEVEL/2,5))d6 points [...]
```


LegacyKing:
Though correct, the default for the casterlevel calculations assume a min. Unless the values indicate otherwise, in which case we need to correct thousands of such formulas in the spells.

romen:
The output sheet was showing 2147483647d6 (or something close to MAX_INT), that's why I checked the data file and figured out this was a typo. Adding the additional parenthesis and min solves the issue and shows the correct value.
Is there another way to fix this?

LegacyKing:
If you're getting a multiple digit number like that, then I fear there is something much worse going on in the code base. I'd rather address the deeper issue than band-aid over the issue. Can you provide a copy of the super-powered character capable of obliterating any enemy with a single spell? "*.pcg"

romen:
Here is the all-obliterating school savant arcanist, whose shriek every dungeon delving creature fears!! The [Menas Merrion.zip archive](https://github.com/PCGen/pcgen/files/532182/Menas.Merrion.zip) contains the .pcg file, alongside rendered PDF and HTM versions of the Standard ftl OS where you can see the omni-annihilating Ear-Piercing Scream spell description.

Hope it helps! Let me know how I can further help!

LegacyKing:
I'll investigate this later, time to get back to school work.

LegacyKing:
Merging, since time is not available.

PR #1795, opened 2016-10-24
Fix bug: DATA-3082 Raging Vitality not working for Bloodrage

LegacyKing:


honorablediscord:
was there a fix to this issue? i seem to be having it as well.

PR #1812, opened 2016-11-07
globally prefer isEmpty to .length == 0

grimreaper:


thpr:
< 100 files per PR please

PR #1813, opened 2016-11-07
instanceof always returns false for null

grimreaper:


grimreaper:
I don't think there are likely to be real performance impact but I'll try running some benchmarks later and check.

PR #1852, opened 2016-11-12
partial of #1812 (analysis)

grimreaper:


code/src/java/pcgen/core/analysis/BonusCalc.java
```diff
@@ -58,7 +58,7 @@ public static double bonusTo(
 		final Collection<BonusObj> aBonusList,
 		final PlayerCharacter aPC)
 	{
-		if ((aBonusList == null) || (aBonusList.size() == 0))
+		if ((aBonusList == null) || (aBonusList.isEmpty()))
```
thpr: Unnecessary parens around isEmpty

PR #1854, opened 2016-11-13
clean up Gui2InfoFactory

grimreaper:


cpmeister:
Why are you mix & matching uses of StringUtils.join() and StringJoiner? There should at least be some consistency on which one is used. That said there is also the Collectors.joining() method as well. I think the less reliance on 3rd party libraries the better.

grimreaper:
I mix and match since they are slightly different use cases though I managed to rewrite. I prefer using a stream + collector so that's what I'm doing.

PR #1883, opened 2016-11-21
clean up a fair bit of the startup code

grimreaper:


code/src/java/pcgen/gui2/PCGenFrame.java
```diff
@@ -2057,11 +1990,9 @@ public CustomEquipResult showCustomEquipDialog(CharacterFacade character,
 				new EquipCustomizerDialog(this, character, equipBuilder);
 		Utility.setComponentRelativeLocation(this, eqDialog);
 		eqDialog.setVisible(true);
-		CustomEquipResult result =
-				eqDialog.isCancelled() ? CustomEquipResult.CANCELLED : eqDialog
-					.isPurchase() ? CustomEquipResult.PURCHASE
-					: CustomEquipResult.OK;
-		return result;
+        return eqDialog.isCancelled() ? CustomEquipResult.CANCELLED : eqDialog
```
thpr: IMHO, this should be an IF not a double-inline - too hard to read the way it is

code/src/java/pcgen/gui2/PCGenFrame.java
```diff
@@ -1795,11 +1742,15 @@ public void run()
 			statusBar.setSourceLoadErrors(worker.get());
 			//now that the SourceFileLoader has finished
 			//handle licenses and whatnot
-			StringBuilder sec15 = new StringBuilder(" ");
-			sec15.append(readTextFromFile(ConfigurationSettings.getSystemsDir() + File.separator
-					+ "opengaminglicense.10a.txt"));
-			sec15.append(loader.getOGL());
-			section15 = sec15.toString();
+            String
+                    sec15 =
```
thpr: Since it is no longer a StringBuilder, why even have the local variable sec15?

code/src/java/pcgen/gui2/tools/Utility.java
```diff
@@ -323,10 +306,6 @@ public static void selectDefaultBrowser(Component parent)
 
 		if (PCGenSettings.getBrowserPath() == null)
 		{
-			//No action, as we have no idea what a good default would be...
```
thpr: This logic is incorrect, since the IF test was not reversed when only the else was used

LegacyKing:
We also have failing tests.

PR #1886, opened 2016-11-22
1884 Added correct weapon proficiencies for the Picaroon archetype.

reesmanp:
No idea if the formatting is correct, but the code works as expected.

Issue 1884


LegacyKing:
Works for me.

PR #1903, opened 2016-11-28
[CODE-3102] Multiple equipmentset export (Freemarker) fix.

eslafgh:
Changing the CalcEquipSetId does not change data to be output.
Need to set the CalcEquipmentList after the id change to get the
equipment data to change.


LegacyKing:
@cpmeister or @thpr can you review this. Thanks!

PR #1907, opened 2016-11-29
[DATA-3080] gtpsop and guild guide updates

mjmeans:


mjmeans:
It's functional and tested with 6 older PFS characters of different classes. There are some known issues with the implementation noted in the readme.md file. I'll work on resolving those in separate PR(s).

While this doesn't break exiting PFS PC's it does leave an ability category with "too many selections" for Faction that the user will have to resolve by removing, and then add the correct active faction and retired faction abilities so all faction traits and abilities remain functional. This is the cleanest solution, as Andrew and I discussed on IM.

PR #1908, opened 2016-11-30
Added subrace Forest Gnome

FoxShot:


LegacyKing:
1. Failing tests
2. Forest Gnome is not in the SRD, and cannot be included. Your only choice is Rock.
3. If the SRD does not contain it, we cannot include it. The Player's Handbook and the SRD are not the same. Please review what is included in the SRD - https://media.wizards.com/2016/downloads/DND/SRD-OGL_V5.1.pdf

Cheers,

PR #1909, opened 2016-11-30
Added rest of Warlock lvl 1 spells

FoxShot:


LegacyKing:
I see no reference to those spells in the SRD -- https://media.wizards.com/2016/downloads/DND/SRD-OGL_V5.1.pdf

As such, we cannot include it.

LegacyKing:
Page 111 - 
1st Level
Charm	Person
Comprehend	Languages
Expeditious	Retreat
Hellish	Rebuke
Illusory	Script
Protection	from	Evil	and	Good
Unseen	Servant

PR #1818, opened 2016-11-07
simlify arraycopy

grimreaper:

PR #1911, opened 2016-12-02
Fix regression from PR #1818 that breaks the display of the skill tre…

Zaister:
…e view for the TYPE/NAME view


Shredder121:
There was also a second occurrence still not fixed.
But it might be outside of the scope of this PR.

Zaister:
Yes, indeed. The second occurrence will also be addressed but is not relevant to this PR. In fact it seems the class SpinningTabbedPane that contains that occurrence isn't used at all. Nevertheless it should be fixed.

PR #1912, opened 2016-12-02
Fix regression in SpinningTabbedPane.getMovableTabIndices() introduced

Zaister:
in PR #1818


cpmeister:
Is this expected? I mean, if there is only a single tab this will try to create an array of size 0. Is that the intention?

Zaister:
I'm not sure but that's what it seems the class is supposed to do (check out the comments). However as of now, this class is not actually used at all anywhere.

PR #1914, opened 2016-12-04
[BUG-3112] Unnecessary javadoc and sources jars in libs and classpath.

eslafgh:
Use ext: 'jar' for the PCGen-Base and PCGen-Formula dependencies so the
corresponding javadoc and sources jars are not included as dependencies.
Not sure why this works but it works.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -158,9 +158,11 @@ dependencies {
     compile group: 'xml-apis', name: 'xml-apis', version:'1.4.01'
     compile group: 'com.incors.plaf', name: 'kunststoff', version:'2.0.2'
     compile group: 'net.sourceforge.argparse4j', name: 'argparse4j', version: '0.7.0'
-    compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.83'
-    compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.95'
     compile group: 'org.slf4j', name: 'slf4j-api', version: '1.7.21'
+
+    // Use ext: 'jar' to stop javadoc and sources jars being included
+    compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', ext: 'jar', version:'1.0.83'
+    compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', ext: 'jar', version:'1.0.95'
 
     runtime group: 'ch.qos.logback', name: 'logback-classic', version:'1.1.6'
 


LegacyKing:
@thpr this is up your alley. Can you review?

thpr:
Given the description, I don't think this has a result I would appreciate.

I intentionally include the sources, since these are NOT external - they are internal, and when I am diagnosing and want to control-click on a method name I want it to take me into the actual code, which requires the SOURCES JAR file. 

I don't want to link projects in eclipse because that starts to presume other's project names and also forces others that wouldn't otherwise want to have those projects to download them for the main to correctly resolve.

So I'm not sure this change is appropriate.  Having said that, if there is a way to have this download them for development purposes, but not for the actual distribution, I'm all for it - those don't need to go into the external distribution.

eslafgh:
You may have a point - I have 2 projects setup using Idea, in one the sources in the gradle cache are automatically associated with the PcGen-* libraries even if I apply my change and up the versions. In the other however they are not, the only reason I can think of is that in the second I rewrote build.gradle to produces a cleaner build (IMO) and it has always had the ext: 'jar' change I proposed in the pull request. Both projects were "import project" from the build.gradle file.

After some testing/digging around I think I have a better solution to the problem. Change the way you publish the PCGen-{base,Formula} libraries. Put the sources and javadoc artifacts into a different configuration.

In build.gradle for pcgen-base and pcgen-formula change the publications section to

```
    publications {
        ivy(IvyPublication) {
            from components.java
            configurations {
                sources {}
                javadoc {}
            }
            artifact(sourceJar) {
                type "sources"
                conf "sources"
            }
            artifact(javadocJar) {
                type "javadoc"
                conf "javadoc"
            }
            descriptor.withXml {
                asNode().info[0].appendNode('description', description)
            }
        }
    }
```
With the sources and javadocs no longer in runtime conf they are no longer placed in the libs dir or classpath. I chose the "javadoc" and "sources" configurations as that seems to be the convention if you check out the other ivy.xml files in your gradle cache.

With this setup Eclipse and Idea automagically download the sources when importing/refreshing from build.gradle and associate them with the external library just like they do for many other libraries e.g. commons-io.

The javadocs are not downloaded but if you want them you can do
```
apply {
    ...
    plugin 'idea'
    plugin 'eclipse'
}

idea { module { downloadJavadoc = true } }
eclipse { classpath { downloadJavadoc = true } }
```
On build.gradle refresh/import this  will cause the download of the javadocs into the gradle cache for the PcGen libraries and many of the other external libraries. They do not get auto associated with the library for me but I did try and work out why.

PR #1919, opened 2016-12-18
[PFS] Fix Chronicles and Boons formatting

mjmeans:
Moves PFS Boons our of Special Qualities to their own table. Fix formatting of PFS Boons and Chronicles tables.


preview/d20/5e/Standard.htm.ftl
```diff
@@ -1753,6 +1753,37 @@ ${pcstring('ABILITYALL.Special Ability.VISIBLE.${specialQuality}.TYPE=SpecialQua
  </tr>
 <!-- STOP Feat Table -->
 
+<!-- START PFS Boon Table -->
```
LegacyKing: Unless 5th edition D&D has PFS Boons (which I'm not aware of) can we remove this change.

mjmeans:
You are correct that 5e doesn't use PFS Boons per-se. But neither does 5e have PFS Chronicles, yet there is a pre-existing section in that file for chronicles (that I didn't put there). But there is also a "PFS Chronicles" section in both outputsheets/d20/5e as well. Maybe those should be removed as well?

mjmeans:
Should they also be removed from sagaborn and 4e?

mjmeans:
I also see that there is a reference to pfs in d20\4e\pdf\common_sheet\block_pc_header_row.xslt. It is referring to the pfs character ID. Maybe that should be removed as well?

LegacyKing:
Eventually, all non-relevant sections should be removed from those.
However, excising blocks is a tricky business and can render a sheet
dead to export, which is why I have not taken the time to remove those.

I don't want to add more work to that project, when/if I ever decide I
have time to tackle it.

mjmeans:
Well, I included the removal of the PFS sections from 5e. Is there more for me to do?

PR #1920, opened 2016-12-18
added blessing of fervor to oracle list

ProtoJazz:


ProtoJazz:
I would assume it's covered on page 42 of the apg
Spells: An oracle casts divine spells drawn from the
cleric spell lists (see pages 226–229 of the Pathfinder RPG
Core Rulebook).

PR #1934, opened 2016-12-31
Fix CODE-3115: NPE when trying to display info about an ability

Zaister:


Zaister:
The problem here is that the NPE occurs when a FACT has no VISIBILITY set. Recently some new FACTs were added without a VISIBILITY tag, and this triggered the problem.

Now, according to the pcgen.cdom.content.ContentDefinition class from which the FactDefinition and FactSetDefinitions classes used here are derived, the default for the visibility is HIDDEN. However, this isn't implemented in a way that helps here. The visibility property of the class is not initialized, and the comments state:
>The value if this is "null" should be interpreted as "HIDDEN", but is necessary to have as a separate 
>value for establishing whether it was set by data.
And on the getVisibility() method:
>Note: This may be null, if the default visibility for ContentDefinition (HIDDEN) is used and hasn't been >explicitly set by the setVisibility method.
I don't know why this was implemented in this way, and where it might be needed that a visibility could be null, but I'd rather fix this at the source than fix a symptom, as this PR does. If anyone knows more about this implementation, please elaborate.

In the meantime, this fixes the acute problem.

Zaister:
@cpmeister It would be great if this could be merged, as the bug it fixes really impacts the ability tab currently.

grimreaper:
LGTM

LegacyKing:
Likewise, this looks safe to merge. @thpr can you review this please.

PR #1946, opened 2017-01-03
NEWTAG-596 add dropIntoContext function

thpr:
Update to use Dynamic Solver Manager
Formula Library Version 101


LegacyKing:
@cpmeister needs review please.

PR #1952, opened 2017-01-14
do more work to eliminate PCGenTestCase

grimreaper:
- this reduces the complexity of PCGenTestCase (the reason it exists)
- it also removes a dep where it isn't needed.

Ideally PCGenTestCase can be turned into a trait eventually


grimreaper:
bah, I ran 'test' locally instead of 'fulltest'

PR #1953, opened 2017-01-14
Bump a variety of dep versions

grimreaper:


LegacyKing:
This is a remerge - can we eliminate the other commits already in the system please. Thanks!

grimreaper:
done!

grimreaper:
FWIW "incubating" releases are still final releases. Its just a required suffix during their apache integration.

PR #1956, opened 2017-01-14
simplify Combatant.java

grimreaper:


code/src/java/gmgen/plugin/Combatant.java
```diff
@@ -302,13 +298,13 @@ public void endRound()
 	 */
 	public void heal(int heal)
 	{
-		setStatus(hitPoints.heal(heal));
+		this.status = hitPoints.heal(heal);
```
thpr: Not sure I agree with this style of change.  The purpose of having a set* method is so that any necessary side effects can be done in that method.  That may not be the case here, at this time, but what is the basis of reasoning for not using the set method?

PR #1962, opened 2017-01-14
Convert from shift-left to doLast as the former is deprecated

grimreaper:


LegacyKing:
Conflicted, please resolve.

PR #1965, opened 2017-01-15
clean up some svn references

grimreaper:


LegacyKing:
I would have merged this, but it's touching on a java file that I don't have much understanding of to mess with.

PR #1968, opened 2017-01-15
After bumping gradle to version 3.3 also bump wrapper (and re-gen)

grimreaper:


LegacyKing:
Why are we losing the .bat file?

PR #1975, opened 2017-01-15
use map functions

grimreaper:


code/src/java/pcgen/cdom/enumeration/AspectName.java
```diff
@@ -105,13 +105,8 @@ public int getOrdinal()
...
 	{
 		initializeNameMap();
-		AspectName aspect = nameMap.get(name);
-		if (aspect == null)
-		{
-			aspect = new AspectName(name);
-			nameMap.put(name, aspect);
-		}
-		return aspect;
+        AspectName aspect = nameMap.computeIfAbsent(name, k -> new AspectName(name));
```
thpr: Spaces/Tabs

grimreaper:
closing and will recreate with updated settings. Formatting complaints are the major features of my PRs which is good and bad. It means the logic is staying the same, but its also the hardest to fix since its mostly IDE settings.

Thanks!

PR #1983, opened 2017-01-16
we never actually invoke private ctors (and doing so is an anti-pattern)

grimreaper:


thpr:
This exists to "fool" code coverage tools into going to 100%

PR #1984, opened 2017-01-16

grimreaper:


code/src/test/plugin/PluginBuildTest.java
```diff
@@ -69,6 +69,7 @@ public static Test suite()
 	/**
 	 * Check for the presence of all 'bonus' token plugins.
 	 */
+	@org.junit.Test
```
thpr: Why the fully defined org.junit.Test in this file and not others?

grimreaper: no good reason, will fix

PR #1987, opened 2017-01-17
convert all simple junit3 test cases to junit4

grimreaper:


code/src/test/pcgen/core/CampaignTest.java
```diff
@@ -23,80 +23,77 @@
...
 
+	@Test
 	public void testURL() throws Exception
 	{
 		final String eCommerceName = "Barcommerce";
 		final String eCommerceURL = "http://www.barcommercesite.com/product_info.php?products_id=12345&affiliate_id=54321";
 		final String eCommerceDesc = "Support PCGen by buying this source now!";
 
-		assertEquals("No URLs in the campaign to start", true, testCamp
+		Assert.assertEquals("No URLs in the campaign to start", true, testCamp
```
thpr: Should we static insert these methods for some sanity/less file churn?

grimreaper: likely. This also doesn't build. I'll rework this PR.

PR #1988, opened 2017-01-17
simplify Tips.java

grimreaper:


code/src/java/translation/util/Tips.java
```diff
@@ -70,37 +75,49 @@ public static void generatePOT(File rootDirectory, String potFilename)
...
-					File[] tipsFiles = subfiles[i].listFiles(filter);
-					for (int j = 0; j < tipsFiles.length; j++)
+					File[] tipsFiles = subfile.listFiles(filter);
+					for (File tipsFile : tipsFiles)
 					{
-						log("Found {0}", tipsFiles[j]);
-						// for each non comment line of the file, put its content in a Set<String>
+						log("Found {0}", tipsFile);
+						// for each non comment line of the file, put its content in a
+						// Set<String>
```
thpr: If you are going to wrap single line comments in your formatter, then perhaps they should be  converged to multi-line comments if they are that long

code/src/java/translation/util/Tips.java
```diff
@@ -340,16 +361,28 @@ public static void generateTips(File rootDirectory, File translation, String tra
 									String translatedLine = tipsTranslated.get(readLine);
 									if (translatedLine == null)
 									{
-										log("null translated line in {1}, original {0}", readLine, translation);
+										log(
```
thpr: While you're in this file, can you see if there is a sane way to break up this method?  This is like 8 indents in, which is a bit silly

PR #1997, opened 2017-01-18
convert the config directory to a proper command line argument

grimreaper:


grimreaper:
I think when I resolved conflicts I missed an import. Closing and will recreate locally

PR #2003, opened 2017-01-19
simplify 'todo' related code

grimreaper:


code/src/java/pcgen/gui2/facade/TodoFacadeImpl.java
```diff
@@ -33,6 +29,7 @@
  * 
  * @author James Dempsey &lt;jdempsey@users.sourceforge.net&gt;
  */
+@SuppressWarnings("TodoComment")
```
thpr: Why would we suppress this?

grimreaper: IntelliJ in particular gets annoyed by the text string "todo". This is useless when you're writing a TODO class.

thpr: oy, ok

PR #2007, opened 2017-01-20
remove two old files

grimreaper:
as far as I could tell neither of these are used anymore


thpr:
I think these are used by the installer deployment?

grimreaper:
maybe? I'm not sure I have a way to test that. I did grep the codebase for the filename and didn't notice anything obvious. They also reference sourceforge and other ancient things.

@LegacyKing do you have any idea?

LegacyKing:
I believe Maven was replaced by gradle. We haven't updated those files. I don't think we'll break anything by removing them.

PR #2008, opened 2017-01-20
remove build-gradle.xml

grimreaper:
as far as I could tell this is an ancient ant compat file


grimreaper:
meh, I misunderstood what was happening. This needs to be fixed, but differently

PR #2009, opened 2017-01-20
remove $Id$ across the codebase

grimreaper:


grimreaper:
This change was automatically generated.  (although I did one file by hand)
`sed -i '' '/\$Id\$/d' $(ag -lQ '$Id$')`

Although its many files, you should review the script which generated the change, not the change itself.

PR #2012, opened 2017-01-20
dropkick everything

grimreaper:
This is a combination of multiple other PRs to prove they don't conflict.


LegacyKing:
A lot of this might be best to run by the dev group. Authors, Created on and the like are all standards that have been in place since the project inception.

thpr:
@grimreaper  Unless you really need this PR, I don't need to see them

grimreaper:
close PR

PR #2013, opened 2017-01-21
Remove Laf entirely

grimreaper:
LaF is going to be gone in GUIv3 so it might as well be removed early.
This reduces the surface area required to update the GUI


LegacyKing:
I'd be hesitant to remove something until the replacement is in place. Otherwise, we've just removed a feature that we are using. Normally we build the replacement first, then integrate, test and then after it looks good, we remove the old structure. This appears to be way too premature to me. 6.07.02 losing the LaF when the GUI3 isn't up and running would be bad.

grimreaper:
There isn't really a replacement feature, but understood this might be too early. Closing PR

PR #2014, opened 2017-01-21
remove dead code CharacterSheetFrame

grimreaper:


LegacyKing:
This is in the GUI2? That's our current gui. @cpmeister is this not used?

PR #2015, opened 2017-01-21
clean up some GUI code

grimreaper:


code/src/java/pcgen/gui2/csheet/CharacterSheetPanel.java
```diff
@@ -103,17 +103,13 @@ public CharacterSheetPanel()
...
-			@Override
-			public Thread newThread(Runnable r)
-			{
-				Thread thread = new Thread(r, "Character-Sheet-Refresher-Thread");
-				thread.setDaemon(true);
-				thread.setPriority(Thread.NORM_PRIORITY);
-				return thread;
-			}
-		});
+            Thread thread = new Thread(r, "Character-Sheet-Refresher-Thread");
```
thpr: Another file with tabs vs spaces.  Getting very hesitant to do lots of your merges because it's creating inconsistency that other automated formatting will then try to redo and we get into thrashing lines that didn't really change.  Please convert to tabs.

grimreaper: I suspect this is just a setting in my IDE which needs to get turned on again. Will find it and fix.

grimreaper: found the setting: it was set to use spaces instead of the tab character. Future PRs should be fixed. Please continue to point out other style changes that I should fix. Ideally all of this could be stored in the repo so future contributors have an easier time :)

PR #2017, opened 2017-01-21
remove @author tags

grimreaper:
This information is stored in VCS so don't bother keeping it in code.


LegacyKing:
This is a rather large departure, I would have asked on the _dev list, before undertaking this. Since not all of this is properly in git. We started in cvs, then svn and now we're in git.

thpr:
Conflicts.  Smaller # of files are also good, since I do want to review even automated changes, since in the past automated changes have "eaten too much" at times.

PR #2018, opened 2017-01-21
simplify a ton of code

grimreaper:


code/src/java/pcgen/persistence/lst/StatsAndChecksLoader.java
```diff
@@ -101,7 +101,7 @@ else if ("ALIGNMENTNAME".equals(key))
 			return null;
 		}
 		String name = token.substring(colonLoc + 1);
-		if ((name == null) || (name.isEmpty()))
+		if (name.isEmpty())
```
LegacyKing: I've noticed we use certain conventions because of bad actors in the past, will this cause any issues?

code/src/java/pcgen/persistence/lst/URIEntry.java
```diff
@@ -196,10 +189,10 @@ public URIEntry getRelatedTarget(String fileName)
 	 *            The offset from the root URI for the URIEntry
 	 * @return A new URIEntry from the given Campaign name, rootURI and offset
 	 */
-	public static URIEntry getURIEntry(String campaignName, URI rootURI,
-		String offset)
+	static URIEntry getURIEntry(String campaignName, URI rootURI,
+								String offset)
 	{
-		if (offset == null || offset.isEmpty())
+		if ((offset == null) || offset.isEmpty())
```
LegacyKing: Why does this use both == null and .isEmpty instead of the .isEmpty from earlier? Seems like an inconsistent standard here.

code/src/java/pcgen/pluginmgr/InteractivePlugin.java
```diff
@@ -35,8 +35,7 @@
  */
 public interface InteractivePlugin extends PCGenMessageHandler
 {
-	public final static int NORMAL_PRIORITY = 10;
-	
+
```
LegacyKing: Can you explain what this is actually removing?

code/src/java/pcgen/rules/persistence/CDOMControlLoader.java
```diff
@@ -126,7 +119,7 @@ private boolean parseSubLine(LoadContext context, String val, URI source)
 			return false;
 		}
 		String name = lineIdentifier.substring(colonLoc + 1);
-		if ((name == null) || name.isEmpty())
+		if (name.isEmpty())
```
LegacyKing: I'm getting confused by the inconsistent standards being applied. Why do we have a two part AND conditional in some cases, but in others we drop it to a single conditional?

code/src/java/pcgen/persistence/lst/StatsAndChecksLoader.java
grimreaper: in this case name can't possibly be null since token.substring never returns null. I don't think this will cause issues.

code/src/java/pcgen/persistence/lst/URIEntry.java
grimreaper: since I can't prove that offset won't be null.

code/src/java/pcgen/pluginmgr/InteractivePlugin.java
grimreaper: nothing. Nothing uses this.

code/src/java/pcgen/rules/persistence/CDOMControlLoader.java
grimreaper: we drop it to a single condition where we can prove that the null check is useless, especially from the line right before.

code/src/java/pcgen/rules/persistence/CDOMKitLoader.java
```diff
@@ -81,8 +82,10 @@ public boolean parseSubLine(LoadContext context, Kit obj, String val,
 		return true;
 	}
 
-	private <CC extends BaseKit> boolean subParse(LoadContext context, Kit kit,
-			CDOMSubLineLoader<CC> loader, String line)
+	private static <CC extends BaseKit> boolean subParse(LoadContext context,
+														 CDOMObject kit,
```
thpr: This indent seems "out of control"

code/src/java/pcgen/pluginmgr/PluginManager.java
```diff
@@ -80,17 +82,19 @@ public int compare(InteractivePlugin arg0, InteractivePlugin arg1)
 	public void startAllPlugins()
 	{
 		PCGenMessageHandler dispatcher = msgHandlerMgr.getPostbox();
-		for(InteractivePlugin plugin : pluginMap.keySet())
+		for(final Entry<InteractivePlugin, Boolean> interactivePluginBooleanEntry : pluginMap
```
thpr: Why such a long localized/temporary variable name?  "entry" would be sufficient, IMHO.  Also, not sure it needs to be "final" - another case where I don't think it adds much value and just adds another copy of a keyword for people to have to analyze

code/src/java/pcgen/rules/persistence/ChoiceSetLoadUtilities.java
```diff
@@ -71,25 +73,31 @@ private ChoiceSetLoadUtilities()
...
 				QualifierToken<T> qual = getQualifier(context, sc,
-						primitive);
+						primitive
+				);
 				if (qual == null)
 				{
 					PrimitiveCollection<T> pcf = getSimplePrimitive(context,
-							sc, primitive);
+							sc, primitive
+					);
```
thpr: Is the separate line for ); our code standard?  (Don't think it is?)

PR #2021, opened 2017-01-21
Remove readme.txt from code standards

grimreaper:
- the Sun page no longer exists
- the Jalopy doesn't exist
- I've been trying to update the IntelliJ files so they are at least
closest to master (though eclipse might be reasonable too)
- the IntelliJ settings are old


LegacyKing:
I'd agree. But I'll let @thpr comment, since he'd know.

thpr:
This is fine - our master for code standards should really be the wiki anyway

PR #2032, opened 2017-01-21
globally prefer junit.org instead of junit.framework; this will make it easier to convert other tests to Junit 4

grimreaper:


thpr:
Conflicts

thpr:
Also, please have <100 files

PR #2037, opened 2017-01-22
Preview Proficiency and Hit Dice

C-S-McFarland:
I don't see a simple and easy to reference Proficiency Bonus and Hit Dice shown on the preview. So I replaced the Misc Table and the Base Attack Bonus.


grimreaper:
why remove the existing stats instead of adding new ones?

C-S-McFarland:
My understanding is that Base Attack Bonus and these miscellaneous stats (Miss chance, arcane failure, armor check, max dex, spell resist) are no longer stats in 5e. So in a preview, they would not be helpful, and take up real-estate that could display other data.

grimreaper:
makes sense Review LGTM

PR #2040, opened 2017-01-24
simplify some more PlayerCharacter code

grimreaper:


code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -6874,12 +6869,10 @@ public void incrementClassLevel(final int numberOfLevels, final PCClass globalCl
 			for (int i = 0; i < numberOfLevels; ++i)
 			{
 				int currentLevel = getLevel(pcClassClone);
-				final PCLevelInfo playerCharacterLevelInfo = addLevelInfo(pcClassClone.getKeyName());
```
thpr: On what basis can this be removed?  Does this method not have side effects?

grimreaper: ah, it can't be. good catch! I think I misread the method earlier.

PR #2042, opened 2017-01-26
simplify NameGenPanel

grimreaper:


grimreaper:
looks like I missed something locally

PR #2044, opened 2017-01-28
eliminate some usages of Vector

grimreaper:


code/src/java/plugin/encounter/EncounterModel.java
```diff
@@ -62,15 +50,9 @@ public EncounterModel(String parentDir)
 	 */
 	public int getCR()
 	{
-		ReadXML xml;
-		File f = new File(dir + File.separator + "4_1.xml"); //$NON-NLS-1$
-		int i;
 		float cr = 0;
 
-		xml = new ReadXML(f);
-		xml.getTable();
```
thpr: Do we know this (and the constructor above it) have no side effects?  The older code is not always known for being properly object oriented

grimreaper: it has side effects (it reads an XML file) but as far as I could tell, it then drops it on the floor. getTable() should make use of it, but its certainly an empty operation.

This might have the side effect of logging if the file doesn't exist or is invalid but its really weird and not required for the remainder of the logic.

PR #2051, opened 2017-01-29
reduce amount of stuff happening inside of PlayerCharacter

grimreaper:


grimreaper:
this stuff doesn't build. guess I missed this one

PR #2059, opened 2017-01-30
Map is nice; make use of it

grimreaper:


thpr:
I won't hold up this one for it (but will wait for travis to complete)... but at some point can you inline a lot of the now unnecessary local variables?  (like the first half dozen changes are examples)

grimreaper:
hrm, looks like travis didn't like this one. Yes, I'll work on inlining useless local variables.

PR #2061, opened 2017-01-30
DATA-5159 Samsaran Shards of the Past Skill List

draconswan:
Fixing issue with Samsaran Shards of the past not displaying proper
skill list.


LegacyKing:
Just an FYI, that could have been simplified down to a CHOOSE TYPE=Base.

draconswan:
Good to know.  I was looking for an example, and I copied how Skill Focus worked.

PR #2064, opened 2017-01-31
remove some dead code and fix a bug

grimreaper:


code/src/java/plugin/dicebag/gui/DiceBagPluginView.java
```diff
@@ -153,7 +155,7 @@ public void internalFrameClosing(InternalFrameEvent e)
 					WindowConstants.DO_NOTHING_ON_CLOSE);
 			}
 			else if ((answer == JOptionPane.NO_OPTION)
-				&& (answer == JOptionPane.YES_OPTION))
```
grimreaper: this is the bug fixed - previously this was always false

code/src/java/plugin/dicebag/gui/DiceBagModel.java
```diff
@@ -233,10 +215,9 @@ public double rollDie(int index)
 	 * @param expression Die expression to parse
 	 * @return The double value of the expression.
 	 */
-	public double rollDie(String expression)
+	static double rollDie(String expression)
```
thpr: This is silly.  If this is going to be converted to static, it should just directly call the static in RollingMethods - in other words, this static method should be inlined.

code/src/java/plugin/dicebag/gui/DiceBagPluginController.java
```diff
@@ -241,8 +243,7 @@ protected void openInitialBags()
 				{
 					theModel.loadDiceBag(file);
 					noLoads = false;
-				}
-				catch (Exception e)
+				} catch (Exception e)
```
thpr: This is a formatting violation - braces go on their own lines

grimreaper:
Replaced with #2319

PR #2066, opened 2017-02-02
clean up more code

grimreaper:


code/src/java/plugin/dicebag/gui/DiceBagView.java
```diff
@@ -499,8 +512,8 @@ public void actionPerformed(ActionEvent e)
 
 			if ("ROLL".equals(command))
 			{
-				setRollResult(m_exprField.getText(), m_bag.rollDie(m_exprField
-					.getText()));
+				setRollResult(m_exprField.getText(),
```
thpr: Why give up the error checking in this change?

grimreaper: There isn't any error checking being done. I literally just inlined the function it gets resolved to.

PR #2068, opened 2017-02-02
Bump date to 2017

grimreaper:
"    // Ensure this always runs - we don't want an incorrect date     "


grimreaper:
this conflicts and git is acting weird when I try and resolve it. Closing and replacing with a new one (#2281).

PR #2072, opened 2017-02-03
Move some duplicated code into a utility class

grimreaper:


code/src/java/pcgen/core/term/TermUtil.java
```diff
@@ -0,0 +1,36 @@
+package pcgen.core.term;
```
thpr: Need LGPL license

grimreaper: done; also fixed "foo" as a variable name"

PR #2073, opened 2017-02-04
non util code cleanup

grimreaper:


code/src/java/pcgen/core/Description.java
```diff
@@ -359,11 +348,7 @@ else if (str.equals("%"))
 		}
 		if ( theVariables != null )
 		{
-			for ( final String var : theVariables )
-			{
-				buf.append(Constants.PIPE);
-				buf.append(var);
-			}
+			theVariables.stream().map(var -> Constants.PIPE + var).forEach(buf::append);
```
thpr: This does 2 StringBuffers because you have a PIPE+var.  So this will be a huge performance hit that I don't think is valuable just to do .stream()

code/src/java/pcgen/io/PCGVer2Creator.java
```diff
@@ -2552,7 +2511,7 @@ private void appendWeightLine(StringBuilder buffer)
 		buffer.append(IOConstants.LINE_SEP);
 	}
 
-	private String chosenFeats(PCTemplate pct)
+	private String chosenFeats(CDOMObject pct)
```
thpr: This is where I don't agree with a universal "be generic" rule.  It misleads users into thinking this is universal, when it's actually not.  This only works for Templates, and should be left specific as a signal to developers.

code/src/java/pcgen/io/PCGVer2Creator.java
```diff
@@ -2662,14 +2618,14 @@ private static String createComment(String s)
...
-	private String tempBonusName(final Object creator, Object target)
+	private static String tempBonusName(final Object creator, Object target)
 	{
 		final StringBuilder cb = new StringBuilder(100);
 
 		cb.append(IOConstants.TAG_TEMPBONUS).append(':');
 		if (creator instanceof CDOMObject)
 		{
-			final CDOMObject oCreator = (CDOMObject) creator;
+			final Identified oCreator = (CDOMObject) creator;
```
thpr: Wait, you're converting from A stuff = (A) thing  to B stuff = (A) thing?  Why not B stuff = (B) thing?
(This is a really confusing signal to a reader)

code/src/java/pcgen/io/PCGVer2Parser.java
```diff
@@ -474,18 +469,6 @@ private void cacheLine(String s)
 		cache.put(s.substring(0, s.indexOf(':')), s);
 	}
 
-	private void checkSkillPools()
-	{
-		int skillPoints = 0;
-
-		for (final PCClass pcClass : thePC.getClassSet())
-		{
-			skillPoints += pcClass.getSkillPool(thePC);
```
thpr: As much as I hate to have to ask this - have you verified this method does not have side effects?  Remember a lot of older code was not written by folks with an object-oriented history, so it doesn't always follow OO rules.

code/src/java/pcgen/io/PCGVer2Parser.java
```diff
@@ -2880,22 +2771,25 @@ private void parseFeatsHandleAppliedToAndSaveTags(
...
-								UserSelection.getInstance(),
-								UserSelection.getInstance());
-						}
-						else
-						{
-							thePC.addAbility(cnas, UserSelection.getInstance(),
-								UserSelection.getInstance());
-						}
-					}
+					Arrays.stream(assoc)
```
thpr: Heck no, this is unreadable.  Please leave the IF

code/src/java/pcgen/io/PCGVer2Parser.java
```diff
@@ -4260,21 +4154,17 @@ else if (IOConstants.TAG_NAME.equals(childTag))
...
-								.silentlyGetConstructedCDOMObject(
-									Ability.class, AbilityCategory.FEAT,
-									featKey);
-
-					if (anAbility != null)
-					{
-						metaFeats.add(anAbility);
-					}
-				}
+				element.getChildren()
```
thpr: Again, unreadable to anyone who doesn't write SCALA for a living - please leave as a FOR

code/src/java/pcgen/io/PCGVer2Parser.java
```diff
@@ -6170,25 +6056,23 @@ private void resolveLanguages()
 					{
 						Collection<Language> selected =
 								(Collection<Language>) ss.getSet(thePC);
-						for (Language l : selected)
+						selected.stream().filter(cachedLanguages::contains).forEach(l ->
```
thpr: I don't think we want full method behaviors inside parenthesis like this.  My 2c

PR #2078, opened 2017-02-05
DATA-3165: Numerous changes to enable Occult Adventures to load cleanly

ElroTheOnk:
Edits mostly to *_datacontrols to add assumed archetype flags that didn't exist, but some typo fixes in OA source itself also. Note that many of the OA archetypes will still not function as expected (because the various datacontrol toggles are not necessarily implemented), but at least it now loads without errors.


data/pathfinder/paizo/roleplaying_game/bestiary_3/b3_abilities_race.lst
```diff
@@ -1118,10 +1117,6 @@ Volcano Kami												CATEGORY:Special Ability	TYPE:RaceAbility.SpecialQuality
...
-Spell-Like Abilities			KEY:Shikigami ~ Spell-Like Abilities		CATEGORY:Special Ability	TYPE:ShikigamiRacialAbility.SpecialQuality.SpellLike																																																																																																																																																																																																																																						DESC:(CL 6th; concentration +8) At will-invisibility (self only), statue (self only) 3/day-hide from animals, purify food and drink 1/week-commune with nature (CL 12th)
-
```
LegacyKing: What does B3 have to do with OA?

data/pathfinder/paizo/roleplaying_game/bestiary_3/b3_abilities_race_oa.lst
```diff
@@ -0,0 +1,8 @@
...
+Spell-Like Abilities			KEY:Shikigami ~ Spell-Like Abilities		CATEGORY:Special Ability	TYPE:ShikigamiRacialAbility.SpecialQuality.SpellLike																																																																																																																																																																																																																																						DESC:(CL 6th; concentration +8) At will-invisibility (self only), statue (self only) 3/day-hide from animals, purify food and drink 1/week-commune with nature (CL 12th)
```
LegacyKing: This style of set up goes in the OA folder and is loaded if the B3 version is not loaded. We avoid messing with the originals and do the modifications at the source. OR, you can merely have OA load the creature and avoid this work around.

data/pathfinder/paizo/roleplaying_game/occult_adventures/_occult_adventures.pcc
```diff
@@ -56,18 +55,19 @@ TEMPLATE:oa_templates.lst
...
 SPELL:support/oa_spells_arg.lst|PRECAMPAIGN:1,INCLUDES=Advanced Race Guide
-SPELL:support/oa_spells_acg.lst|PRECAMPAIGN:1,INCLUDES=Advanced Class Guide
 SPELL:support/oa_spells_ma.lst|PRECAMPAIGN:1,INCLUDES=Mythic Adventures
 SPELL:support/oa_spells_mc.lst|PRECAMPAIGN:1,INCLUDES=Monster Codex
 
 # Needed for a Medium archetype
-ABILITY:@/pathfinder/paizo/roleplaying_game/bestiary_3/b3_abilities_race.lst|(INCLUDE:CATEGORY=Internal,Race Traits ~ Kami (Shikigami)|CATEGORY=Special Ability,Shikigami ~ Improvised Weapon Mastery|CATEGORY=Special Ability,Shikigami ~ Spell-Like Abilities)|!PRECAMPAIGN:1,Bestiary 3
-RACE:@/pathfinder/paizo/roleplaying_game/bestiary_3/b3_races.lst|(INCLUDE:Kami (Shikigami))|!PRECAMPAIGN:1,Bestiary 3
+ABILITY:@/pathfinder/paizo/roleplaying_game/bestiary_3/b3_abilities_race_oa.lst|!PRECAMPAIGN:1,Bestiary 3
+RACE:@/pathfinder/paizo/roleplaying_game/bestiary_3/b3_races_oa.lst|(INCLUDE:Kami (Shikigami))|!PRECAMPAIGN:1,Bestiary 3
```
LegacyKing: The OA extensions should be in a /support folder of the OA folder, and load these as set up. We don't alter the original books when possible.

ElroTheOnk: Agree they *should*, but as mentioned above - doesn't work.

data/pathfinder/paizo/roleplaying_game/bestiary_3/b3_abilities_race_oa.lst
ElroTheOnk: Unfortunately this approach doesn't work because of a conflict withing the abilitycategory file that can't be worked around (I think a limitation of the implementation of INCLUDE). So you end up either requiring B3 (which is ugly as hell for one familiar), or this (I accept) nasty hack. Or I'm missing something...

data/pathfinder/paizo/roleplaying_game/bestiary_3/b3_abilities_race.lst
ElroTheOnk: A B3 critter is added as a specific familiar type via an archetype ability

ElroTheOnk:
Hi Andrew - have commented against your comments (!), what next?

LegacyKing:
Howdy, the pcc INCLUDE obviously has issues, so instead, we make the changes to the local. Your implementation makes the changes in B3, instead of locally to the source that requires this. You're on the right track, just wrong location.

data/pathfinder/paizo/roleplaying_game/occult_adventures/_occult_adventures.pcc
LegacyKing: Does work, just not implemented here. Plenty of other examples of this already in the repo.  Look at the Advanced Class Guide for an example.

ElroTheOnk:
Hi Andrew - hope this is what you meant...

PR #2079, opened 2017-02-06
Corrected item Type tags

net-diver:
Updated magic item's type tags to be consistent with current system


LegacyKing:
I'll accept this, but I am trying to revise the slot system, and make it easier when we transition to the formula system.

PR #2087, opened 2017-02-07
clean up overland util code

grimreaper:


code/src/java/plugin/overland/util/Localized.java
```diff
@@ -146,20 +140,19 @@ public String toString()
...
-				String name;
-				if (attribute == null)
-					name = child.getTextTrim();
-				else
-					name = child.getAttributeValue(attribute);
-				if (lang != null && !lang.isEmpty())
-					addName(lang, name);
-			}
-		}
+		children.stream()
```
thpr: Another I think is too complex

PR #2088, opened 2017-02-07
clean up PCGTrackerPlugin a bit

grimreaper:


code/src/java/plugin/pcgtracker/PCGTrackerPlugin.java
```diff
@@ -287,14 +263,14 @@ public void handleOpen()
...
-				if (PCGFile.isPCGenCharacterOrPartyFile(selectedFile))
-				{
-					messageHandler.handleMessage(new RequestOpenPlayerCharacterMessage(this, selectedFile,
-						false));
-				}
-			}
+			Arrays.stream(chooser.getSelectedFiles())
+			      .filter(PCGFile::isPCGenCharacterOrPartyFile)
+			      .forEach(selectedFile -> messageHandler.handleMessage(new
+					      RequestOpenPlayerCharacterMessage(
```
thpr: This is another situation where I don't think this universally increases clarity... method inside a method call which is part of a stream is pretty deep.

PR #2093, opened 2017-02-10

grimreaper:


outputsheets/xsltsl-1.1/cmp.xsl
```diff
@@ -13,7 +13,6 @@
   <doc:reference xmlns="">
     <referenceinfo>
       <releaseinfo role="meta">
-        $Id: cmp.xsl,v 1.4 2004/10/10 06:18:57 balls Exp $
```
thpr: This is a problem - it's being used as content and we need a replacement somehow

grimreaper:
will resolve separately

PR #2130, opened 2017-02-10

grimreaper:


thpr:
conflicts

code/src/java/pcgen/system/CharacterManager.java
```diff
@@ -57,6 +57,17 @@
  * a listenable list that allows users of this class to not
  * only see what characters are open but to easily track any
  * changes to the list of available characters.
+<<<<<<< HEAD
```
thpr: GAACK

grimreaper: resolved

PR #2155, opened 2017-02-10

grimreaper:


code/src/java/pcgen/core/ClassType.java
```diff
@@ -19,6 +19,11 @@
 
 import pcgen.base.lang.UnreachableError;
 
+
+/**
+ * {@code Campaign}.
```
thpr: This seems out of place?

grimreaper: resolved

LegacyKing:
Conflicted

grimreaper:
resolved

PR #2171, opened 2017-02-11
remove boxing

grimreaper:


code/src/utest/pcgen/base/util/WeightedCollectionTest.java
```diff
@@ -146,7 +146,7 @@ public void testSimple() {
 		assertTrue(wc.add(I1));
 		assertFalse(wc.isEmpty());
 		assertTrue(wc.contains(I1));
-		assertTrue(wc.contains(new Integer(1))); // value semantic
```
thpr: I think you're missing the point here.  This test EXPLICITLY wants a new  object, hence the "value semantic" comment

code/src/utest/pcgen/base/util/WeightedCollectionTest.java
```diff
@@ -471,7 +471,7 @@ public void testGetWeight() {
 		assertTrue(wc.isEmpty());
 		assertTrue(wc.add(I1));
 		assertEquals(1, wc.getWeight(I1));
-		assertTrue(wc.add(new Integer(1))); // value semantic
```
thpr: Same as above

code/src/utest/pcgen/base/util/WeightedCollectionTest.java
grimreaper: I think I noticed this and changed the way it gets tested in a different PR and forgot about it in  this one; closing

PR #2179, opened 2017-02-11
remove trailing spaces from properties files

grimreaper:


system/Languages/LanguageBundle.properties
```diff
@@ -22,8 +22,8 @@ in_alignLabel = Alignment
 in_sourceLabel = Source
 in_orientation = Click to chang orientation of tables
 in_infoScrollTip = Any requirements you don't meet are in italics
-in_qualifyMess = You do not qualify for
```
thpr: As bad as the choice may be to have a space here - are you sure the code that uses this doesn't expect it?  How is this processed in practice?

grimreaper: The only of it is here:
code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
2152:					LanguageBundle.getFormattedString("in_qualifyMess", domain.getDisplayName()));
and there is no trailing space in the non-english or other property files.
code/src/java/pcgen/resources/lang/LanguageBundle.properties
142:in_qualifyMess=You do not qualify for {0}.

PR #2187, opened 2017-02-11
inline variables where possible

grimreaper:


code/src/utest/plugin/function/testsupport/AbstractFormulaTestCase.java
```diff
@@ -131,13 +134,13 @@ else if (valueOf instanceof Double)
...
 
-	protected void isNotValid(String formula, SimpleNode node,
-		FormatManager<?> formatManager, Class<?> assertedFormat)
+	protected void isNotValid(String formula, SimpleNode node)
 	{
 		SemanticsVisitor semanticsVisitor = new SemanticsVisitor();
 		FormulaSemantics semantics =
 				managerFactory.generateFormulaSemantics(localSetup.getFormulaManager(),
-					getGlobalScope(), assertedFormat);
+					getGlobalScope(), null
```
thpr: This is hobbling future function and making it harder to use going forward

PR #2208, opened 2017-02-11
TestCase from junit 3 -> 4

grimreaper:


code/src/utest/plugin/function/TestUtilities.java
```diff
@@ -21,40 +21,39 @@
...
 	{
-		if (delta < 0)
-		{
-			throw new IllegalArgumentException(
-				"Delta for doubleEqual cannot be < 0: " + delta);
-		}
 		double diff = d1 - d2;
-		return ((diff >= 0) && (diff < delta))
-			|| ((diff < 0) && (diff > -delta));
+		return ((diff >= 0) && (diff < TestUtilities.SMALL_ERROR))
```
thpr: Fundamentally I disagree with this merge - it is forcing a specific delta that the utility class claims it knows better than the caller.  That will not be universally true, and therefore this should not be changed.  I'll merge anyway because I really am less concerned about our test class structure, but for awareness...

PR #2221, opened 2017-02-11
remove last updated on

grimreaper:


LegacyKing:
Denied. Data uses these to track when we updated or included errata. These should be left alone as many volunteers only know this information based upon the files.

PR #2226, opened 2017-02-11
clarify if statements

grimreaper:


code/src/java/pcgen/io/exporttoken/WeaponToken.java
```diff
@@ -3121,12 +3117,9 @@ private static boolean appendSeperator(Equipment eq)
 	 */
 	private static boolean isNonStandard(Equipment eq)
 	{
-		if (eq.isType("Natural") || eq.isType("Both") || eq.isType("Melee")
-			|| eq.isType("Ranged"))
-		{
-			return false;
-		}
-		return true;
+		return !(
```
thpr: So a question here on clarity.  Is this easier to read with the outer parens than it would be if you did !a && !b && !c && !d ?

PR #2231, opened 2017-02-11
collapse identical if blocks

grimreaper:


code/src/java/pcgen/io/ExportHandler.java
```diff
@@ -2408,71 +2408,35 @@ private int dealWithFilteredTokens(String aString, PlayerCharacter aPC)
 		if ("CATCHPHRASE".equals(aString.substring(1)))
 		{
 			String catchPhrase = display.getCatchPhrase();
-			if (catchPhrase.equals(Constants.NONE))
```
thpr: This doesn't look collapsed - it looks gone - as do a few others?

PR #2252, opened 2017-02-11
make functions static where possible

grimreaper:


thpr:
Not a fan of these - this is something we have debated multiple times with multiple developers, and pretty much always settled the same place - this completely changes signaling to a developer and actually increases confusion.

PR #2273, opened 2017-02-11
make method static where possible

grimreaper:


LegacyKing:
@grimreaper - are you getting notifications on the broken builds? Several of your PRs are broken and need to be fixed. We don't want a review/merge delay because it's broken. Cheers,

grimreaper:
yes, I get notified. Several of them are only broken since they depend on other PRs. For now, please ignore them and I'll clean them up when I do my regular rebases.

PR #2282, opened 2017-02-12
Include both "test" and "slowtest" as two separate builds for PRs

grimreaper:


LegacyKing:
Uh, so... @grimreaper no tests have passed. In fact, none seem to have started?

PR #2284, opened 2017-02-13
maybe fix Jenkins

grimreaper:
I'm unsure if this will help or not, but maybe it needs to be explicitly enabled?


grimreaper:
from testing it seems we'll have to change a config value instead of modifying build.gradle

PR #2287, opened 2017-02-13
Make PMD fatal

grimreaper:
This adds excludes for all existing violations, but make it possible to
fatal on PMD errors. Over time we can remove more and more exclusions.

0 PMD rule violations were found.


grimreaper:
replaced with #2324

PR #2298, opened 2017-02-14
Enable the buildscan plugin

grimreaper:
This still requires manual work to use (-Dscan) but this makes it
possible.


grimreaper:
#2323 replaces this

PR #2321, opened 2017-02-18
remove clean from default build

grimreaper:
We already clean in CI so make developer builds reasonable.


grimreaper:
#2323 replaces this

PR #2323, opened 2017-02-18
General improvements to build.gradle

grimreaper:
- bump to gradle 3.4
- move common test tasks into withType
- move printing into common tasks
- stop printing on started tests (speeds up tests)
- enable buildscan and idea and tasks-tree as generally useful plugins
- bump pmd version
- bump fb-contrib version
- stop clean by default
- remove author lines


build.gradle
```diff
@@ -267,126 +262,82 @@ clean {
...
-    }
-
-    // Report each test as it is started.
-    beforeTest { descriptor ->
-    	logger.lifecycle("Running test: " + descriptor)
-    }
 }
 
 // Do the lot!
-task all(dependsOn: ['clean', 'build', 'slowtest', 'javadoc', 'allReports']) {
```
LegacyKing: Why are we removing clean across the board?

code/gradle/autobuild.gradle
```diff
@@ -32,7 +30,7 @@ task copySite(type: Copy) {
     outputs.upToDateWhen { false }
     from "${projectDir}/code/gradle/site"
     into "${destAutobuildDir}"
-    expand(copyright: '2016', version: version, timestamp: buildTimestamp)
+    expand(copyright: '2017', version: version, timestamp: buildTimestamp)
```
LegacyKing: Any reason we can't automate the current year?

grimreaper: Need to figure out how. Its likely possible

build.gradle
grimreaper: Gradle should work correctly without `clean` in the vast majority of cases. When it doesn't, its a bug. We already explicitly run `clean` in CI.

code/gradle/autobuild.gradle
grimreaper: done

grimreaper:
> Why are we including a binary ,jar?
I'm not actually sure. It looks like we already included it, and I'm just updating it.
 gradle/wrapper/gradle-wrapper.jar        | Bin 54208 -> 54212 bytes

I suspect that most other uses of gradle don't include it, but I haven't had the time to investigate this properly yet.

grimreaper:
going to split this up into separate PRs

grimreaper:
ref #2607

PR #2340, opened 2017-02-19
simplify

grimreaper:


thpr:
The level of indentation and parenthesis here makes me debate whether this is "simpler" in the new code

PR #2346, opened 2017-02-19

grimreaper:


grimreaper:
this was a bad old PR

PR #2353, opened 2017-02-20
Partial revert of some of the Tips changes

grimreaper:
This does break some style rules, but please treat this as a revert, not a code submission (i.e., I'll get to fixing the style issues eventually)


LegacyKing:
@thpr looks good to me. Build is broken, and this seems to be the fix.

PR #2395, opened 2017-02-25
Remove duplicate words; Use id, not name

grimreaper:


LegacyKing:
We really need to teach you how to make DATA PRs stick out, as this looks like another code PR

PR #2401, opened 2017-02-26
Replace custom code in MiscUtils with standard Java code

grimreaper:


code/src/java/gmgen/util/MiscUtilities.java
```diff
@@ -24,19 +22,13 @@
 {
 	private MiscUtilities()
 	{
-		//Do not allow instantiation of utility class
```
thpr: Not sure if this is being flagged by another tool as useless, but this comment is actually used to suppress either CheckStyle or PMD (I forget which) complaining about empty methods

grimreaper: Can't be pmd cause we fatal on errors; can't be checkstyle cause codacy would have flagged it. Maybe its old?

PR #2405, opened 2017-02-26
[utest] convert all of utest to junit4

grimreaper:
This can't easily be split up since various classes depend on each
other.

I verified that both before and after 17060 tests run and succeed.


grimreaper:
Note that you may want to look at #2406 and #2407 in advance of this one. This includes the others, but the others are parts that can be easily pulled out into separate commits (and thus this can be rebased and made easier to review).

grimreaper:
I found a way to split this up a bit; closing in place of more reviewable PRs.

PR #2408, opened 2017-02-26
convert all of selectionactor to junit4 (#2405)

grimreaper:


grimreaper:
this is just changing imports which isn't the important parts. Closing to make others more visibile.

PR #2409, opened 2017-02-26
Convert all of gmgen to junit4 (#2405)

grimreaper:


grimreaper:
looks like some of this got merged separately and is no longer needed.

PR #2419, opened 2017-03-02
DATA-3189: Legendary Item Powerful Ability Fix

draconswan:
Allowing any legenday item to select the first level of the Powerful
item ability.


data/pathfinder/paizo/roleplaying_game/mythic_adventures/ma_equipmods.lst
```diff
@@ -28,7 +28,7 @@ Intelligent			KEY:LEGEND_INTELLIGENT		TYPE:Legendary	VISIBLE:QUALIFY	ITYPE:Mythi
...
 Perfect Surge		KEY:LEGEND_PERFECT_SURGE	TYPE:Legendary	VISIBLE:QUALIFY					SOURCEPAGE:p.173	PRETYPE:1,MajorArtifact								SPROP:item's legendary surge can apply to any d20 roll. The surge adds a further +2 bonus when applied to one of the types of rolls initially chosen for it.
-Powerful			KEY:LEGEND_POWERFUL		TYPE:Legendary	VISIBLE:QUALIFY					SOURCEPAGE:p.173	PRETYPE:1,MinorArtifact,MajorArtifact					SPROP:item with this ability has two additional uses of legendary power per day. This ability can be taken up to three times.
+Powerful			KEY:LEGEND_POWERFUL		TYPE:Legendary	VISIBLE:QUALIFY					SOURCEPAGE:p.173															SPROP:item with this ability has two additional uses of legendary power per day. This ability can be taken up to three times.
```
LegacyKing: This is actually not the proper method to fix this as far as I can tell. TYPE should be changed to what that PRETYPE is, so those items are correctly caught.  PRETYPE should actually be something standard. ITYPE might be Legendary I suppose.

draconswan:
Go ahead and cancel/remove this pull.  I'll make all updates on the 3190 PR

data/pathfinder/paizo/roleplaying_game/mythic_adventures/ma_equipmods.lst
draconswan: I'm not seeing the issue with the way this is currently implemented, from what I read on TYPE/PRETYPE/ITYPE at least.
TYPE should a list of item types that have the ability to use this equipment mod, in this case, any item that has (and qualifies for) Legendary.
PRETYPE should be prereqs that enabled this mod, in this case Nothing,MinorArtifact,MajorArtifact.
ITYPE only applies to Intelligent because it's the only one that unlocks further sub-abilities.

PR #2420, opened 2017-03-02
DATA-3190

draconswan:
Fixing various bugs with Legendary Item Abilities


draconswan:
The Requirement was wrong for the one I remove it from.  The others are correct.  I'll update the data to be properly formed (I didn't know it was improper, just fixing the discrepancies between how it was coded and the actual rules)

draconswan:
I'm not seeing the issue with the way this is currently implemented, from what I read on TYPE/PRETYPE/ITYPE at least.
TYPE should a list of item types that have the ability to use this equipment mod, in this case, any item that has (and qualifies for) Legendary.
PRETYPE should be prereqs that enabled this mod, in this case Nothing,MinorArtifact,MajorArtifact.
ITYPE only applies to Intelligent because it's the only one that unlocks further sub-abilities.

PR #2425, opened 2017-03-05
[gradle] Remove layoutNsisBase

grimreaper:
It never worked in the first place and has caused issues lately.

### DIFF
--- a/code/gradle/release.gradle
+++ b/code/gradle/release.gradle
@@ -113,39 +113,6 @@ ext {
     }
 }
 
-task layoutNsisBase(type: Copy, dependsOn: copyToRoot) {
-    description="Assemble the required files for all PCGen installations"
-    doFirst {
-        // Release notes are required!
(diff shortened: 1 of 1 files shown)


grimreaper:
@LegacyKing You're the release guru. Is this stuff actually needed or is it for a feature that never worked?

LegacyKing:
I actually am not a Release Guru, I just happen to be the one doing the releases. The release guru is supposed to be on the Code team, since releases are part of the compiling of code and packaging it.  However, that all said, I know NSIS is part of the packaging, I know those files are generated as part of the release process, but I cannot say for any certainty, which ones are part of the process.

LegacyKing:
This seems to be used during the release process on Windows. I'd be very leery of altering this without doing test runs first.

grimreaper:
agreed. I wasn't sure and was mostly confused by recent failures in this code. Those failures may have come from elsewhere so dropping this PR.

PR #2468, opened 2017-03-08
DATA-3187 : Titan Strike

rcapehart:
Fixed data entry for Titan Strike to work properly.


LegacyKing:
This is a very messy commit. Looks like a file regression, as this should not have 167 lines altered back in time for a one line fix.

PR #2491, opened 2017-03-08
USe the same name as parent

grimreaper:


thpr:
I have to say I don't quite understand this one - there are mnemonics with using shorthand related to the object type that this is eliminating - what is the added value of this and is it worth it?

PR #2498, opened 2017-03-08
add missing braces

grimreaper:


code/src/java/plugin/overland/model/TravelMethodFactory.java
```diff
@@ -91,11 +91,11 @@
 			File[] dataFiles = path.listFiles(new XMLFilter());
 			SAXBuilder builder = new SAXBuilder();
 
-			for (int i = 0; i < dataFiles.length; i++)
+			for (File dataFile : dataFiles)
```
LegacyKing: This change not following the Braces fix PR theme.

code/src/java/plugin/overland/model/TravelMethodFactory.java
```diff
@@ -91,11 +91,11 @@
...
 			{
 				try
 				{
-					Document methodSet = builder.build(dataFiles[i]);
+					Document methodSet = builder.build(dataFile);
```
LegacyKing: Same here, not just a brace change.

PR #2500, opened 2017-03-08
Still don't add empty string

grimreaper:


LegacyKing:
Not sure I understand why you want to remove the + "" usually people code in things for a reason.

PR #2501, opened 2017-03-08
Add parens

grimreaper:


code/src/java/pcgen/cdom/facet/base/AbstractAssociationFacet.java
```diff
@@ -241,7 +241,7 @@ public int getCount(IDT id)
 	public boolean isEmpty(IDT id)
 	{
 		Map<S, A> componentMap = getCachedMap(id);
-		return componentMap == null || componentMap.isEmpty();
+		return (componentMap == null) || componentMap.isEmpty();
```
LegacyKing: I understand the rest where we're encapsulating the whole formula, but why are we just adding parens here to one half and not everything?  Seems like needless churn if we do it this way, especially if the second half isn't in anything.

PR #2502, opened 2017-03-08
Use standard options for GenericLoader and related

grimreaper:


code/src/java/pcgen/persistence/lst/GenericLoader.java
```diff
@@ -118,7 +106,7 @@ public final T parseLine(LoadContext context, T object, String lstLine,
 		return null;
 	}
 
-	protected void newConstructionActions(LoadContext context, T po)
+	private void newConstructionActions(LoadContext context, T po)
```
thpr: If converting this to private, it should simply be deleted.  At one point this was used for separate lists of items outside the context, but it is no longer used.

thpr:
Approved if checks all pass

PR #2507, opened 2017-03-08
[image] remove unused ico file

grimreaper:


grimreaper:

LegacyKing:
Code folder. @thpr is the one to determine. We change the icon used based upon release status IIRC. Or at least that has been the standard.

PR #2528, opened 2017-03-08
jicon can have reduced scope too

grimreaper:


code/src/java/plugin/notes/gui/JIcon.java
```diff
@@ -232,14 +221,12 @@ protected void launchFile()
 						Logging.errorPrint(e.getMessage(), e);
 					}
 				}
-	
-				//Windows
-				if(SystemUtils.IS_OS_WINDOWS)
+				else if(SystemUtils.IS_OS_WINDOWS)
```
thpr: agree with Codacy here

thpr:
Approved if Travis, Appveyor pass

PR #2530, opened 2017-03-08
[docs] Modernize Java version required

grimreaper:


grimreaper:

LegacyKing:
xdocs? No idea who controls that folder. Need to consult with @thpr since I'm not even sure what that file does...

PR #2537, opened 2017-03-08
Also fix for 'test'

grimreaper:


thpr:
Can't remove new String - that is intentionally driving a value semantic test!!

PR #2541, opened 2017-03-08
Convert to dual language: python 2 and 3

grimreaper:


docs/doc_tools/all_link_anchors_to_lowercase.py
```diff
@@ -20,7 +20,7 @@
 def list_all_html_files():
     html_files = list()
     for root, dirs, files in os.walk("../"):
-        # html_files.extend(filter(lambda x: x.endswith(".html"), files))
+        # html_files.extend(filter(lambda x: x. endswith(".html"), files))
```
LegacyKing: I know Python, and x.endswith should not contain a space, that would not follow convention.

grimreaper:
fixed

docs/doc_tools/all_link_anchors_to_lowercase.py
LegacyKing: Not sure what you "Fixed", the same change is still present.

grimreaper:
re-resolved. Some of my git magic overwrote the fix

PR #2548, opened 2017-03-08
remove unboxing

grimreaper:


LegacyKing:
Not sure what this change is intended to achieve. Can you explain it?

PR #2551, opened 2017-03-08
single character append

grimreaper:


LegacyKing:
I'm not sure I understand the requested change. Any chance you can explain it?

PR #2560, opened 2017-03-08
[image] button1 is never referenced

grimreaper:


grimreaper:

LegacyKing:
I don't know I'm getting tagged for CODE folder items. Anything in the code folder is obviously Code Team purview. Only thing I mess with in the code folder- Junit Tests updates for the various PC's, and the Release Version when updating the versions during a new release. Beyond that, I have nothing that should concern Content/Data. Hit me up if the issues are in system/gamemodes, docs, outputsheets, preview, and of course data.

PR #2561, opened 2017-03-08
remove useless comment

grimreaper:


LegacyKing:
Not a javadoc, looks like a commented out fragment of code. As it has history, not going to mess with it. Javadoc is meant to explain something.

PR #2572, opened 2017-03-08
fix spelling of setUp method

grimreaper:


code/src/utest/pcgen/persistence/lst/CampaignSourceEntryTest.java
```diff
@@ -37,7 +37,7 @@
 	private URI sourceUri;
 	
 	@Before
-	public void setup() throws Exception
+	public void setUp() throws Exception
```
LegacyKing: Is this referenced incorrectly elsewhere? Otherwise, setup() should be valid, as setup is a valid word.

grimreaper: junit looks for setUp by reflection.

PR #2581, opened 2017-03-08
inline variables where possible

grimreaper:


thpr:
Too much of this is taking away useful template in a test system.  Some of it actually breaks tests I have in other branches.  When there is a template in a test, we need to not be in a rush to inline.

PR #2592, opened 2017-03-08
collapse catch blocks; don't extend from Object

grimreaper:


code/src/java/pcgen/cdom/facet/FacetLibrary.java
```diff
@@ -33,12 +32,11 @@
 	
 	private FacetLibrary()
 	{
-		//Do not instantiate
 	}
 
 	private static Map<Class<?>, Object> facets = new HashMap<>();
 
-	public static <T extends Object> T getFacet(Class<T> cl)
+	public static <T> T getFacet(Class<T> cl)
```
thpr: I'm actually curious about this suggested change - is this from a tool saying extending Object isn't necessary?  Because extending Object does actually have meaning - it forces the class provided to getFacet to not be an interface - it has to be a concrete class... so as a result, the call to getBean has specific effects... do we know the effects work if it's not a "hard" class?

grimreaper: This is from an automated linter. I did not actually know about this subtlety with 'extends from Object'. 
I'll drop this.

PR #2601, opened 2017-03-08
Fix NPE when saving character introduced with PR #2254.

Zaister:


grimreaper:
This is a better fix than #2604

grimreaper:
better fix: #2601

grimreaper:
LGTM

Zaister:
Fixes CODE-3144

PR #2604, opened 2017-03-08
Revert "for loop does not loop"

grimreaper:
Reverts PCGen/pcgen#2254

See comments https://github.com/PCGen/pcgen/commit/853653be6ff62e8d7bf99c465a95f0427a1f744e#commitcomment-21204150


grimreaper:
better fix #2601

PR #2607, opened 2017-03-09
Bump a variety of versions

grimreaper:
fb-config, spring, checkstyle, gradle, etc.
also add a couple of useful plugins


thpr:
Why are these needed?  I have spent WAY too much time fixing version bumps and formula is still broken from the last bumps that were made... so I want reasons these are bumped rather than "they are new"

grimreaper:
In general I consider out of date dependencies to bugs worthy of being fixed on their own merit. If in the future we do want to upgrade for a particular feature or bugfix it is far easier to do so if you're already close. In addition upstream changed things for a reason and presumably we want the same things.
I try hard to keep my PRs small and thus bisectable. The issues in the formula system I ended up tracking down to some windows specific behavior.

PR #2608, opened 2017-03-09
DATA-3198 Stunning Fist uses/day

rcapehart:
fixed math stunning fist feat for monk
OLD: StunningFistMonkLVL+(TL-StunningFistMonkLVL)/4
NEW: floor(MonkLVL+(MonkLVL/4))


Pavgran:
Are you sure this is right?
Stunning fist uses per day equals to monk level plus 1/4 of non-monk levels.
In your formula, it is just 5/4 of monk level

data/pathfinder/paizo/roleplaying_game/core_rulebook/cr_feats.lst
```diff
@@ -166,7 +166,7 @@ Step Up											CATEGORY:FEAT		TYPE:Combat
...
...																	ABILITY:Ability Focus|AUTOMATIC|Stunning Fist																																																																																																																																																				BONUS:VAR|StunningFistDC|2|PREABILITY:1,CATEGORY=FEAT,Ability Focus (Stunning Fist)	BONUS:VAR|StunningFistDC|10+(TL/2)+WIS	BONUS:VAR|StunningFistAttack|floor(MonkLVL+(MonkLVL/4))|TYPE=NoStack															SOURCEPAGE:p.135	BENEFIT:You must declare that you are using this feat before you make your attack roll (thus, a failed attack roll ruins the attempt). Stunning Fist forces a foe damaged by your unarmed attack to make a Fortitude saving throw (DC %1), in addition to dealing damage normally. A defender who fails this saving throw is stunned for 1 round (until just before your next turn). A stunned character drops everything held, can't take actions, loses any Dexterity bonus to AC, and takes a -2 penalty to AC. You may attempt a stunning attack once per day for every four levels you have attained (but see Special), and no more than once per round. Constructs, oozes, plants, undead, incorporeal creatures, and creatures immune to critical hits cannot be stunned.|StunningFistDC
```
LegacyKing: I'm a bit leery of this fix. You've hard coded this to actual Monk Level Variable, instead of being neutral and accessible to non-Monks that would increase this. Question I have, why doesn't the other variable suffice that prompts this change from class neutral, to class specific? Is this feat only allowed by a Monk? Do the hybrid classes not get to use this feat at all as well as any archetypes?

I suspect this requires a little more work that dives into a more advanced realm. I suggest this, ABILITY:Internal|AUTOMATIC|Stunning Fist ~ Monk|PREVARGTEQ:MonkLVL,1  Then add that internal ability into the class ability file, and put this formula into it '

BONUS:VAR|StunningFistAttack|MonkLVL|TYPE=Monk
BONUS:VAR|StunningFistLVL|-MonkLVL

Change the 'DEFINE:StunningFistMonkLVL|0' to 'DEFINE:StunningFistLVL|0'
Then fix the formula in the feat - 

BONUS:VAR|StunningFistAttack|floor(MonkLVL+(MonkLVL/4))|TYPE=NoStack
 to
BONUS:VAR|StunningFistAttack|floor(StunningFistLVL/4)|TYPE=NoStack
BONUS:VAR|StunningFistLVL|TL

That should correct the issues. This gives the non-Monk /4 level boost, it makes the Monk specifically cancel out the TL for the MonkLVLs.

Of course, check the rest of the Pathfinder sources to make sure the old variable we just changed is not used, if so, please fix accordingly.

Cheers,

rcapehart: First, I agree that it's not complete. I got it working for monk but am guessing it's broken for non-monks.

When you take an archetype do you no longer get class levels? 

Stunning first is treated two ways, one way for monks and one way for everyone else as far as I understand it.  For Monks its Monk level plus 1/4 monk level rounded down. For everyone else its 1/4 total level. 

Assuming we agree this doesn't work
OLD: StunningFistMonkLVL+(TL-StunningFistMonkLVL)/4
Could we not just change this
NEW: floor(MonkLVL+(MonkLVL/4))
to
floor(MonkLVL+(TL/4))
That should give monk level bonus if you have monk levels and also give the 1/4 bonus everyone including monks get. Depending on how archetypes are treated. 

A quick look seems like only monks get any special circumstances with stunning fist. 

It could be the case that the feat entry could be floor(TL/4) and that something should be done to the class abilities for monk to fix it. 

If what I have said just repeats what you've said and I just need to work on understanding all that data you typed up just let me know. I know I have a lot to learn. 

I'll start working on implementing what you said now! 😄

Pavgran: > For Monks its Monk level plus 1/4 monk level rounded down.

Wrong. Plus 1/4 **non-monk** level rounded down

rcapehart: Why?

rcapehart:
Because I can't read that's why.

data/pathfinder/paizo/roleplaying_game/core_rulebook/cr_feats.lst
rcapehart: Because I can't read. That's why.

LegacyKing: @rcapehart Just checking in, any updates?

rcapehart: Not yet. Life is getting in the way right now. I haven't forgotten about this just haven't had time to get to it. @LegacyKing

rcapehart: I tried to implement what you suggested @LegacyKing but I didn't have any success. I reverted back and started trying to understand what you said better. 

Looking further into the issue it looks like it's on the Unchained Monk side, Standard Monk in my tests seems to give the right number of uses per day, even paired with a non-monk class.

Unchained Monk is giving stunning fist uses equal to (it looks like) 1+TL/4. Which I'm guessing means it's leaving monk levels stuck at 1 when applying Unchained Monk which doesn't make sense because things like BAB and saves are changing the way they're supposed to and they're based on MonkLVL also. 

I'm going to go back to looking and testing, just wanted to update.

PR #2611, opened 2017-03-09
CODE-2017: Don't hide edit menu on mac

grimreaper:
While this should happen in the OSX adapter, there is some brokenness
around this code. In addition this feature is likely not to continue
existing in Java 9.


LegacyKing:
@grimreaper no such Jira ticket...

grimreaper:
bah - I meant CODE-2554

PR #2650, opened 2017-03-11
Make use of common code in DisplayUtilities

grimreaper:


code/src/java/pcgen/core/display/DisplayUtilities.java
```diff
@@ -36,41 +39,20 @@ private DisplayUtilities()
...
-	 * @param separator
-	 *            The separating string
 	 * @return A 'separator' separated String containing the Display Name of the
 	 *         given Collection of CDOMObject objects
 	 */
 	@Deprecated
+	@Contract(pure = true)
 	public static String joinDisplayName(
-		Collection<? extends CDOMObject> cdoCollection, String separator)
+		Collection<? extends CDOMObject> cdoCollection)
```
thpr: This is one of those places where I think in-lining is questionable and requires some thought in the process.... under what circumstances do you not in-line?

thpr:
Approved if checks pass

PR #2651, opened 2017-03-11
Make use of a standard Comparator instead of a custom one

grimreaper:


thpr:
Approved if checks pass

PR #2661, opened 2017-03-11
Fix itest

grimreaper:
I went a little overboard with reducing scope and visibility;
we really need to automate slowtest. 

Also the "nicer" map calls do not work if the map can correctly contain null values. We should fix this as the root cause eventually, but get tests to work again.


LegacyKing:
@thpr I believe this is supposed to be the fix PR

PR #2664, opened 2017-03-12
Bug fix:Data-3162 Eye for Plunder trait 

cavalancian:
Bonus to perception should be a situational bonus


LegacyKing:
Best to create feature branches. You should not be doing fixes from your own master. It'll get messy before you realize it.

PR #2665, opened 2017-03-12
Ensure line endings and permissions are correct for pcgen.sh

eslafgh:
I think this cleaner that the method used in #2381

Tested it on linux but don't have access to a windows box to try it on.


grimreaper:
LGTM

LegacyKing:
@thpr can we get a quick review on this?

javydreamercsw:
Looks straight forward, just need to be tested really. Commited a related change in 7ddf5e47693c62a0bdc7fb6afb4025d0b318f695

PR #2676, opened 2017-03-12
Remove extranious '*' from javadocs (id: 5:19939)

grimreaper:


LegacyKing:
@grimreaper Conflicted

PR #2682, opened 2017-03-12
Remove extranious '*' from javadocs (id: 4:16787)

grimreaper:


LegacyKing:
@grimreaper conflicted

PR #2696, opened 2017-03-17
Bloodrager Archetype Support

evilpixie87:
Fixes to datacontrols and globalvar to better support the bloodrager's archetypes


LegacyKing:
Massive changes, and adding support for a system that is not being used for these classes. i.e. ARCHETYPE Abilities should never, ever use the TYPE system for the Hybrids, so adding the support for TYPE instead of using only FACTs is a waste of space.  Also, you're missing the sub-selection changes, which would impact the features in the abilities_class.lst

PR #2697, opened 2017-03-18
Support for Bloodrager Archetypes

evilpixie87:
Trying again. 12 lines added to DataControls, and 12 to GlobalVar.

This SHOULD allow alteration/removal of individual bloodline powers.


grimreaper:
This still looks like the entire file was modified.

PR #2700, opened 2017-03-19
Use modern TravisCI Infrastructure

grimreaper:
This moves us to using the TravisCI container based infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,18 @@
 # Run tests run on travis-ci after each commit and before each merge
 
 language: java
+dist: trusty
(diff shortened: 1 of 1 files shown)


LegacyKing:
And we're now conflicting... please address @grimreaper

grimreaper:
The packages shouldn't be changing anything - it just changes the way they get installed.

LegacyKing:
@grimreaper fix

grimreaper:
this needs to be reworked a bit now. Closing until I have time to handle it.

PR #2714, opened 2017-03-22
[OS] Avoid use of empty paragraph tag

grimreaper:


LegacyKing:
I suppose I should be thankful for the 4e sheets getting lovin... Sad we have nothing to really test against these sheets...

PR #2717, opened 2017-03-22
[system] collapse empty XML tags

grimreaper:


LegacyKing:
LGTM. Only nit - why are we losing our indents on some of these tags? XML is all about properly being able to nest tags inside, but if the indents are off, that nesting doesn't look correct.

PR #2721, opened 2017-03-24
[gradle] and and -> and

grimreaper:


grimreaper:
@LegacyKing can I ask you to merge this so there is less for future code reviews?

PR #2743, opened 2017-03-26
Reduce indirection of REFERENCE_SORTER

grimreaper:


thpr:
Approved if checks pass

PR #2762, opened 2017-03-29
Data fixes

vincent-lhote:


LegacyKing:
Looks good, but in the future, can we separate these out. Easier to review one set of changes, then three different changes.  And consider rebasing if you're pulling in upstream updates, that will keep the commit for upstream from appearing in the commit list. (cleaner PRs is always nice).

PR #2763, opened 2017-03-30
Remove filename from comments

grimreaper:


LegacyKing:
Coding best practices I keep getting taught is the file name is part of the beginning comments. C#, C++, C, Java, Python. Not sure why we're dumping that out just to save a line.  Other than a personal preference, is there any reason to toss out what is industry best practices?

grimreaper:
> is there any reason to toss out what is industry best practices?
I have no idea where you're learning that from. There is virtually no industry code I've worked with that keeps the filename in the comments. Files get copied, renamed, moved, or changed all the time.

PR #2779, opened 2017-04-04
Fix minor language mistakes in the documentation.

goeranu:
While reading through the documentation, I stumbled on a number of minor editorial mistakes for which I submit fixes here.

In addition, I removed the section on running on Fedora.  Fedora 7 is very out of date, and on current Fedoras (24 and 25) PCGen works just fine with the standard Java (OpenJDK 1.8.0).


LegacyKing:
Thanks for the contribution. https://www.hipchat.com/gzYdCc3Vn if you want to join other discussions or would like suggestions for other areas we could use your eye for details. :+1:

PR #2791, opened 2017-04-11
PFS update for People of the North

evilpixie87:
Updated PotN files for PFS


LegacyKing:
pcgen.persistence.lst.DataTest > missingFilesTest FAILED
    org.junit.ComparisonFailure: Some data files are missing. expected:<[]> but was:<[Missing file pathfinder/paizo/player_companion/people_of_the_north/_pfs/people_of_the_north.jpg used by pathfinder/paizo/player_companion/people_of_the_north/_pfs/_.pcc<br>
    ]>
        at pcgen.persistence.lst.DataTest.missingFilesTest(DataTest.java:200)

mjmeans:
I looked over this PR.. PFS support should be implemented exclusively within the _pfs folder under its parent source.

Add EXCLUDEs to the _pfs\_.pcc file for those parent sources that should be completely excluded in PFS.

For items that need to have a different behaviour, add a _pfs\*.lst file that includes the .MODs, .COPYs, etc that need to be applied to the original source in order to make it PFS complaint. Then add its reference in the _pfs\_.pcc file with a PRECAMPAIGN tag, like below. This PRECAMPAIGN is needed because the entire _pfs\_*.pcc file will be scanned whether or not the parent source is actually loaded and we need to make sure the PFS specific changes are only loaded if the parent source was loaded.
     ABILITY:pfs_{file}.lst|PRECAMPAIGN:1,INCLUDES={original source's key}

The only change to the parent source will be to add |Pathfinder_PFS to it's GAMEMODE.  ...   THERE SHOULD BE NO OTHER CHANGE TO THE PARENT SOURCE UNLESS 1) There is no other way to make it PFS compatible, AND 2) You are absolutely sure that the change will not affect any -PFS character now or in the future.

Lastly, you need to add the new source in data\pathfinder_pfs\paizo\pathfinder_society\roleplaying_guild_guide\_roleplaying_guild_guide.pcc.

The above should probably be added to the PFS docs for PCGen.

PR #2793, opened 2017-04-11
DATA-3206

rcapehart:
uc_abilities_class
changed gunsmith
added gunslinger battered gun
uc_abilitycategories
added gunslinger battered gun
uc_equip_arms_armor
added gunslinger starting battered guns


LegacyKing:
I'm sure the implementation makes perfect sense. I don't quite follow it, but as long as it works for you, then I'll accept it. But for when I have had rest, what is it accomplishing? 

For all that is happening, it seems a simple CHOOSE:STRING would have worked just as well...

rcapehart:
I was trying out a CHOOSE with a template and kits but having trouble with it. I suggested maybe it be selection under class options but didn't know how to get it there so you told me to set up an ABILITYCATEGORY for that. I got that to work. 

I'm absolutely sure there's other, and even better ways to do it. I was able to get this to work and it made sense to me. 

I'm not certain what you're asking though by "what is it accomplishing?" I can say, I had a player at my table that interpreted it to mean any pistol when it says choose from pistol, musket, or blunderbuss. That's not the case and giving the user a place to select that battered (broken to everyone but the gunslinger - something I didn't try to make apply to the equipment line) starting firearm is useful. 

I'm certainly willing to rework it, but will need to be pointed in the right direction.

PR #2796, opened 2017-04-12
FREQ: DATA-3228 Method of bypassing PRESTAT requirements (Core Rulebo…

LegacyKing:
…ok Feat and Equipment only)


LegacyKing:
@evilpixie87 review the implementation, and feel free to expand to other sources. I'll merge this in tomorrow. Note, the CombatFeatInt is left as that is a specific instance, and not global. This change is only for Global changes. And happy to say, this should transition to the FORMULA SYSTEM nicely.

PR #2805, opened 2017-04-13
Artful dodge fix DATA-3226

dude-rtfm:
All Paizo feats done


LegacyKing:
Looks good, and I actually have nothing against it. But to be brutally honest, you created a lot more work than was necessary. No need to introduce a new variable, when the one I just introduced would have handled this nicely. I'll merge it in after I finish review the last few files.

PR #2839, opened 2017-05-17
Chronicles of the righteous pfs support

evilpixie87:
Fixed a typo


evilpixie87:
Andrew.... I ran this 5 times, 3 different ways. No errors in my log. Please look at it.

PR #2843, opened 2017-05-19
UC_abilities_class moved katana to weapon proficiency in ninja class …

kypvalanx:
…feature

* removed Exotic Weapon Proficiency (Katana)
* tested that weapon focus Katana could still be taken


LegacyKing:
Can you cite the rule where Ninja's cannot use Katana's one-handed?  I need that answered before I'm going to remove that from the class. (That's what your change is doing, forcing Ninja's to only use a Katana Two-handed).

kypvalanx:
I did not realize that this was how it was being implemented, i assumed that the proficiency worked as the exotic proficiency since it was called out rather than being grouped as a marshal proficiency.

I would expect that the existing implementation is correct based on [this](http://paizo.com/paizo/faq/v5748nruor1fm#v5748eaic9o78).

PR #2863, opened 2017-06-04
Move gradle output into a dedicated directory.

grimreaper:
Gradle and windows do not seem to play nice with each other in ways I
don't fully understand.

This moves pcgen into a dedicated output directory instead as this is
better practice in the first place. We may need some slight changes to
other release scripts but that will happen in due course

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -224,7 +224,7 @@ task copyToRoot(type: Copy, dependsOn: [copyToLibs, jar, converterJar]) {
     from "$buildDir/libs/pcgen-${version}-batch-convert.jar"
     from "$buildDir/release/pcgen.exe"
     from "$projectDir/code/pcgen.sh"
-    into "$projectDir"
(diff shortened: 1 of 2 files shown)


LegacyKing:
What exactly is going into the output folder?

grimreaper:
pcgen-batch-convert.jar  pcgen.jar                pcgen.sh

LegacyKing:
Okay, release on Windows is working - at least the clean build cut and dry. How will this impact the actual bundle of releases though?  Running through my step by step process. If all goes well, I'll actually do a 6.07.02 tonight...

grimreaper:
I don't know the exact process for doing releases. Can you tell me what you run and I'd be happy to test/fix.

LegacyKing:
The process is on the wiki (look under release)... Okay, looks like if I do it on my clean repo it'll work. It barfed because I have a lot of unfinished work putzing about.

PR #2919, opened 2017-06-18
Bump fop 2.2; removed unneeded dep

grimreaper:


LegacyKing:
Will this fix the broken portrait for the exports?

grimreaper:
Probably not because I can't reproduce. Portrait export works just fine for me. 
Point me to the CODE bug (again?) please.

PR #2922, opened 2017-06-18
bump pmd to latest version

grimreaper:


code/standards/pmdruleset.xml
```diff
@@ -69,7 +69,5 @@
   <rule ref="rulesets/java/unnecessary.xml">
     <exclude name="UselessParentheses" />
   </rule>
-  <rule ref="rulesets/java/unusedcode.xml">
-    <exclude name="UnusedModifier" />
```
LegacyKing: Is this no longer necessary?

grimreaper: It doesn't work with the newest version, and to properly use the reports we need to merge some of my other code PRs that fix the underlying issues anyways.

PR #2923, opened 2017-06-18
Add configuration files for codacity

grimreaper:
Its kind of ridiculous that codacity doesn't have a way to point it to
the real files without including them in the project root. Its
questionable if we need codacity in the first place, but given we have
it, lets at least make it more useful than it otherwise would be.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,0 +1,1 @@
+code/standards/checkstyle.xml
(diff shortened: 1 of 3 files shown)


LegacyKing:
I'm confused, looks like you added a path in the files, is that all that is required? I was under the impression the root directory in our repo was pcgen...

grimreaper:
I added symlinks to the real files.

PR #2937, opened 2017-06-24
corrected class spell list associations

net-diver:


LegacyKing:
@evilpixie87 Please review this.

PR #2939, opened 2017-06-24
Wyvaran race support

evilpixie87:
Modified wyvaran race to support new alternate racial traits.


evilpixie87:
Missed changes at some point. Fixed changes to sortkeys...

PR #2966, opened 2017-06-28
corrected type listing

net-diver:


LegacyKing:
This appears to be from the later merged PR. Closing as it's redundant.

PR #2991, opened 2017-07-07
[CODE-3185] Fix the top(N) for the dice roller

kheinzelmann:
When top(N) is specified in the roll(String) method take the top(N) and not the bottom(N) dice.

### DIFF
--- a/code/src/java/pcgen/core/RollingMethods.java
+++ b/code/src/java/pcgen/core/RollingMethods.java
@@ -216,7 +216,7 @@ public final class RollingMethods
         return IntStream.generate(() -> roll(sides - reroll) + reroll)
             .limit(times)
             .sorted()
-            .limit(numToKeep)
+            .skip(times-numToKeep)
             .sum();
     }
 
--- a/code/src/java/pcgen/base/util/RandomUtil.java
+++ b/code/src/java/pcgen/base/util/RandomUtil.java
@@ -32,7 +32,7 @@ public final class RandomUtil
 {
 
 	/** this is used by the random selection tools */
-	private static final Random RANDOM = new Random(System.currentTimeMillis());
+	private static Random random = new Random(System.currentTimeMillis());
 
 	private RandomUtil()
 	{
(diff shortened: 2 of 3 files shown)


grimreaper:
Can we add a test case for this too ?

code/src/java/pcgen/core/RollingMethods.java
```diff
@@ -216,7 +216,7 @@ private static int roll(
         return IntStream.generate(() -> roll(sides - reroll) + reroll)
             .limit(times)
             .sorted()
-            .limit(numToKeep)
+            .skip(times-numToKeep)
```
grimreaper: Another option is to keep this line and replaced .sorted() with `.sorted(Comparator.reverseOrder())`

kheinzelmann: I tried this it does not work with IntStream objects, at least not for me.

kheinzelmann:
I would be willing to attempt adding a test. Do you wont me to add it as another commit to this pull request or open a new pull request?

grimreaper:
> I would be willing to attempt adding a test. Do you wont me to add it as another commit to this pull request or open a new pull request?

I'd prefer another commit on this PR so its more obvious what the test is testing, but that won't block the merge for me.

code/src/java/pcgen/core/RollingMethods.java
grimreaper: not a big deal either way. They are similarly readable to me.

kheinzelmann:
Thinking about this I'm not entirely sure how to add a test fixture for this. I would either need to inject the dice rolls as fixed numbers or be able to retrieve the list of generated random numbers. I could possibly cheat if I could set the random number generator seed but that is not currently exposed.

grimreaper:
It would be a good idea to expose the random seed as a Dependency Injection in the ctor. Another option is to look at the existing `DiceTest.randomDiceAssertions` to see how else it mght be done.

kheinzelmann:
DiceTest appears to only check if the dice rolls in the correct range.
RandomUtils is a static class but I should be able to inject via a setter. Let me see what I can do without making a mess of this.

PR #3001, opened 2017-07-07
don't call new string

grimreaper:


thpr:
This is intentional to force a value test

PR #3005, opened 2017-07-07
Remove non-javadoc comments and turn them into Javadocs

grimreaper:
This is a prep PR for reducing style errors relating to lack of Javadocs


thpr:
I thought we were just removing all the "see" stuff since the docs would be inherited by javadoc if it overrides...?

grimreaper:
We are. I think this is one of those PRs that I had changed, and then done another pass later cleaning it up. Going to drop this one and just combine them.

PR #3013, opened 2017-07-07
Still don't add empty string

grimreaper:


thpr:
No way this should all be in one commit - way too much with a core class, especially when the PR claims it's about adding empty strings

PR #3023, opened 2017-07-07
remove transient from non-serializeable classes

grimreaper:


code/src/java/pcgen/cdom/content/KnownSpellIdentifier.java
```diff
@@ -146,28 +147,24 @@ public Integer getSpellLevel()
...
 			{
 				returnList.addAll(pc.getSpellsIn(list, spellLevel));
-			}
+			});
 			return returnList;
 		}
-		List<Spell> spellList = new ArrayList<>();
-		for (Spell sp : ref.getContainedObjects())
+		Collection<Spell> spellList = new ArrayList<>();
+		ref.getContainedObjects().forEach((Spell sp) ->
```
thpr: Another place where we should talk about how to make this simpler.

grimreaper: This is one where the name does not match the content. I suspect this was the result of a bad merge. Closing.

PR #3025, opened 2017-07-07

grimreaper:


code/src/test/plugin/PluginBuildTest.java
```diff
@@ -60,6 +60,7 @@ public static Test suite()
 	/**
 	 * Check for the presence of all 'bonus' token plugins.
 	 */
+	@org.junit.Test
```
thpr: Why fully articulated when basically none of our other tests do that?

PR #3027, opened 2017-07-07
jicon can have reduced scope too

grimreaper:


code/src/java/pcgen/io/Compatibility.java
```diff
@@ -83,8 +82,7 @@ static PCTemplate getTemplateFor(PCTemplate template,
 		return null;
 	}
 
-	@NotNull
-	public static String getKeyFor(PCTemplate pct)
+	public static @NotNull String getKeyFor(PCTemplate pct)
```
thpr: Are these equivalent?  I thought outside meant it applied to parameters and return?

PR #3061, opened 2017-07-08
Make use of system images

grimreaper:


thpr:
Approved if checks pass

PR #3093, opened 2017-07-19
NEWTAG-595 Add getFact Function to new formula system

thpr:


code/src/java/plugin/function/GetFact.java
```diff
@@ -0,0 +1,245 @@
...
+			return null;
+		}
+	}
+
+	private FormatManager<?> allowThreeArguments(SemanticsVisitor visitor,
+		FormulaSemantics semantics, Node[] args)
+	{
+		//Turn scope node into a scope name
+		Node scopeNode = args[0];
+		if (!(scopeNode instanceof ASTQuotString))
```
grimreaper: I don't know enough about how this functionality works but is there any way to pull these into separate classes to avoid the need for instanceof?

thpr: Right now, no, there is no way to have two functions with the same name but different formats allowed as args.  That would require much deeper changes back in the formula system to change that assumption.  That's probably a useful discussion, but longer term change no matter what.  Second, this is using instanceof in order to have a useful error message, not just crash due to an invalid cast.  Third, we don't want to do a double dispatch because right now ASTQuotString is entirely generated by JavaCC/JJTree and thus it is beneficial to avoid too many modifications to that class (and its peers) so as to ease forward compatibility if for some reason the class wants to be changed by the automated tools.  Lastly, we can't defer to .jjtAccept as would normally be done because we want to strictly enforce a *direct* String, not a String brought in by a variable.  I think that answers all of the code best practices that this is not using :)

thpr:
Replaced with PR 3258

PR #3094, opened 2017-07-19
NEWTAG-597 Add Dice Format

thpr:


thpr:
This is approved even with Codacy complaints since they are all in utest code.

LegacyKing:
It's been a month, can we merge this?

LegacyKing:
@grimreaper can check this out... but I'm merging this and the CDOM fix... hoping that will correct the GETFACT error we are having. If not, will need to investigate.

LegacyKing:
Apparently it won't build... reverting.

PR #3105, opened 2017-07-20
expand rulesets/java/design.xml

grimreaper:


thpr:
Please reconcile with 3194 and only submit 1 PR

PR #3108, opened 2017-07-21
Prefer native functions over helpers for BigDecimal

grimreaper:


thpr:
Approved if checks pass

thpr:
So I think this exposes why BigDecimalHelper exists:
junit.framework.ComparisonFailure: Expected 0th uparsed item to be equal expected:<3[150]|Any> but was:<3[.15E+3]|Any>
The problem is that trim trailing zeros by default is happy to turn 10 into 1E+1, and we want it human readable, so we have a method that ensures the scale is never set to higher than zero.

grimreaper:
makes sense. closing. 
(this was a part of a series of other PRs, the pre-reqs likely didn't go through)

PR #3109, opened 2017-07-21
pull out overridden abstract functions that remain abstract

grimreaper:


thpr:
Approved if checks pass

PR #3110, opened 2017-07-21
continue making use of collectionutils

grimreaper:


code/src/java/plugin/lsttokens/spell/CompsToken.java
```diff
@@ -95,7 +96,7 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
 		{
 			sb.append(Constants.LST_DOT_CLEAR);
 		}
-		if (added != null && !added.isEmpty())
+		if (!CollectionUtils.emptyIfNull(added).isEmpty())
```
thpr: This seems questionable to me.  This trades a one-operation command (null check) for a method call, returning an object, and then checking it's empty.  I also question whether this is actually clearer...

grimreaper:
I think I was going somewhere with this and dropped it mid-way. Sorry for the noise.

PR #3111, opened 2017-07-21
Empty collection parsing

grimreaper:


code/src/java/pcgen/util/NullUtils.java
```diff
@@ -0,0 +1,5 @@
+package pcgen.util;
```
thpr: I'm sorry, what is this file?

code/src/java/pcgen/persistence/GameModeFileLoader.java
```diff
@@ -408,7 +409,7 @@ private static GameMode loadGameModeMiscInfo(String aName, URI uri)
 		gameMode.setFeatTemplate(feat);
 
 		int[] dieSizes = gameMode.getDieSizes();
-		if (dieSizes == null || dieSizes.length == 0)
+		if (ArrayUtils.nullToEmpty(dieSizes).length == 0)
```
thpr: Again, I think these need discussion.  This seems like churn and opinion, not changes that would be universally valuable.

PR #3112, opened 2017-07-21
move comparitor into lambda

grimreaper:


thpr:
I feel like this is already in another PR, but either way, approved if checks pass

PR #3113, opened 2017-07-21
Remove empty static init with incorrect info abot JTreeTable

grimreaper:


thpr:
Approved if checks pass

PR #3120, opened 2017-07-23
Simplify PrimitiveUtilities and ChooseInformationUtilities

grimreaper:


thpr:
I like the conversion to streams, but let's change the contract on these to the collection cannot be null... can you do that and then extend that enforcement onto ChooseInformation.composeDisplay?

grimreaper:
Removed the emptyIfNull checks and added contracts & annotations. Good now?

PR #3122, opened 2017-07-25
make use of common code in DisplayUtilities

grimreaper:


thpr:
Rewrite of utility method to remove utility is ... controvertial

PR #3124, opened 2017-07-25
Remove bad copyright info

grimreaper:


thpr:
You need to talk to Andrew about these

grimreaper:
cc @LegacyKing

LegacyKing:
I added myself to make these files correct for the license. Hence the disclaimer at the front.

PR #3136, opened 2017-07-25
instanceof

grimreaper:


code/src/java/pcgen/core/DataSet.java
```diff
@@ -298,7 +298,7 @@ private void createGearBuySellSchemes()
     @Override
 	public List<AbilityFacade> getPrereqAbilities(AbilityFacade abilityFacade)
 	{
-		if (abilityFacade == null || !(abilityFacade instanceof Ability))
+		if (!(abilityFacade instanceof Ability))
```
thpr: What's the basis for this change?

PR #3137, opened 2017-07-25
create BigDecimal from string

grimreaper:


code/src/java/plugin/lsttokens/deprecated/TemplateFeatToken.java
```diff
@@ -165,7 +165,7 @@ public void applyChoice(CDOMObject owner, CNAbilitySelection choice,
 					.doubleValue();
 			if (cost > 0.0001)
 			{
-				pc.adjustAbilities(AbilityCategory.FEAT, new BigDecimal(cost));
+				pc.adjustAbilities(AbilityCategory.FEAT, new BigDecimal(String.valueOf(cost)));
```
thpr: I understand the general theory here, you don't want folks to do new BigDecimal(3.1), you want new BigDecimal("3.1")... but in this case we already have cost as a value, I don't know why we would go through a String?  What value does this really add?

PR #3140, opened 2017-07-25
fix errors documented in rulesets/java/design.xml

grimreaper:


code/src/testcommon/plugin/lsttokens/testsupport/TokenRegistration.java
```diff
@@ -63,7 +64,7 @@ public static void register(PrerequisiteParserInterface ppi)
 		}
 	}
 
-	public static final Set<LstToken> tokenSet = new HashSet<>();
+	private static final Collection<LstToken> tokenSet = new HashSet<>();
```
thpr: We still need to talk on these.  I don't find making Set universally into collection to be semantically useful... I feel it removes information about programmer intent, which actually makes it a negative change, so I don't buy into tools that automatically feel generic is good.

PR #3148, opened 2017-07-25
enum values should not use .equals()

grimreaper:


thpr:
Test indicates behavior has changed.  == has different null behavior than .equals.

PR #3149, opened 2017-07-25
extend from better common class

grimreaper:


thpr:
Semantics are important, these mislead

PR #3158, opened 2017-07-25
remove unused code in LstLoadeR

grimreaper:


code/src/java/pcgen/core/GameMode.java
```diff
@@ -593,12 +593,7 @@ public XPTable getLevelInfo(final String xpTableName)
 	 */
 	public void addLevelInfo(final String xpTableName, final LevelInfo levInfo)
 	{
-		XPTable xpTable = xpTableInfo.get(xpTableName);
-		if (xpTable == null)
-		{
-			xpTable = new XPTable(xpTableName);
-			xpTableInfo.put(xpTableName, xpTable);
-		}
+		XPTable xpTable = xpTableInfo.computeIfAbsent(xpTableName, XPTable::new);
```
thpr: This is not the same - it destroys the name in the XPTable

PR #3164, opened 2017-07-25
Remove three rules StringToString InefficientEmptyStringCheck StringInstantiation

grimreaper:


thpr:
Changes too much in one place - including controversial items and with one memory leak included

PR #3165, opened 2017-07-25
remove unused variable

grimreaper:


code/src/java/plugin/notes/gui/NotesTreeNode.java
```diff
@@ -797,7 +795,7 @@ public NotesTreeNode createChild(String newName)
 			populate();
 		}
 
-		while (notDone)
+		while (true)
```
thpr: I highly suspect this is NOT the intended design... there is a bug in this code and the tool is catching a symptom not the root cause.

PR #3175, opened 2017-07-25
remove DisplayUtilities.java which was entirely deprecated and now unused

grimreaper:


code/src/java/plugin/exporttokens/deprecated/WeaponProfsToken.java
```diff
@@ -58,8 +55,9 @@ public String getToken(String tokenSource, CharacterDisplay display,
 	{
 		if (SettingsHandler.getWeaponProfPrintout())
 		{
-			return DisplayUtilities.joinDisplayName(
-				display.getSortedWeaponProfs(), ", ");
+			return StringUtils.join(
```
thpr: Use a collector, not StringUtils... this is NOT joining what you expect it to join...

PR #3180, opened 2017-07-25
Squashed commit of the following:

grimreaper:


code/src/java/pcgen/util/Comparators.java
```diff
@@ -27,25 +27,67 @@
...
 
-	private static final ToStringComparator tSC = new ToStringComparator();
-	private static final ToStringIgnoreCaseComparator tSICC = new ToStringIgnoreCaseComparator();
-	private static final ToStringIgnoreCaseCollator tSICCol = new ToStringIgnoreCaseCollator();
-	private static final IntegerComparator iC = new IntegerComparator();
-	private static final NumberComparator nC = new NumberComparator();
-	private static final DateComparator dC = new DateComparator();
-	private static final HashCodeComparator hCC = new HashCodeComparator();
-	private static final TreeTableNodeComparator treeNodeComp = new TreeTableNodeComparator();
+	private static final Comparator<Object> tSC =
```
thpr: This is now gibberish.

The previous class name was semantically useful, limiting it to abbreviated variable names with anonymous Comparators is not.  The names weren't great before, but this patch eliminates much of the semantic information we had.  Can we make these better names?

PR #3181, opened 2017-07-25
invert if

grimreaper:


code/src/java/pcgen/gui2/tabs/models/ConcurrentDataView.java
```diff
@@ -78,22 +78,21 @@ public ConcurrentDataView()
...
+			return dataMap.get(obj);
+		}
+		else
 		{
 			try
 			{
 				return future.get();
-			}
-			catch (InterruptedException | ExecutionException ex)
+			} catch (InterruptedException | ExecutionException ex)
```
thpr: Settings on braces still needs work

grimreaper: ya, you could tell how old this commit is :)

PR #3182, opened 2017-07-25
remove some more simple ones

grimreaper:


thpr:
Disagree on startsWith.

charAt(0) can cause exceptions where startsWith will not.  They are NOT logically equivalent

PR #3183, opened 2017-07-25
remove useless default

grimreaper:


thpr:
Injects a bug into the code

PR #3205, opened 2017-07-26
remove unused variables

grimreaper:


thpr:
Vast majority of these are not unused

PR #3210, opened 2017-07-26
move processing around so that we can use removeIf

grimreaper:


thpr:
Needs a fix for compile error

PR #3211, opened 2017-07-26
use removeIf instead of a loop

grimreaper:


thpr:
Error needs to be fixed

PR #3218, opened 2017-07-26
Add migration

grimreaper:


code/src/java/pcgen/system/ConfigurationSettings.java
```diff
@@ -90,19 +85,6 @@ public static void setCountry(String country)
...
 
-	/**
-	 * @return the current user directory
-	 */
-	public static String getUserDir()
-	{
-		return SystemUtils.USER_DIR;
-	}
-
-	public static ConfigurationSettings getInstance()
```
thpr: I don't  get this - why spread lots of nulls rather than having a convenience method?  What's the logic here?

PR #3220, opened 2017-07-26
reduce the number of warnings to fix

grimreaper:


code/src/java/plugin/overland/model/RoomBoardFactory.java
```diff
@@ -69,10 +72,12 @@ public static RoomBoard load(File dataDir)
 
 					methodSet = null;
 					dt = null;
-				}
-				catch (Exception e)
+				} catch (Exception e)
```
thpr: braces

code/src/java/plugin/overland/model/TravelMethodFactory.java
```diff
@@ -89,7 +95,7 @@
 		if (path.isDirectory())
 		{
 			File[] dataFiles = path.listFiles(new XMLFilter());
-			SAXBuilder builder = new SAXBuilder();
+			SAXEngine builder = new SAXBuilder();
```
thpr: Don't understand the semantic value to a reader here, all I see is a tool saying "do this" without applying thought to the process.

PR #3223, opened 2017-07-26
non-public ought not to be public

grimreaper:


thpr:
Let's talk about this too - this is one of the last things I ever want to see a tool enforcing - package is of questionable value in Java for a number of historical reasons, so AFAIK, the world is composed of public, protected, private... and the semantics are important.  If it's limited to package in code, it's often done for a reason, and I'm concerned these changes are still removing information from code

PR #3234, opened 2017-07-26
duplicate if condition

grimreaper:


code/src/java/pcgen/core/pclevelinfo/PCLevelInfo.java
```diff
@@ -359,8 +359,6 @@ public boolean equals(Object o)
 				&& skillPointsRemaining == other.skillPointsRemaining
 				&& classKeyName.equals(other.classKeyName)
 				&& ObjectUtil.compareWithNull(statsPreModified,
-					other.statsPreModified)
-				&& ObjectUtil.compareWithNull(statsPreModified,
```
thpr: Since this is .equals, this probably needs to check statsPostModified, not remove a duplicate of Pre

grimreaper: Good point; fixed

PR #3235, opened 2017-07-26
extranious ctor

grimreaper:


code/src/java/pcgen/core/PCClass.java
```diff
@@ -116,14 +116,6 @@
 	private String classKey = null;
 
 	/**
-	 * Default Constructor. Constructs an empty PCClass.
-	 */
-	public PCClass()
```
thpr: ...and this is why automated tools don't work... and why one of those things that is on the list is "we really should have architectural tests"... "Loadable" interface effectively requires a zero-constructor interface.... items are built through reflection.

PR #3256, opened 2017-07-29
Improve Output

thpr:
Fixes a bug in PC's clone method (duped items added to facets in
constructor)

Enhances output to check view of CDOMObjects

Escapes Strings on output (does it universally - may need a better
method since old export system didn't do it on equipment)


LegacyKing:
Approved to cause regression in Goldielocks test. This has shown existing Data Code has a race condition which the Formula System will address.

thpr:
Replaced by PR #3321

PR #3268, opened 2017-07-30
Migrate from xmlunit 1.x to 2.x+legacy

grimreaper:


grimreaper:
We should eventually migrate using this guide, but this lets us get some of the gains (speed and bug fixes) without going through the migration cost
https://github.com/xmlunit/user-guide/wiki/Migrating-from-XMLUnit-1.x-to-2.x

PR #3271, opened 2017-07-30
Warpriest Channel Damage Fix update

evilpixie87:
Originally replaced faulty formula with variable from previous ability that channeling damage calls back to in the book. Later realized that variable may not carry forward when using certain archetypes. Replacing variable with correct formula from earlier ability.


evilpixie87:
Conflict is because of reworked, but still flawed commit already merged!

PR #3273, opened 2017-07-30
Remove 'xalan' library

grimreaper:
We no longer make use of it

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,7 +154,6 @@ dependencies {
     compile group: 'jep', name: 'jep', version:'2.4.1'
     compile group: 'org.freemarker', name: 'freemarker', version:'2.3.26-incubating'
     compile group: 'org.jdom', name: 'jdom2', version:'2.0.6'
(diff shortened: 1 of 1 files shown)


LegacyKing:
@grimreaper are we absolutely certain of this?  I've been discovering broken features in data programming Starfinder and I suspect some of the changes made have been the culprit.

eslafgh:
Its still required, there is a direct dependency on xalan in the pcgen code.

```
$ jdeps -cp $(echo libs/* | tr ' ' :) output/pcgen.jar | grep xalan
pcgen.jar -> libs/xalan-2.7.2.jar
      -> org.apache.xml.utils                               xalan-2.7.2.jar
$ git grep org.apache.xml.utils
code/src/java/pcgen/rules/persistence/token/AbstractToken.java:import org.apache.xml.utils.XMLChar;
```

The change passes all test because xalan is still present as its pulled in by fop.

grimreaper:
Thanks for the info!

PR #3275, opened 2017-07-30
Tighten up some code quality options

grimreaper:
This does the following
- bumps checkstyle to 8.1 (mainly new checks)
- enables all clone related checks, and fixes reported errors
- makes some gradle lint warnings fatal
- fix some minor brace related formatting issues

### DIFF
--- a/code/src/java/pcgen/core/PlayerCharacter.java
+++ b/code/src/java/pcgen/core/PlayerCharacter.java
@@ -275,6 +275,8 @@ import pcgen.util.Delta;
 import pcgen.util.Logging;
 import pcgen.util.enumeration.AttackType;
 import pcgen.util.enumeration.Load;
+
+import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.TestOnly;
 
 public class PlayerCharacter implements Cloneable, VariableContainer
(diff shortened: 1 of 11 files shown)


code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -5512,7 +5514,8 @@ public boolean loadDescriptionFilesInDirectory(final String aDirectory)
 				                          } catch (IOException exception)
```
thpr: Catch needs proper brace location

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -7265,17 +7264,32 @@ public PlayerCharacter clone()
...
 		{
-			PCLevelInfo newLvlInfo = info.clone();
-			aClone.levelInfoFacet.add(aClone.id, newLvlInfo);
+			for (PCLevelInfo info : getLevelInfo())
+			{
+				PCLevelInfo newLvlInfo = info.clone();
+				aClone.levelInfoFacet.add(aClone.id, newLvlInfo);
+			}
+		}
+		catch (CloneNotSupportedException e)
```
thpr: Why are we catching and subsuming this error rather than letting it flow up the stack?

grimreaper: Since this is a checked exception and we'd need to add this to the signature for a long length. Another option is to catch and throw a different exception.

thpr: This should not be subsumed.  Those who call clone should know the risks... besides it's called in two places... not  that hard to address it where it's used.

grimreaper:
going to split this out and look into a nicer way of handling the exceptions

PR #3335, opened 2017-08-06
Global import ordering

grimreaper:


code/src/java/gmgen/plugin/PcgSystemInitiative.java
```diff
@@ -20,14 +20,15 @@
...
 
-import gmgen.plugin.dice.Dice;
 import pcgen.cdom.util.CControl;
 import pcgen.cdom.util.ControlUtilities;
 import pcgen.core.Globals;
 import pcgen.core.PCStat;
 import pcgen.core.PlayerCharacter;
 import pcgen.core.display.CharacterDisplay;
 
+import gmgen.plugin.dice.Dice;
```
thpr: Your config file currently reads gmgem not gmgen

karianna:
@grimreaper Can you take a look at @thpr feedback?

code/src/java/gmgen/plugin/PcgSystemInitiative.java
LegacyKing: I don't see this GMGem reference here... please verify?

LegacyKing:
This looks good to me. I see no misspelled GMGen entries.

PR #3343, opened 2017-08-09
DATA-3327

evilpixie87:
Fixed Gunslinger archetypes Pistolero and Musket Master


mjmeans:
Does this also resolve that the weapon proficiency for pistolero archetype =only grants proficiency with one-handed firearms instead of all firearms?

mjmeans:
"AUTO:WEAPONPROF|TYPE=Firearm" is probably wrong because it grants proficiency in all firearms. It should be one-handed firearms only for pistolero. But that might require all firearm weapon equipment items to have a new type added to them to specify which are one-handed an which are two-handed.

evilpixie87:
Actually, BOTh archetypes had the correct firearm proficiency offering. However, the gunslinger proficiency "kill switch" wasn't implemented properly AND neither proficiency "killed" weapon proficiencies OR altered gunsmithing.

PR #3347, opened 2017-08-10
Path of Shadow fixes

evilpixie87:
Fixed Nightblade proficiencies and several errors in Stygian Stalker archetype


LegacyKing:
Issues

PR #3354, opened 2017-08-13
Fix copyRoot

javydreamercsw:
Actually copy files to Root


LegacyKing:
Without a workaround, the windows builds fail. We'll have to incorporate a proper set up for this.

javydreamercsw:
I'll work on another way to fix this. Closing this PR.

PR #3356, opened 2017-08-13
Remove conditonal in release.gradle

grimreaper:
This entire section should be changed to rely on native
Gradle caching. In the meantime remove a conditional inside of an identical one.


LegacyKing:
Due to change in Gradle closing this now. If this needs to be addressed, please open new PR.

PR #3357, opened 2017-08-13
Make use of system images

grimreaper:


code/src/java/plugin/notes/gui/JIcon.java
```diff
@@ -54,9 +53,9 @@
...
 
 	/**
-	 *  Creates new form JIcon
+	 * Creates new form JIcon
 	 *
-	 *@param  name  Name of the file to load (full path)
+	 * @param file file object for the JIcon
 	 * @param plugin
 	 */
 	JIcon(File name, NotesPlugin plugin)
```
javydreamercsw: The parameter needs to change from name to file for the code to work. This doesn't compile.

grimreaper: I think this was a mis-merge considering that the javadoc was changed. Thanks

javydreamercsw:
Seems good to go. Ship it!

PR #3361, opened 2017-08-17
Implement ADMIN-52

javydreamercsw:
After building the project, you'll have all the jars and exe in root/pcgen folder for you to run there.

Distributions seems fine to me just need an expert eye to take a look.


LegacyKing:
Major failing. Looks like the imports need to be fixed to reflect directory structure changes.

javydreamercsw:
I'm unable to reproduce. All those imports are correct and compile on my local copy. The PR process might be not doing it right. Hard to tell without access to the checked out repository. That I like about Jenkins.

javydreamercsw:
There's no change needed to the imports, all is handled in the gradle files pointing to new source root.

javydreamercsw:
Bump...

LegacyKing:
This is giving an unknown repo and conflicts galore. :(   Without the input of the other major players, I'm not comfortable making a broad change in absence of the team input.

LegacyKing:
I think it might be safer and easier to just EXPORT into another folder completely, instead of changing the existing structure. I would feel fine approving that without the team input since that is simply the BUILD.

javydreamercsw:
Please update the issue with your proposed changes so I can take a look.

javydreamercsw:
I closed since this approach is no longer desired.

PR #3390, opened 2017-09-11
Fix issue CODE-3187

javydreamercsw:
1) Add Windows 32 JRE to installer.
2) Installer selects right JRE version during installation.


javydreamercsw:
This is way cleaner...

PR #3398, opened 2017-09-14
Fix bug: DATA-3350 Winter Witch archetype includes text from Bonded W…

LegacyKing:
…itch archetype


warstoke:
This bug is still in place Comment added to https://pcgenorg.atlassian.net/browse/DATA-3350

PR #3480, opened 2017-10-23
Ap8 fix

Llisandur:
Restore missing content from 3.5e version so it loads without errors.


LegacyKing:
I'm seeing some things added that have no bearing to the source. 

1) The Skill Focus just needs to change to BONUS:ABILITYPOOL|Skill Focus|2, and that's it. No need to add a feat and mod skill focus for a type.

2) Abilities that have no connection to the source, should not be added.

3) Personal note, need a better mechanic for Natural Attacks.

4) KEY is missing from all the special abilities.

Llisandur:
Trying to convert from the 3.5e source was doing some weird things, so I just implemented the proper Pathfinder version from Bestiary 4. The whole adventure needs to be scrapped and re-added from the complete Pathfinder version source.

LegacyKing:
Is there a Pathfinder version of this source available?

Llisandur:
paizo.com/products/btpy9nme
Similar to the Rise of the Runelords Anniversary Edition, Paizo has released an all-in-one adventure path book for Curse of the Crimson Throne.

LegacyKing:
We should probably implement that, but for now. I'll accept.

PR #3498, opened 2017-11-02
CODE-3218: Fixed feet tattoos being in their own slot

gonrada:
Not really a code bug, but this should correct it.

### DIFF
--- a/system/gameModes/Pathfinder/equipmentslots.lst
+++ b/system/gameModes/Pathfinder/equipmentslots.lst
@@ -4,7 +4,7 @@
 #
 # Lists the DEFAULT number of slots for each type
 #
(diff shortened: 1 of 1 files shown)


system/gameModes/Pathfinder/equipmentslots.lst
```diff
@@ -4,7 +4,7 @@
 #
 # Lists the DEFAULT number of slots for each type
 #
-NUMSLOTS:DEFAULT	HEAD:1	HANDS:2	TORSO:1	LEGS:2	SHIELD:1	TATTOOFEET:2
+NUMSLOTS:DEFAULT	HEAD:1	HANDS:2	TORSO:1	LEGS:3	SHIELD:1
```
LegacyKing: Uh, why is Legs 3 now?

LegacyKing:
This has nothing to do with Java, it's all LST, which means it should be classified as a Data bug.

gonrada:
I agree. Should the ticket be closed and opened as data bug?

PR #3511, opened 2017-11-18
DATA-3464: Implementing VMC Fighter

draconswan:
Implemented Change in Feat progression for all VMC implementations
Implemented Fighter VMC


LegacyKing:
Please make it a habit to do PRs as branches. Merge commits should never be part of a PR, practice proper rebasing.

PR #3512, opened 2017-11-20
Burdenless_Armor_fix

evilpixie87:
Minor fix to make burdenless armor increase carrying capacity.


LegacyKing:
Has this been tested? That tag is designated as Load.Lst file only.

PR #3516, opened 2017-11-22
Fix for the Fantasy Character Sheet preview for 3.5e

AVellucci:
Adjusted the html file to correct an issue with the Armor & Shield fields when having only one of them equipped.
https://i.imgur.com/Id2M7p9.png
The change made will fix the issue seen in the linked image.
The current code makes the armor and shield properties hard to read due to a repeating image that should be a solid background instead.
This "fix" corrects the repeated bad image and adjusts the code to show the proper background image.


karianna:
This is only failing because of a core dependency build issue (I've reported on the mailing list).  Once that's resolved and this PR passes appveyor then I think it's good to merge.

PR #3534, opened 2017-12-05
Update deps

karianna:


karianna:
So I *think* whomever reviews and merges this should use the "Rebase and merge" option.

karianna:
Hmm apparently master to master is not allowed - I'll branch and do this again

PR #3538, opened 2017-12-06
corrected item name to Maniacal Devices

net-diver:
Ring of Mechanical Devices doesn't exist and is actually a Ring of Maniacal Devices


LegacyKing:
Can someone verify this? Thanks! (@karianna @Nylanfs )

Llisandur:
image
Snapshot from APG 2nd edition pdf.

Llisandur:
Needs an entry in migration.lst as well.

PR #3546, opened 2017-12-10
Bloodline revamp

LegacyKing:


LegacyKing:
Okay, proof of concept draft. Crossblooded should be granting a double bonus on the feats (That's expected and will be fixed next). Right now, I just want this reviewed from a "will it work properly" point of view.

PR #3601, opened 2018-01-03
fixing return values so eclipse with java 1.8 is happy

mertonmonk:


mertonmonk:
This should make eclipse happy with 1.8 java

PR #3608, opened 2018-01-08
NEWTAG-597 - turning off VariableProcessor caching

mertonmonk:
VariableProcessor caching produces unreliable results because sometimes it will try to cache a variable used in a PREREQ that hasn't had all relevant BONUS values loaded.  It's better to rely on the caching in BonusManager.

### DIFF
--- a/code/src/java/pcgen/core/VariableProcessor.java
+++ b/code/src/java/pcgen/core/VariableProcessor.java
@@ -99,7 +99,12 @@ public abstract class VariableProcessor
 		CachableResult(Float result, boolean cachable)
 		{
 			this.result = result;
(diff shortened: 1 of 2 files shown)


thpr:
Why is this referring to NEWTAG-597?

mertonmonk:
It's related to the work to get the Function-GetFact plugin to work. In testing data that should have worked once the getfact plugin was working we discovered some other issues, so I just fixed them under the same JIRA instead of creating a new one.
-Bryan
    On Monday, January 8, 2018, 11:06:17 AM CST, thpr <notifications@github.com> wrote:  
 
 Why is this referring to NEWTAG-597?

PR #3610, opened 2018-01-09
added missing classes for spell

net-diver:
http://paizo.com/pathfinderRPG/prd/ultimateMagic/spells/interrogation.html#interrogation


LegacyKing:
Uh, why is there 124 lines changed for a 2 line addition? Makes it extremely difficult to review.

net-diver:
When I went to redo from scratch I got this alert and figure it is the cause of the major change despite only a few words actually added. Given the small edit I just copy/pasted it out of notepad which was set for ANSI encoding... note sure why I would cause issues when it didn't in the past.

**"We've detected the file encoding as windows-1252. When you commit changes we will transcode it to UTF-8. "**

LegacyKing:
Not sure I want to introduce I potential change issue... Can we get your git to respect the settings?

PR #3613, opened 2018-01-10
Wizard Powers Varaibles

Llisandur:
Moved school level bonuses from class definitions to school abilities.
Added variables to school powers that didn't have them.
Changed school powers to use separate power level and progression variables.


LegacyKing:
@evilpixie87 Can you do a sweep after this is merged to replace the Archetype "TYPEs" with proper PREVAR Booleans? Thanks!

PR #3642, opened 2018-01-14
CODE-3225 Fix MODIFYOTHER so that it correctly filters for object type

thpr:
Note:
This structure of having inline classes like this is a bit of a hack to demonstrate the process works, but we still need it to work.  Longer term, there are ways to make this more sensible in how groupings are formed


thpr:
Yes, apparently Objects.equals is now the same behavior.  So we should pull that from the base library and then handle the side effects.

thpr:
Fixed

PR #3645, opened 2018-01-15
Updates for Oxygen

thpr:
This is what I need to keep Gradle as part of the project (and not constantly trying to change the name of the project) with EGit.  Open to reducing the change if there are ways around some of this


karianna:
As discussed on HipChat this fails for me, but that's more likely because my Gradle Plugin for Oxygen is broken (have reported a bug).  So if it works for you I say we merge.

PR #3654, opened 2018-01-15
gui3: remove

grimreaper:
gui3 does not compile under Java 9 and is part of an aborted experiment
to try JavaFX. We can fix and restore if its needed in the future.


karianna:
We can make it compile under Java 9 easily enough, but agree, let's kill it.

thpr:
Approved if this doesn't blow up the build (travis passes = approved)

PR #3657, opened 2018-01-17
renamed items to avoid duplicate listings

net-diver:
renamed items to be consistent with other sources and avoid redundant entries of the same item but with slightly different names


data/pathfinder/paizo/player_companion/adventurers_armory/aa_equip.lst
```diff
@@ -106,10 +106,10 @@ Bladeguard																					TYPE:Goods.Alchemical.Consumable						COST:40	WT:
...
 Distilled Terrap Sap																			TYPE:Goods.Alchemical.Consumable						COST:30	WT:0.5																	SOURCEPAGE:p.10								SPROP:Provides +5 alchemical bonus to saves against scent attacks within 20' for 1d6 rounds
-Firework (Desnan Candle)		NAMEISPI:YES	OUTPUTNAME:Firework, Desnan Candle								TYPE:Goods.Alchemical.Consumable						COST:5	WT:0																		SOURCEPAGE:p.10
-Firework (Paper Candle)						OUTPUTNAME:Firework, Paper Candle								TYPE:Goods.Alchemical.Consumable						COST:1	WT:0																		SOURCEPAGE:p.10
-Firework (Skyrocket)						OUTPUTNAME:Firework, Skyrocket								TYPE:Goods.Alchemical.Consumable						COST:50	WT:1																		SOURCEPAGE:p.10
-Firework (Starfountain)						OUTPUTNAME:Firework, Starfountain								TYPE:Goods.Alchemical.Consumable						COST:500	WT:100																	SOURCEPAGE:p.10
+Desnan Candle (Firework)		NAMEISPI:YES	OUTPUTNAME:Firework, Desnan Candle								TYPE:Goods.Alchemical.Consumable						COST:5	WT:0																		SOURCEPAGE:p.10
```
LegacyKing: If you change the key value, then you should add SORTKEY to keep similar items together. i.e. all the fireworks are found in D, P, and S.

PR #3659, opened 2018-01-18
Updates for Base Library v 124 and Formula Library v 142

thpr:
Most of the changes relate to the formula system now using FormatManager rather than Class to identify and control this.  This was a change that provided a significant enhancement to the formula system's ability to infer information and "get" objects without having to have a syntax that is ridiculously verbose.

Also fix Context so they are not exposing this or suffering from
initialization issues during construction


code/src/java/pcgen/rules/context/AbstractReferenceContext.java
```diff
@@ -68,7 +69,26 @@
 import pcgen.util.Logging;
 import pcgen.util.StringPClassUtil;
 
-public abstract class AbstractReferenceContext implements ObjectDatabase
+/**
```
karianna: You realise you've re-created invokedynamic here ;-).

thpr: Umm, no, not really (not to mention most of this code predated Java 7 ;) ).  

Can we get a benefit out of that observation?  Does it make sense to have some portion of our code come out of Java and into something that can actually produce the invokedynamic instruction?  Would we reduce code by 4X or more doing that? (or be able to get some better function - since even as complex as AbstractReferenceContext is, there are still things we want that it doesn't do?

karianna: Yeah - we could use MethodReferences and MethodHandles that get lazily resolved.  Again a pretty major global change though.

PR #3660, opened 2018-01-18
Reduce use of Globals.getContext()

thpr:
Background: Globals.getContext() is generally bad behavior, as it's a loader function, and really should be avoided at runtime.  While we have many uses of it, we want to reduce those over time, especially those in code we know to be strategic.  This is another pass at reducing the calls to Globals.getContext() and replacing it with alternative methods of accessing the context or go get the required information directly from another object.

### DIFF
--- a/code/src/java/pcgen/cdom/helper/CNAbilitySelection.java
+++ b/code/src/java/pcgen/cdom/helper/CNAbilitySelection.java
@@ -30,8 +30,8 @@ import pcgen.cdom.enumeration.Nature;
 import pcgen.cdom.enumeration.ObjectKey;
 import pcgen.core.Ability;
 import pcgen.core.AbilityCategory;
-import pcgen.core.Globals;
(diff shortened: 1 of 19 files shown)


LegacyKing:
LGTM

code/src/java/pcgen/cdom/helper/CNAbilitySelection.java
```diff
@@ -103,7 +103,7 @@ public String getPersistentFormat()
 	}
 
 	public static CNAbilitySelection getAbilitySelectionFromPersistentFormat(
-		String persistentFormat)
+		LoadContext context, String persistentFormat)
```
karianna: Have we thought about using a DI framework like Guice / Dagger etc?

LegacyKing:
@karianna , we already use Spring. Are you suggesting we switch or incorporate another framework?

karianna:
Nope Spring would be fine

PR #3664, opened 2018-01-18
renamed items to avoid duplicates

net-diver:
renamed items to avoid duplicate listings between different sources and ensure consistency


LegacyKing:
Duplicate and conflicted.

PR #3667, opened 2018-01-19
Concede power to gradle on .classpath and .project

thpr:
(and ignore /bin since we now let eclipse build there)


LegacyKing:
This won't break any of the IDE's or build processes? Otherwise looks fine to me. 

We should probably clean up some other archaic hold outs scattered in the code folder while we're at it.

PR #3669, opened 2018-01-20
renamed items to be consistent

net-diver:
renamed items to be consistent with other sources


Llisandur:
These should have a KEY tag without the period (.) character.

PR #3672, opened 2018-01-20
NEWTAG-615 Reduce Table Format to only the lookup column

thpr:
This reduces the TABLE[...] format syntax to only have one sub-argument.  Previously, it had both the lookup and result columns.  This now seems unnecessary.  The result column format is already provided in the COLUMN[...] format of the column used in lookup(...), so this is redundant information that is, at best, error catching.  

At worst, it is much more confusing since a table with 3 columns, NUMBER, STRING, and NUMBER, must be referred to as two different formats if the target column type is included in the TABLE format.  That confusion is unnecessary, as the information is available elsewhere, and there is no added value from providing the redundant format information.


LegacyKing:
@thpr conflicted.

thpr:
Replaced by PR#3682

PR #3681, opened 2018-01-21
Added Third Party source for alcohol feats

kypvalanx:
added TPK games Feats of Legend 20 alcohol feats that i've been using for a campaign.


LegacyKing:
Hi @kypvalanx thanks for the new source submission, these actually have a different process. (They are vetted by our Publisher Liaison Monkey to verify that we have a) Permission to use the source, b) What permission level said publisher is allowing, c) is the source in compliance with all Publisher Intellectual Property). We do this on our New Source Repository (https://github.com/PCGen/pcgen-newsources).  Are you able to submit this over to that repository? Also, we have a Jira system to track these vetting steps over here > https://pcgenorg.atlassian.net/projects/NEWSOURCE/issues/NEWSOURCE-332?filter=allopenissues 

@Nylanfs FYI - This is lacking an OGL, and I haven't looked to see if TPK Games has publisher permissions yet.

kypvalanx:
I can do that.  the book says it's published under OGL so it should be fine... i think.

PR #3712, opened 2018-01-27
Equipment Overhaul - CR

Llisandur:
Proposed changes for new equipment style. Currently only includes equipment and proficiencies from Core Rulebook.

Parenthesis removed from as many equipment as possible to conform to official Paizo source statblocks or item descriptions. Removed all slashes from item names.
Proficiency names fixed to remove parenthesis.
Where it makes sense, used UI name (left-most text) and KEY rather than OUTPUTNAME.
Scrolls and potions fixed to use EQMOD similar to wands.
CR magic items changed to copy of base items rather than individual items.
DESC and QUALITY tags added for missing item information.
ABILITY tags added for some items to add information to CombatBonus, SaveBonus, and SkillBonus areas of character sheets. SPROP tags removed for same information.


evilpixie87:
Won't all these equipment name changes cause folks to need to rebuild current inventories when they load their characters?

LegacyKing:
Conflicted in light of other PRs being incorporated. Please resubmit after a good squash and rebase (clean up).

LegacyKing:
Also, the title is misleading, you have other sources not CR, which is why this PR became conflicted.

PR #3720, opened 2018-01-28
[Do Not Merge] CR - Equipment and Proficiency Overhaul

Llisandur:
Proposed changes for new equipment style. Currently only includes equipment and proficiencies from Core Rulebook.
Parenthesis removed from as many equipment as possible to conform to official Paizo source statblocks or item descriptions. Removed all slashes from item names.
Proficiency names fixed to remove parenthesis.
Where it makes sense, used UI name (left-most text) and KEY rather than OUTPUTNAME.
Scrolls and potions fixed to use EQMOD similar to wands.
CR magic items changed to copy of base items rather than individual items.
DESC and QUALITY tags added for missing item information.
ABILITY tags added for some items to add information to CombatBonus, SaveBonus, and SkillBonus areas of character sheets. SPROP tags removed for same information.
Equipment names changed in migration.lst.
Old equipmods changed from line copy to .COPY in case of non-KEY changes.


Llisandur:
Copied from HipChat

> [8:25 PM] Regan Anderson: So what's the deal with the artisan tools that messes up if you combine them into a single item?
[8:27 PM] Gwen T: Each craft skill has a PREITEM that requires the exact tool to prevent the -2 penalty. The masterwork versions increase that specific skill. Those bonuses wouyld be lost and the penalty would become universal
[8:28 PM] Gwen T: TEMPBONUS:PC|SKILL|Craft (Carpentry)|-2|TYPE=Circumstance|!PREITEM:1,TYPE=CraftToolsCarpentry
[8:29 PM] Regan Anderson: The eqmod that changes the base item to the specific tool adds the type required to prevent the penalty. Also, the penalty is a tempbonus so you can just not apply it.
[8:30 PM] Gwen T: Does the EQMOD add the Masterwork +2??
[8:30 PM] Regan Anderson: Yes, if you add masterwork to it. I already tested all of it.
[8:32 PM] Gwen T: OK. I didn't see that part. If that's all sorted, then no problem. So when you buy the tools, you choose what kind they are??
[8:33 PM] Regan Anderson: Yes. All of the individual types are moved to eqmods. The base tool has MOS:REQUIRED, and the only choices are the tool type and masterwork. It adds to price correctly, too.
[8:33 PM] Gwen T: OK.
[8:33 PM] Gwen T: Still not loving the extended DESC's, though.
[8:34 PM] Regan Anderson: I made instruments the same, but I might revert that before I put in the latest version of the thing.
[8:35 PM] Regan Anderson: DESCs are on the DESC tag, so they won't appear on the character sheet unless you add that it.
[8:35 PM] Gwen T: If they work, I have no problem with it. There are just sooo many moving parts to this I can't really get a grip on all of it.
[8:35 PM] Regan Anderson: Yeah. I may cut it up like you said earlier.
[8:36 PM] Gwen T: What do you mean about the DESCs not appearing?? They show up on MY character sheets....
[8:36 PM] Regan Anderson: Where are they showing?
[8:37 PM] Gwen T: Immediately under the item name.
[8:37 PM] Regan Anderson: Ah, you mean in the inventory management tab?
[8:37 PM] Gwen T: No....
[8:38 PM] Regan Anderson: Which character sheet are you using?
[8:38 PM] Gwen T: Hold on... I'm pulling one up
[8:40 PM] Gwen T: 
    File uploaded: upload.png
    [https://s3.amazonaws.com/uploads.hipchat.com/403927/3808143/Eecdo0vk9AQm7Ju/upload.png]
[8:41 PM] Regan Anderson: On the left?
[8:41 PM] Gwen T: On the right.
[8:42 PM] Gwen T: For example, the robe of infinite twine. THAT'S the DESC
[8:42 PM] Regan Anderson: That stuff has always been there. It annoys me too, which is part of what I'm trying to cut down. I didn't mess with that item at all. That's in the SPROP tag.
[8:43 PM] Gwen T: What?? The SPROP appears on the sheet, but the DESC is on screen only?
[8:43 PM] Regan Anderson: It *should* be in the DESC tag, but no one has been using the DESC tag.
[8:43 PM] Regan Anderson: Yes.
[8:43 PM] Regan Anderson: Unless you call the DESC tag on the character sheet, which none of them do.
[8:45 PM] Gwen T: Oh! Then I've got no complaints. I'd actually LIKE to see the write up while purchasing as long as it's not on the sheet. All I like on the sheet is stuff you're likely to need to reference during combat.
[8:45 PM] Regan Anderson: Same here.

PR #3729, opened 2018-01-30
Inspired_panache_fix

evilpixie87:
Added defines and aspect to Inspired Blade Panache


LegacyKing:
Why'd we drop the display ability?

evilpixie87:
I was helping someone in HipChat before work and this was just the way I found to make things work after 2 or 3 failures. Was rushing a bit because of the hour. I can replace it if you'd prefer.

LegacyKing:
It looks like it had a purpose? Don't want to remove it, if it did...

evilpixie87:
OK, I put that back!

PR #3735, opened 2018-01-31
NEWTAG-617 First pass of PREREQUISITE

thpr:
Includes framework (token at this point) for REQUIREMENT, but the REQUIREMENT token is guaranteed to fail at LST load.

This is the first in a series of items that provides for interaction of the new formula system to data control of allowed/granted.  This implements (a beta version as it is subject to negotiation with the data team) of PREREQUISITE:

This should enable certain items (e.g. VISION) to be fully converted to Dynamic/Formula items.

Note there are certain deficiencies with this system as it is currently designed.  The following features still work with PRExxx, but this new token will not:
- Build Ability Prerequisites for the Prerequisite Tree view for the UI
- Display in the UI of what the PREREQUISITE situation is (in the HTML description of an object)

In addition, there is some existing items where I think it is debate-able as to how they should actually be implemented, which is why the "new" system doesn't assume it should solve them automagically.  The following features still work with PRExxx, but this new token will not:
- Enforce any Race / Alignment common restrictions for UI display
- Perform Equipment auto-resizing based on PRESIZE
- Leverage EqMod EQMODTYPE=MagicalEnhancement to automatically trigger COSTDOUBLE
- Automatically impact NATURALATTACKS SIZE on an object with PRESIZE:

### DIFF
--- a/code/pluginbuild.xml
+++ b/code/pluginbuild.xml
@@ -2323,10 +2323,24 @@ jar-all-plugins - Generate the plugin jar files
 				</patternset>
 			</fileset>
 		</jar>
+		<jar jarfile="${lstplugins.dir}/LstToken-PREREQUISITE.jar" manifest="${src.java.dir}/plugin/lsttokens/manifest.mf">
+			<fileset dir="${build.classes.dir}">
+				<patternset>
+					<include name="plugin/lsttokens/PrerequisiteLst.class" />
+				</patternset>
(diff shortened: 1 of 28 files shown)


LegacyKing:
Can we possibly shorten this tag to PRE and the other proposed to REQ?

Can we discuss this Saturday on Hipchat? Thanks!

PR #3736, opened 2018-02-01
Proficiencies Cleanup

Llisandur:
I found a non-destructive way to remove parenthesis and such from proficiencies.


Llisandur:
Failed tests due to character output tests not being able to handle changed proficiency names. Characters load normally in actual environment.

LegacyKing:
That isn't a failed test, that's an expected outcome. When that is the case, we have to run the unit test(s) and update the values. Not hard to do.

Llisandur:
Extra commits because of rebase. I'll resubmit.

PR #3737, opened 2018-02-02
Change Old EQMOD KEYs From Line Copy to .COPY

Llisandur:
Just a change from copying the full line to .COPY, in case anything within the new keys gets updated. This way the update only needs to be applied once.


Llisandur:
Failed tests because new KEY EQMODs have different SPROP.

LegacyKing:
We'll need to update the tests to reflect the changes

PR #3739, opened 2018-02-02
Change name from PCGenModifier to FormulaModifier

thpr:
Also some document cleanups and minor bug fixes

This performs a global rename of PCGenModifier to FormulaModifier in order to improve semantics.  This is also a preparatory PR for a much larger change that I wanted to avoid this "rename pollution" in a much more functional change.

### DIFF
--- a/code/src/java/pcgen/rules/context/TrackingManufacturer.java
+++ b/code/src/java/pcgen/rules/context/TrackingManufacturer.java
@@ -286,6 +286,6 @@ class TrackingManufacturer<T extends Loadable> implements ReferenceManufacturer<
 	@Override
 	public boolean isDirect()
 	{
-		return false;
+		return rm.isDirect();
 	}
 }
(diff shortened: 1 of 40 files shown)


code/src/java/pcgen/rules/context/TrackingManufacturer.java
```diff
@@ -286,6 +286,6 @@ public String unconvert(T arg0)
 	@Override
 	public boolean isDirect()
 	{
-		return false;
+		return rm.isDirect();
```
LegacyKing: This sticks out as different. I do not see 'rm' defined in the changes. Is this intentional?

mertonmonk: rm is set in the initializer method as a ReferenceManufacturer that is passed in.  This looks fine.

PR #3740, opened 2018-02-03
NEWTAG-617 First ALLOW Implementation

thpr:
Includes framework (token at this point) for ENABLE, but the ENABLE token is guaranteed to fail at LST load.

This is the first in a series of items that provides for interaction of the new formula system to data control of allowed/granted. This implements (a beta version as it is subject to negotiation with the data team) of ALLOW:

This should enable certain items (e.g. VISION) to be fully converted to Dynamic/Formula items.

Note there are certain deficiencies with this system as it is currently designed. The following features still work with PRExxx, but this new token will not:

    Build Ability Prerequisites for the Prerequisite Tree view for the UI
    Display in the UI of what the ALLOWsituation is (in the HTML description of an object)

In addition, there is some existing items where I think it is debate-able as to how they should actually be implemented, which is why the "new" system doesn't assume it should solve them automagically. The following features still work with PRExxx, but this new token will not:

    Enforce any Race / Alignment common restrictions for UI display
    Perform Equipment auto-resizing based on PRESIZE
    Leverage EqMod EQMODTYPE=MagicalEnhancement to automatically trigger COSTDOUBLE
    Automatically impact NATURALATTACKS SIZE on an object with PRESIZE:

### DIFF
--- a/code/src/java/plugin/lsttokens/EnableLst.java
+++ b/code/src/java/plugin/lsttokens/EnableLst.java
@@ -1,6 +1,5 @@
 package plugin.lsttokens;
 
-import pcgen.base.formatmanager.FormatUtilities;
 import pcgen.base.formula.inst.NEPFormula;
 import pcgen.base.lang.StringUtil;
 import pcgen.cdom.base.CDOMObject;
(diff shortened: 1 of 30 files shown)


thpr:
Please do not merge, per Andrew's questions on the predecessor PR

LegacyKing:
LGTM

thpr:
This needs a code review, and Codacy can be ignored (the Prerequisite code is approaching deprecation, and the package visibility is intentional.

code/src/java/plugin/lsttokens/EnableLst.java
```diff
@@ -0,0 +1,57 @@
...
+	{
+		return CDOMObject.class;
+	}
+
+	@Override
+	protected ParseResult parseNonEmptyToken(LoadContext context, CDOMObject obj,
+		String value)
+	{
+		return new ParseResult.Fail("Not supported since it is not monitored in an ongoing fashion");
+//		NEPFormula<Boolean> formula =
```
karianna: commented out code?

PR #3743, opened 2018-02-04
Remove (some) useless @see references.

thpr:
Note that between @Override and features of modern IDEs, these are all
redundant, and actually more likely to become broken links than provide
any value


LegacyKing:
LGTM

PR #3744, opened 2018-02-04
Change minor method semantics on Reference Manufacturer

thpr:
Effectively this introduces a new (slightly more powerful) FormatManager
derivative called an ObjectManager, and then uses that for
ReferenceManufacturer objects.  This additional capability will be
expanded and leveraged for a number of items in the formula system going
forward.

Also removes one (confusing and likely buggy) unused method

### DIFF
--- a/code/src/java/pcgen/cdom/base/ObjectManager.java
+++ b/code/src/java/pcgen/cdom/base/ObjectManager.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018 (C) Thomas Parker <thpr@users.sourceforge.net>
+ * 
+ * This library is free software; you can redistribute it and/or modify it under the terms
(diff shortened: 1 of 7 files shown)


code/src/java/pcgen/cdom/base/ObjectManager.java
```diff
@@ -0,0 +1,41 @@
...
+import pcgen.base.util.FormatManager;
+
+/**
+ * An ObjectManager is an extension of a FormatManager used in PCGen to serve as a
+ * "Groupable" FormatManager.
+ * 
+ * @param <T>
+ *            The format of object for which this ObjectManager provides services
+ */
+public interface ObjectManager<T extends Loadable> extends FormatManager<T>
```
karianna: Could have an improved name?  `ObjectManager` is very generic....

PR #3746, opened 2018-02-04
Data 3533 make wildblooded bloodlines selectable

allardhoeve:
This is not the correct fix for bloodlines, but the system is more complex than I know how to fix. One of the problems is that since 796429d83ad6c588cf8a0ec310a28be75a6da56d all bloodlines need fixing. If there is documentation on how to do this, I'd be happy to help.

This works for all bloodlines, except Linnorm (needs some solution to pop up a second chooser) and Primal (needs some solution to choose one of four elements). These will (looking at the Draconic) bloodline, probably be more Ability Pools?

### DIFF
--- a/data/pathfinder/paizo/roleplaying_game/ultimate_magic/um_abilities_class.lst
+++ b/data/pathfinder/paizo/roleplaying_game/ultimate_magic/um_abilities_class.lst
@@ -1589,6 +1589,33 @@ Swarm Trap		KEY:Ranger Trap ~ Swarm Trap		SORTKEY:Trap Swarm	CATEGORY:Special Ab
 
 ###Block: Sorcerer archetypes
 
+###Block: new style selector abilities for bloodllines
+# COMMENT: These abilities are used in the popup when a bloodline is selected, they automatically grant
+#          the bloodline abilities themselves. Core rulebook has all sorts of variables assigned that
+#          are used in the bloodline powers themselves. These are not ported over because these
+#          bloodlines seem to use an older style of definition that does not use the new variable system.
+
+Anarchic Bloodline	CATEGORY:Sorcerer Bloodline	TYPE:SorcererBloodlineChoice	ABILITY:Special Ability|AUTOMATIC|Sorcerer Bloodline ~ Anarchic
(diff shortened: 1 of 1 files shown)


allardhoeve:
The awkward master merge is because there is a file called `gradlew.bat` with weird line endings in the way of merging. I cannot get a clean checkout anymore :(

LegacyKing:
    org.junit.ComparisonFailure: Errors encountered while loading Pathfinder RPG Core Supplements expected:<[]> but was:<[Invalid use of MULT:YES Ability Linnorm (Wildblooded Draconic) Bloodline where a target [parens] is required,
    PLEASE TAKE NOTE: If usage locations are reported, not all usages are necessary illegal (at least one is),
    Invalid use of MULT:YES Ability Primal (Wildblooded Elemental) Bloodline where a target [parens] is required,
    PLEASE TAKE NOTE: If usage locations are reported, not all usages are necessary illegal (at least one is)]>
        at pcgen.persistence.lst.DataLoadTest.testLoadSources(DataLoadTest.java:160)

allardhoeve:
Hehe no, I was afraid of that. Is that new way documented somewhere? @Llisandur

Llisandur:
Not yet. I'm working out the documentation/templates at the moment.

allardhoeve:
Let's leave this quick fix open in case someone else runs into the problem, then delete the PR when the issue is resolved?

Llisandur:
Here's what I have so far for a template. I still need to see how other bloodline-affecting archetypes work so I can add them in.


[Sorcerer Bloodline Template.txt](https://github.com/PCGen/pcgen/files/1702885/Sorcerer.Bloodline.Template.txt)

allardhoeve:
Wow, that is an impressive template. Is this all needed for every bloodline?

If you don't mind my asking, what is the rationale for the overhead?





Op wo 7 feb. 2018 12:50 schreef Regan Anderson <notifications@github.com>:

Llisandur:
Basically, to support all of the features of the bloodline and all the ways to gain a bloodline/bloodline power. The overhead is basically a how-to on using the template; you can also just do replaces on the overhead and it'll fill out the template with all of the basics. Things in the TYPE tag of the power abilities are the only things that you have to scroll down and edit individually. Of course, you'll have to scroll down to the powers anyway to add in the individual variables and bonuses and things.

PR #3752, opened 2018-02-05
Pushes ParseResult deeper into the processing for certain tokens.

thpr:
For items that are trying to shortcut a good report on the location of
an error, this will help reduce that problem.  It forces a slightly more
verbose syntax in a few places, but in general helps standardize
behavior and improve error reporting.

### DIFF
--- a/code/src/java/pcgen/rules/persistence/token/AbstractSimpleChooseToken.java
+++ b/code/src/java/pcgen/rules/persistence/token/AbstractSimpleChooseToken.java
@@ -102,9 +102,10 @@ public abstract class AbstractSimpleChooseToken<T extends Loadable> extends
 		}
 		else
 		{
-			if (hasIllegalSeparator('|', activeValue))
+			ParseResult pr = checkForIllegalSeparator('|', activeValue);
+			if (!pr.passed())
 			{
-				return ParseResult.INTERNAL_ERROR;
+				return pr;
 			}
 			Set<PrimitiveCollection<T>> set =
                     new HashSet<>();
--- a/code/src/java/plugin/converter/FavClassConvertPlugin.java
+++ b/code/src/java/plugin/converter/FavClassConvertPlugin.java
@@ -36,6 +36,7 @@ import pcgen.gui2.converter.event.TokenProcessEvent;
 import pcgen.gui2.converter.event.TokenProcessorPlugin;
 import pcgen.rules.context.LoadContext;
 import pcgen.rules.persistence.token.AbstractToken;
+import pcgen.rules.persistence.token.ParseResult;
 
 public class FavClassConvertPlugin extends AbstractToken implements
 		TokenProcessorPlugin
@@ -54,9 +55,15 @@ public class FavClassConvertPlugin extends AbstractToken implements
 		}
 
 		String choices = value.substring(7);
-		if (isEmpty(choices) || hasIllegalSeparator('|', choices))
+		ParseResult pr = checkNonEmpty(choices);
+		if (!pr.passed())
 		{
-			return "Empty Token";
+			return "Empty";
+		}
+		pr = checkForIllegalSeparator('|', choices);
+		if (!pr.passed())
+		{
+			return "Illegal Separator";
 		}
 		boolean foundAny = false;
 		boolean foundOther = false;
--- a/code/src/java/plugin/lsttokens/NaturalattacksLst.java
+++ b/code/src/java/plugin/lsttokens/NaturalattacksLst.java
@@ -63,6 +63,7 @@ public class NaturalattacksLst extends AbstractTokenWithSeparator<CDOMObject>
 {
 
 	private static final Class<WeaponProf> WEAPONPROF_CLASS = WeaponProf.class;
+	private static final int MIN_TOKEN_COUNT = 4;
 
 	/**
 	 * @see pcgen.persistence.lst.LstToken#getTokenName()
@@ -127,7 +128,7 @@ public class NaturalattacksLst extends AbstractTokenWithSeparator<CDOMObject>
 			StringTokenizer commaTok = new StringTokenizer(wpn, Constants.COMMA);
 
 			int numTokens = commaTok.countTokens();
-			if (numTokens < 4)
+			if (numTokens < MIN_TOKEN_COUNT)
 			{
 				return new ParseResult.Fail("Invalid Build of " + "Natural Weapon in "
 						+ getTokenName() + ": " + wpn, context);
@@ -142,9 +143,9 @@ public class NaturalattacksLst extends AbstractTokenWithSeparator<CDOMObject>
 			}
 
 			attackName = attackName.intern();
-			Equipment anEquip = new Equipment();
-			anEquip.setName(attackName);
-			anEquip.put(ObjectKey.PARENT, obj);
+			Equipment naturalWeapon = new Equipment();
+			naturalWeapon.setName(attackName);
+			naturalWeapon.put(ObjectKey.PARENT, obj);
 			/*
 			 * This really can't be raw equipment... It really never needs to be
 			 * referred to, but this means that duplicates are never being detected
@@ -161,7 +162,7 @@ public class NaturalattacksLst extends AbstractTokenWithSeparator<CDOMObject>
 			 * guarantee uniqueness of the key?? - that's too paranoid
 			 */
 
-			EquipmentHead equipHead = anEquip.getEquipmentHead(1);
+			EquipmentHead equipHead = naturalWeapon.getEquipmentHead(1);
 
 			String profType = commaTok.nextToken();
 			pr = checkForIllegalSeparator('.', profType);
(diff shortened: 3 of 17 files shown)


code/src/java/pcgen/rules/persistence/token/AbstractSimpleChooseToken.java
```diff
@@ -102,9 +102,10 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
 		}
 		else
 		{
-			if (hasIllegalSeparator('|', activeValue))
+			ParseResult pr = checkForIllegalSeparator('|', activeValue);
+			if (!pr.passed())
 			{
-				return ParseResult.INTERNAL_ERROR;
+				return pr;
```
karianna: Pattern seems a little unusual that you only return if it hasn't passed.  Why wouldn't it also be returned if it did pass?

code/src/java/plugin/converter/FavClassConvertPlugin.java
```diff
@@ -54,9 +55,15 @@ public String process(TokenProcessEvent tpe)
 		}
 
 		String choices = value.substring(7);
-		if (isEmpty(choices) || hasIllegalSeparator('|', choices))
+		ParseResult pr = checkNonEmpty(choices);
+		if (!pr.passed())
 		{
-			return "Empty Token";
+			return "Empty";
```
karianna: extract Magic Const

code/src/java/plugin/converter/FavClassConvertPlugin.java
```diff
@@ -54,9 +55,15 @@ public String process(TokenProcessEvent tpe)
...
+		}
+		pr = checkForIllegalSeparator('|', choices);
+		if (!pr.passed())
+		{
+			return "Illegal Separator";
```
karianna: extract Magic Const

code/src/java/plugin/lsttokens/NaturalattacksLst.java
```diff
@@ -122,186 +122,160 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
...
 				return pr;
 			}
-			Equipment anEquip = createNaturalWeapon(context, obj, tokString.intern());
+			String wpn = tokString.intern();
 
-			if (anEquip == null)
+			StringTokenizer commaTok = new StringTokenizer(wpn, Constants.COMMA);
+
+			int numTokens = commaTok.countTokens();
+			if (numTokens < 4)
```
karianna: extract Magic Number

code/src/java/plugin/lsttokens/NaturalattacksLst.java
```diff
@@ -122,186 +122,160 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
...
+			if (attackName.equalsIgnoreCase(Constants.LST_NONE))
 			{
-				anEquip.setModifiedName("Natural/Primary");
+				return new ParseResult.Fail("Attempt to Build 'None' as a "
+						+ "Natural Weapon in " + getTokenName() + ": " + wpn, context);
 			}
-			else
+
+			attackName = attackName.intern();
+			Equipment anEquip = new Equipment();
```
karianna: Is this really `anItem` ?

code/src/java/plugin/lsttokens/NaturalattacksLst.java
```diff
@@ -122,186 +122,160 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
...
-		{
-			Type type = Type.getConstant(dotTok.nextToken());
-			anEquip.addToListFor(ListKey.TYPE, type);
-		}
+			// sage_sam 02 Dec 2002 for Bug #586332
+			// allow hands to be required to equip natural weapons
+			int handsrequired = 0;
+			while (commaTok.hasMoreTokens())
+			{
+				final String hString = commaTok.nextToken();
```
karianna: is this a `handsString`?

code/src/java/plugin/lsttokens/NaturalattacksLst.java
```diff
@@ -122,186 +122,160 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
...
+				if  (hString.startsWith("SPROP="))
+				{
+					anEquip.addToListFor(ListKey.SPECIAL_PROPERTIES,
+						SpecialProperty.createFromLst(hString.substring(6)));
```
karianna: extra magic number

code/src/java/plugin/lsttokens/NaturalattacksLst.java
```diff
@@ -122,186 +122,160 @@ protected ParseResult parseTokenWithSeparator(LoadContext context,
...
-		anEquip.put(ObjectKey.ATTACKS_PROGRESS, !attacksFixed);
-		try
-		{
-			int bonusAttacks = Integer.parseInt(numAttacks) - 1;
-			final BonusObj aBonus = Bonus.newBonus(context, "WEAPON|ATTACKS|"
-					+ bonusAttacks);
+			anEquip.put(ObjectKey.WEIGHT, BigDecimal.ZERO);
 
-			if (aBonus == null)
+			WeaponProf cwp = context.getReferenceContext().silentlyGetConstructedCDOMObject(
```
karianna: what's a `cwp`? :-)

code/src/java/pcgen/rules/persistence/token/AbstractSimpleChooseToken.java
thpr: So the ParseResult in this case is simply a check of "did this have legal separators?"  If it failed, then we need to let someone know why.  If it succeeded, it only meant the separators were legal - we still need to parse the token to get the content.

I'm open to a better structure here, but I'm also operating under the principle of "Don't throw exceptions to send back a normal condition failure"... 

I can't reasonably go to a design that checkForIllegalSeparator takes in a 3rd argument which it calls to "continue", as that would produce a wild set of temporary objects that would obfuscate what is really happening. 

I don't want to inline checkForIllegalSeparator because it's very standardized code, multiple if statements, and used hundreds of times, so it really does belong in a small method by itself.

I'm open to options - I'm not a huge fan of doing this, but I haven't thought of a better structure to use in Java.  What's the appropriate design pattern for that type of scenario?

code/src/java/plugin/converter/FavClassConvertPlugin.java
thpr: This will require discussion - I think we have different definitions of "magic"

code/src/java/plugin/lsttokens/NaturalattacksLst.java
thpr: It's specifically Equipment - it HAS to be since the getEquipmentHead is called on it later.  So I think a more generic names is a disservice to the reader.  Now whether it should be "anEquipment" or "equipment" or "naturalWeapon" (the last probably being most appropriate) is up for discussion

code/src/java/plugin/lsttokens/NaturalattacksLst.java
thpr: So what's the real best practice here?  Right now, the "magic" 6 is only 2 lines away from what it's associated with - the hString.startsWith("SPROP=").  To extract that and move it away seems like a disservice to me.  I would think it's MORE likely to cause a problem.  The extreme answer is to extract "SPROP=" and have another constant derive it's length (this guarantees one master set of information that is always consistent)... is that what is really done at this point?

code/src/java/plugin/converter/FavClassConvertPlugin.java
karianna: Is it used more than once?

code/src/java/plugin/converter/FavClassConvertPlugin.java
karianna: Is it used more than once?  If not then ignore me :-)

code/src/java/plugin/lsttokens/NaturalattacksLst.java
karianna: Might be useful to know for the developer what `4` means - ILLEGAL_NUMBER_OF_TOKENS?  Something else?

code/src/java/plugin/lsttokens/NaturalattacksLst.java
karianna: Aha!  `naturalWeapon` would be perfect.

code/src/java/plugin/lsttokens/NaturalattacksLst.java
karianna: Oh sorry, my bad - I read this as an index extraction, ignore me.

code/src/java/pcgen/rules/persistence/token/AbstractSimpleChooseToken.java
karianna: Std thing would be to throw an `IllegalSeparatorsException` as this is an 'exceptional case' (we expect LST data separators to be in the correct format).  Digging into `ParseResult`... Oh I see now.  We're collecting all parse errors to return to the user so we don't bomb out on the first one.  And we're using an internal QueueMessage attached to that interface.  OK, you'd typically want a CQRS / eventbus / message bus put in place here, the queue serves that purpose (a little crudely but OK).  I totally agree with having `checkForIllegalSeparator` extracted.

TLDR - lets leave this as is as a refactor to an event bus system would be global (out of scope for this).

PR #3755, opened 2018-02-11
Implement changes from using Class<T> to ClassIdentity<T>

thpr:
This is the second in a number of steps to make PCGen operate more on the Identity of an object rather than the Class.  This is necessary/valuable since the Ability objects are really dealt with by Identity (Category and Key), not just the class.  This is also true of other types of objects, and it will be necessary to really expand how DYNAMIC objects are used.

This exposes a few quirks - we have to suppress an error because the CATEGORY on an Ability is mutable - and thus when initially loaded they are "null" as their CATEGORY.  That is a strategic cleanup item.  

Also, this is probably the most minimally invasive change to get this next set of capability - it certainly does not expand ClassIdentity to all the locations where it is valuable to use.  Those changes are reserved for later PRs.


karianna:
@thpr Conflict?

code/src/java/pcgen/cdom/reference/CDOMFactory.java
```diff
@@ -142,7 +160,24 @@ public boolean populate(ReferenceManufacturer<T> parentCrm,
 	@Override
 	public ManufacturableFactory<T> getParent()
 	{
+		/*
+		 * CONSIDER This is a limitation that prevents this from being used for
+		 * AbilityCategory. Need to figure out if this should really know it's parent and
+		 * if so, how? (when the problem is that the parentage is really now in
+		 * ClassIdentity)
```
karianna: Could inheritance via traits (interfaces with default methods) help?

code/src/java/pcgen/cdom/reference/PatternMatchingReference.java
```diff
@@ -260,4 +256,17 @@ public String getChoice()
...
+	public Class<T> getReferenceClass()
+	{
+		return all.getReferenceClass();
+	}
+
+	@Override
+	public String getReferenceDescription()
+	{
+		//TODO Is this really what I want?!?
+		return all.getReferenceDescription();
```
karianna: It's a good question :-).  I'm actually wondering about the DRY pattern being broken, e.g. the `getReferenceClass` and `getReferenceDescription` methods both pass through to an underlying object for this class and all others like it.  I wonder if it could be hauled up somehow...

code/src/java/pcgen/cdom/base/Loadable.java
```diff
@@ -19,17 +19,65 @@
...
+	 * Note that the given String may contain a TYPE= or TYPE. prefix (which will be
+	 * discarded) and can contain more than one type (separated by periods '.'). If more
+	 * than one type is provided, this Loadable must be of ALL of the given types for this
+	 * method to return true.
+	 * 
+	 * @param type
+	 *            The type to be checked to see if this Loadable is of the given Type
+	 * @return true if the object is of the given type; false otherwise.
+	 */
 	public boolean isType(String type);
```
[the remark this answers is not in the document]
thpr: Since this is not creating this method, but is simply documenting the method, I don't think a cross-cutting refactor that will hit another ~50 classes should be done as part of one PR.

code/src/java/pcgen/cdom/reference/CDOMFactory.java
thpr: Maybe.  It's specifically CONSIDER not TODO since I want to remind myself of the scenario as I go about polishing ClassIdentity, Manufacturers, etc.  Things are likely to move around a bit and I don't want to actually try to solve it now because this is trying to be as small a PR as possible, knowing more ClassIdentity work is coming.

code/src/java/pcgen/cdom/reference/PatternMatchingReference.java
thpr: This was my way of saying "This isn't done, hopefully someone catches this".. so thanks for catching it... all the items will have polished descriptions now.

As far as the underlying objects, there is a pretty major philosophical debate here that I think deserves an off-line discussion.  There are reasons the tools that catch these delegation behaviors refer to the checking rules as "controversial"... The underlying information and timing of object construction requires one or more good design patterns to be violated, the question is which one.

code/src/java/pcgen/cdom/inst/DynamicCategory.java
```diff
@@ -21,6 +21,12 @@
 import pcgen.cdom.formula.scope.GlobalScope;
 import pcgen.cdom.helper.SpringHelper;
 
+/**
+ * A DynamicCategory is a method for keeping different Formats of Dynamic objects
+ * separated.
```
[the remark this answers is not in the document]
thpr: Done

PR #3760, opened 2018-02-13
Convert ParseResult.Fail to not take in the context

thpr:
This means ParseResult.Fail will not capture the URI where the failure
occurred - rather the LoadContext (which gets back the ParseResult) will
have to pass the URI into the ParseResult reporting mechanisms.  This
corrects for many places where a context was not being passed into the
ParseResult.Fail at construction, so we may have been losing locations
where errors were occurring.


thpr:
Codacy should be ignored - it's an arbitrary limit and this is not regression on that method.

PR #3767, opened 2018-02-16
NEWTAG-619 Add SORTKEY for RollMethod objects

thpr:
This forces all RollMethod objects to have a SORTKEY.

This will be the last class that is sorted internally to PCGen but does not have SORTKEY. With this completed, all of the existing "assumes things are sorted by how they appear first" will be deprecated and removable from PCGen.

This is proposed to be a non-graceful change (meaning it will complain immediately, not provide a transition between versions) as this is a game mode level item and therefore very isolated.  It will also not be catastrophic with this PR if SORTKEY doesn't exist... (the order won't actually change in the UI based on this at this time - it will occur later after the older methods are removed a few PRs from now)

### DIFF
--- a/code/pluginbuild.xml
+++ b/code/pluginbuild.xml
@@ -6355,6 +6355,13 @@ jar-all-plugins - Generate the plugin jar files
 				</patternset>
 			</fileset>
 		</jar>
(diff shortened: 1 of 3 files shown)


thpr:
This should not be merged without explicit approval of Andrew (as data team lead and this impacts Data immediately)

Issue #3753, opened 2018-02-10
 Pathfinder: Eidolon slam damage is incorrect in output


simiczhang:
Steps:
   1. Create any character with class "summoner"
   2. Add 4 levels (to ensure you have at least 6 evolution points)
   3. Create Eidolon and choose "Quadruped" as Eidolon type
   4. Switch to Eidolon's tab
   5. Go to Feats & Abilities -> Racial Abilities -> Eidolon Evolution
   6. Add limb (arms) * 2
   7. Add slam * 2
   8. Go to "Character Sheet" 

Expected:
  Slam damage should be 1d8+x

Actual:
  Slam damage is 3d6+x
![slam damage](https://user-images.githubusercontent.com/2820936/36061316-9efd0438-0e93-11e8-8d3c-87b22660cd39.PNG)

LegacyKing:
Replace Slam with this:
`Slam								KEY:Evolution ~ Slam								CATEGORY:Special Ability	TYPE:EvolutionChoice.Extraordinary											PREABILITY:1,CATEGORY=Special Ability,Evolution ~ Arms																																																																															PREVARGTEQ:EidMaxAttacks,EidUsedAttacks+1	PREVARGTEQ:EidArms,EidArmsUsed+1														DEFINE:EidSlamCount|0																														DESC:The eidolon can deliver a devastating slam attack.																																																																																																								STACK:YES	MULT:YES	CHOOSE:NUMCHOICES=7|NOCHOICE																								ABILITY:Internal|AUTOMATIC|Slam																																																																																																																																																																																																																																																										BONUS:VAR|EidSlamCount|1	BONUS:VAR|EidUsedAttacks|1	BONUS:VAR|EidArmsUsed|1																																																																																																ABILITY:Internal|AUTOMATIC|Eidolon Damage Support
`

and add:
`Eidolon Damage Support	CATEGORY:Internal	BONUS:WEAPONPROF=Slam|DAMAGESIZE|2
`

LegacyKing:
This is now fixed for the release.

PR #3770, opened 2018-02-17
Fix bug: DATA-3550

LegacyKing:
Eidolon Slam Attack x2+ has incorrect Damage


LegacyKing:
Pathfinder: Eidolon slam damage is incorrect in output #3753

PR #3772, opened 2018-02-17
This fixes a problem with .gitignore

thpr:
Ths existing incarnation uses simple words for some very specific
folders, and these pattern matches accidentally impact portions of the
code base, meaning there are places where files can be added but they
will not be detected by git as significant... that's obviously a major
problem, so the .gitignore file needs to be more specific of what it is
ignoring.


LegacyKing:
LGTM

PR #3779, opened 2018-02-18
FORMAT and SCOPE separation

thpr:
This is cleanup work that separates the concepts of FORMAT and SCOPE.

This is necessary to have certain later behaviors work correctly, as
well as to generally avoid confusion (and bugs) later on. This supports,
but is not exclusively preparation for, Library version 147. (Meaning if
147 never happens or the relevant changes are backed out, this is still
a good commit and shouldn't be removed)

In particular, this (1) Shares the Global Scope name through a constant
(2) clearly drives getFact and dropIntoContext to have their first
argument as a FORMAT (3) Changes some tests to use SKILL rather than
EQUIPMENT since EQUIPMENT can't be used as a FORMAT at this time

To consider why getFact and dropIntoContext use FORMAT, consider
something like getFact("EQUIPMENT.PART","xx","FactName"). Right now both
EquipmentHead and EquipmentModifier use "EQUIPMENT.PART" as their scope
- so what would "xx" be? "EQUIPMENT.PART" (a scope) were allowed, then
the name is ambiguous, and the scopes start to control what object names
overlap. That would be BAD. Therefore, getFact must use a FORMAT, where
sub objects are distinguishable. (Note for the moment that since
Equipment is not necessarily unique on a PC - you can have two Longsword
- it can't be used as a proper format because we can't tell which one
the code is talking about...

### DIFF
--- a/code/src/java/pcgen/cdom/facet/ScopeFacet.java
+++ b/code/src/java/pcgen/cdom/facet/ScopeFacet.java
@@ -24,7 +24,7 @@ import pcgen.base.formula.base.VarScoped;
 import pcgen.base.formula.inst.ScopeInstanceFactory;
 import pcgen.cdom.enumeration.CharID;
 import pcgen.cdom.facet.base.AbstractItemFacet;
-import pcgen.cdom.inst.Dynamic;
+import pcgen.cdom.formula.scope.GlobalScope;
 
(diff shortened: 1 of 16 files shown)


thpr:
Redoing with slightly different result

PR #3787, opened 2018-02-19
Eliminate old Sort Requirement/Methods

thpr:
This also eliminates the SortKeyed/SortKeyRequired situation... reduces
them to one interface to avoid confusion


karianna:
@thpr Has a merge conflict now.

PR #3797, opened 2018-02-20
Force CATEGORY to be fixed for all objects

thpr:
A number of shifts here:

(1) ClassIdentity no longer throws exceptions from newInstance - it has to guarantee Reflection issues won't happen (verified during construction if necessary).

(2) ChooseInformation no longer has ClassIdentity, but rather class and persistent format of ClassIdentity.  Similar reasons to the game mode, where the information isn't always accessable.

(3) Category for Ability and CompanionMod are processed in the Loader, rather than a token.  While this may feel like backsliding on tokens, those tokens were never "editor compatible" anyway, so we might as well treat them separately.

(4) All the side effects of #3 are applied to the Context files, which removes a number of now unnecessary methods.


thpr:
Note Codacy will complain about 2 line lengths, but it is counting tabs as 8 characters, not 4, so it is a pair of false positives.  (We need to improve our Codacy config!)

code/src/java/pcgen/persistence/lst/AbilityLoader.java
```diff
@@ -58,25 +58,68 @@ public Ability parseLine(LoadContext context, Ability ability,
 			anAbility.setName(colToken.nextToken().intern());
 			anAbility.put(ObjectKey.SOURCE_CAMPAIGN, source.getCampaign());
 			anAbility.setSourceURI(source.getURI());
+			List<String> additionalTokens = new ArrayList<>();
+			boolean foundCategory = false;
+			while (colToken.hasMoreTokens())
```
karianna: I wonder if streams could be used here.

code/src/java/pcgen/persistence/lst/CompanionModLoader.java
```diff
@@ -40,22 +50,54 @@ public CompanionModLoader()
...
-		if (name == null)
+		StringTokenizer colToken = new StringTokenizer(lstLine, SystemLoader.TAB_DELIM);
+
+		List<String> allTokens = new ArrayList<>();
+
+		//Intentionally a dummy object with no Category - shouldn't be usable
+		CompanionMod loadable = new CompanionMod();
+
+		boolean foundType = false;
+		while (colToken.hasMoreTokens())
```
karianna: worth converting to stream?

thpr:
So longer term, yes, I think this could be done... but since this causes the need for new Predicate, I think that is a better longer term optimization than in this PR.

PR #3799, opened 2018-02-20
Eliminate Qualifier (useless wrapping)

thpr:
Since References now contain their persistent format, Qualifier holds nothing other than the reference.  It  thus has no value and can be eliminated to reduce a class and the associated wrapping


karianna:
I actually felt that Qualifier was a good wrapper (it had a good domain name telling us what it did on the tin..).

thpr:
I'm not attached to this one way or the other - closing

PR #3807, opened 2018-02-23
Eliminates the default Race object in PCGen (rely on data only)

thpr:
This is a "merged" PR between code and data, since it is necessary for them to interact in this change in order to function correctly.  Code and Data are kept separate in the underlying commits so as to allow easier code review/analysis.  

Net is that this is needed in order to allow Race LST files to have "required" FACTs.  The internal default race could never be assigned that, so load would fail... that's unreasonable, and it's better to simply have the default race out in a universal location for the data, so as to allow it to be defaulted to whatever is necessary in a given game mode/data set

The following are the changes for each area:

Code:

This moves a built-in object out of the code.

The <none selected> race is now a problem if the data team wishes to have a required FACT/FACTSET.  Therefore, this pulls out the default race, requiring the data to provide one.

The previous code awareness of <none selected> (including assigning it as the Race to the PC at character creation and sorting it to the top in the UI) has been replaced by code awareness of a GROUP called "Unselected"

This is an incompatible change that will require the data team to implement a Race with the appropriate characteristics (name and GROUP)

Data:

Initial Implementation of replacement for Race 'SIZE'

Prep work to move the default <none selected> UI race into data to allow for the REQUIRED:YES in FACT "BaseSize".

Leverage Global Data Controls


thpr:
Closing due to underlying issues

PR #3818, opened 2018-02-24
Eliminate Default Race

thpr:
This is a combined CODE/DATA commit for Eliminating the Default Race.  Each commit is kept separate, with detailed comments of the contents of the commit.


LegacyKing:
@evilpixie87 don't approve, just review this to see what Data Changes are involved.

PR #3819, opened 2018-02-24
Eldritch Heritage Variable Prerequisite

Llisandur:
Changed the Charisma requirement for Eldritch Heritage line of feats to a variable to allow other abilities to reduce or bypass the requirement.


Llisandur:
Ah, that would be why I had so few matches. Someone needs to go fix Bestiary 3 then.

PR #3824, opened 2018-02-25
Convert Alignment storage to a new variable... (technically a channel)

thpr:
Adds ALIGNMENTINPUT code control

This converts alignment from a FACET into an item controlled in the new formula system.  It also includes various back end infrastructure to allow an item to be "granted" to the PlayerCharacter based on a variable (in the future allowing such a capability for the data team once a token is developed


thpr:
Please ignore Codacy - one is false positive, one is future planning, the other is arguably useless

PR #3827, opened 2018-02-25
Update Datacontrol Usage

LegacyKing:
Okay @thpr - This should handle what you're looking for currently.


thpr:
Ok, I see what is happening here.  The ones that DO have alignment I will need to do when I put in the Alignment patch.

thpr:
Superceded and included in PR #3841

PR #3835, opened 2018-02-28
renamed poison to correct spelling

net-diver:
http://archivesofnethys.com/PoisonDisplay.aspx?ItemName=Blistercap%20spore%20poison


LegacyKing:
I'm sure the spelling correction is legit, but be advised, we need a citation from a Paizo sanctioned sources - PRD, their Books, etc.

evilpixie87:
I looked in the book. The correction IS legit.

PR #3863, opened 2018-03-06
Reduce direct usage of AbilityCategory.FEAT in tests

thpr:
This prepares for removal of the internal constant


thpr:
Note: Due to the "text used as a table for visual layout" in the two Count* test files, it's better to ignore Codacy line length on those files

PR #3864, opened 2018-03-06
Local Functions

thpr:
Note that this cannot use lambda functions since they do not have any
(proper) support for equality in Java.

This implements the following functions:
this() is usable in MODIFY: IF AND ONLY IF it's used on an object with a
resolvable scope.  (It can't be used in the Global Modifiers file, for
example.  Currently it also can't be used on Equipment Head/Modifiers)

source() is usable in MODIFYOTHER: IF AND ONLY IF it's used on an object
with a resolvable scope.  (It can't be used in the Global Modifiers
file, for example.  Currently it also can't be used on Equipment
Head/Modifiers).  source() represents the object which contained the
MODIFYOTHER: token.

target() is usable in MODIFYOTHER: to represent the target where the
Modification is eventually applied.  (it represents the object
identified by the first two arguments to MODIFYOTHER)

### DIFF
--- a/code/src/itest/pcgen/io/testsupport/AbstractSaveRestoreTest.java
+++ b/code/src/itest/pcgen/io/testsupport/AbstractSaveRestoreTest.java
@@ -24,7 +24,7 @@ import java.util.Collections;
 
 import compare.InequalityTesterInst;
 import junit.framework.TestCase;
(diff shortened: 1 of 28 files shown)


thpr:
Need to investigate out of memory issue

PR #3867, opened 2018-03-06
Add tests to ensure tokens don't accidentally capture the full file

thpr:
Unfortunately, the way we break up LST files uses StringTokenizer, which doesn't create new Strings.  Therefore, one accidental move by a token can ingest the full file stored on disk into memory - "permanently".  These additional tests (and one fix in TokenUtilities) are designed to detect that accidental capture (at least in some circumstances) to reduce the probability of it occurring.

Note this does rely on certain behavior by System.gc() that is not guaranteed, but we have nothing better in Java.


thpr:
Codacy should be ignored - the new String call is necessary to avoid a memory leak

PR #3869, opened 2018-03-07
fixed player companion resource selection folder structure for pathfinder

kypvalanx:
moved the pathfinder player companions into a subfolder on the resource selection screen.


Llisandur:
Why? None of the other resources are a subfolder of the core rulebooks.

PR #3876, opened 2018-03-08
Random load errors!

LegacyKing:
@evilpixie87 note the Alter Winds - something in the SPACE between the words was causing grief.

@Llisandur Please make sure when descriptions contain "%" that you escape those out with a second "%"


Llisandur:
Sorry about that. I was going fast and forgot about them.

PR #3898, opened 2018-03-12
Newsource 333

LegacyKing:


LegacyKing:
@Nylanfs - take a look at the LOGO file, and see if you can work the resize magic on it.

PR #3911, opened 2018-03-14
Sorcerer Rebuild - OOG Complete

Llisandur:


Llisandur:
Failed due to darkvision DESC. I changed it to the shorter description text from Bestiary 4. Should I revert it or fix the test?

PR #3920, opened 2018-03-18
Fantasy Grounds Export Sheet Update

Llisandur:
Updated version of the export sheet for Fantasy Grounds. Made specifically for Pathfinder and Fantasy Grounds 3.3.4


Llisandur:
I'll take a look at cleaning it up sometime. There's still some things that need to be fixed, mainly spells, but it works a lot better than the previous version.

PR #3931, opened 2018-03-21
Installer fixes

javydreamercsw:


javydreamercsw:
Really hope this was tested and fixed the issue without any side effects.

karianna:
@javydreamercsw I'd assumed you'd tested this locally :-).  Was there some sanity testing done or should we roll this back?

javydreamercsw:
Yes I did. I meant that it didn't break any other task, etc.

I built and tested the installer, nothing else. I hope the tests would have failed if something else broke.

karianna:
Should be good :-)

PR #3942, opened 2018-03-26
Fix bug: DATA-3600

LegacyKing:
Evolved Companion Feat doesn't work for Cavalier mounts.


LegacyKing:
@draconswan that should be sufficient to correct the issue.

draconswan:
I tested this and it worked.  Thanks Andrew!

PR #3945, opened 2018-03-26
Cleanup (mostly UI) code

thpr:
Remove unused code
Add generics
Remove comments
Remove unused parameters
etc.

This is entirely a code hygiene PR.  The more major changes that appear are situations where a null check was performed but was unnecessary, thus causing an indentation change.


javydreamercsw:
Would suggest addressing the code review observations.

PR #3947, opened 2018-03-27
More Cleanup

thpr:
Mostly Known good Casts (suppress warning)


karianna:
A couple of codacy fixes and then this is good to go.

PR #3948, opened 2018-03-27
Eliminate HPFORMULA from Game Mode

thpr:
Eliminates unused code


code/src/java/pcgen/cdom/facet/HitPointFacet.java
```diff
@@ -314,25 +314,18 @@ public void rollHP(CharID id, PCClass pcc, int level, boolean first)
 						+ (int) bonusCheckingFacet.getBonus(id, "HD", "MAX;CLASS."
 							+ pcc.getKeyName());
 
-			if (SettingsHandler.getGame().getHPFormula().isEmpty())
+			if (first && (level == 1) && SettingsHandler.isHPMaxAtFirstLevel()
```
karianna: could extract this to a local function which says what this is really testing

PR #3954, opened 2018-03-28
Formula Library Version 165

thpr:
(rebuilds how FormulaManager is created)


code/src/itest/plugin/lsttokens/editcontext/template/FaceIntegrationTest.java
```diff
@@ -52,9 +52,9 @@ public void setUp() throws PersistenceLayerException, URISyntaxException
 		super.setUp();
 		TokenRegistration.register(m);
 		primaryContext.getVariableContext().assertLegalVariableID(
-			primaryContext.getActiveScope(), opManager, "Face");
+			"Face", primaryContext.getActiveScope(), opManager);
```
javydreamercsw: Consider having Strings like "Face" as a final static somewhere. It's just error prone to typos and makes it harder if it needs to change for whatever reason later.

thpr:
Improvement made.

PR #3955, opened 2018-03-28
Allow VariableListener objects to be registered with a priority

thpr:


thpr:
Correct - order of listeners under the same priority is not guaranteed (if a controlled processing order is required, a priority should be used)

VariableID objects are constructed under controlled conditions, so they are always "valid" in that sense.  or is that not what you mean by valid?

javydreamercsw:
Yes, exactly what I meant @thpr. Thanks!

PR #3957, opened 2018-03-31
data-3605

xenusuk1:
fixed alchemical TYPE entries.
This is a later version of file attached to the bug.
update to add SPROP descriptions, add some CHARGES and BONUS, add splash weapon props
Next time I'll update the repo file directly.
Happy to be added to data monkey list.


LegacyKing:
Looks good. To prevent issues in your repo, you should branch from master for each separate feature or bug you want to work on.

PR #3971, opened 2018-04-01
Interface Tokens

thpr:
Leverages the Proxy system from -base

This system allows a load token to use an interface rather than the
object class (or ancestor class).  This allows multiple different,
otherwise unrelated objects to use the same load token.

This sets up for allowing Dynamic (which is not - and shouldn't be - a
CDOMObject) to handle variables.

This sets up a simplification of GlobalModifiers as well.

Overall, this also sets up the ability to do a simplification of the
token system - allowing them to ignore the context in many cases and act
directly on an interface (thus making the code clearer)

### DIFF
--- a/code/src/itest/tokencontent/testsupport/AbstractContentTokenTest.java
+++ b/code/src/itest/tokencontent/testsupport/AbstractContentTokenTest.java
@@ -32,6 +32,7 @@ import pcgen.core.Ability;
 import pcgen.core.Campaign;
 import pcgen.core.Deity;
 import pcgen.core.Domain;
+import pcgen.core.Equipment;
 import pcgen.core.EquipmentModifier;
 import pcgen.core.PCCheck;
 import pcgen.core.PCClass;
@@ -174,6 +175,8 @@ public abstract class AbstractContentTokenTest extends AbstractTokenModelTest
 	public void testFromEqMod() throws PersistenceLayerException
 	{
 		EquipmentModifier source = create(EquipmentModifier.class, "Source");
+		Equipment e = create(Equipment.class, "Parent");
+		source.setVariableParent(e);
 		processToken(source);
 		assertEquals(baseCount(), targetFacetCount());
 		activeEqModFacet.add(id, source, this);
--- a/code/src/itest/tokencontent/testsupport/AbstractContentTokenTest.java
+++ b/code/src/itest/tokencontent/testsupport/AbstractContentTokenTest.java
@@ -175,8 +175,8 @@ public abstract class AbstractContentTokenTest extends AbstractTokenModelTest
 	public void testFromEqMod() throws PersistenceLayerException
 	{
 		EquipmentModifier source = create(EquipmentModifier.class, "Source");
-		Equipment e = create(Equipment.class, "Parent");
-		source.setVariableParent(e);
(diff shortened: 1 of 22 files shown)


code/src/itest/tokencontent/testsupport/AbstractContentTokenTest.java
```diff
@@ -174,6 +175,8 @@ public void testFromDomain() throws PersistenceLayerException
 	public void testFromEqMod() throws PersistenceLayerException
 	{
 		EquipmentModifier source = create(EquipmentModifier.class, "Source");
+		Equipment e = create(Equipment.class, "Parent");
```
karianna: what's an `e`? ;-)

code/src/java/pcgen/cdom/facet/GrantedVarFacet.java
```diff
@@ -59,18 +56,19 @@
...
+		String[] list = cdo.getGrantedVariableArray();
+		if (list.length == 0)
+		{
+			return;
+		}
 		Object source = dfce.getSource();
-		if (list != null)
+		CharID id = dfce.getCharID();
+		ScopeInstance inst = scopeFacet.get(id, cdo);
+		for (String s : list)
```
karianna: What's an `s`? :-)

code/src/java/pcgen/rules/persistence/TokenSupport.java
```diff
@@ -90,21 +154,55 @@
...
+		StagingInfo<R, W> info = stagingFactory.produceStaging(ifToken.getReadInterface(),
+			ifToken.getTokenClass());
+		ParseResult parse;
+		try
+		{
+			parse = ifToken.parseToken(context, info.getWriteProxy(), tokenValue);
+		}
+		catch (IllegalArgumentException e)
+		{
+			e.printStackTrace();
```
karianna: Use a logger?

code/src/java/pcgen/rules/persistence/TokenSupport.java
```diff
@@ -220,89 +318,56 @@
...
+		LoadContext context, T loadable)
 	{
-		Class<? super T> cl = (Class<T>) cdo.getClass();
-		CDOMToken<?> token = null;
-		while (token == null)
+		Collection<String> set = new WeightedCollection<>(
+                String.CASE_INSENSITIVE_ORDER);
+		@SuppressWarnings("unchecked")
+		Class<T> cl = (Class<T>) loadable.getClass();
+		for (CDOMInterfaceToken<?, ?> ifToken : TokenLibrary.getInterfaceTokens())
```
karianna: Better to call that `interfaceToken` if could be misconstrued.

code/src/utest/plugin/lsttokens/ModifyLstTest.java
```diff
@@ -49,29 +48,29 @@ public void setUp() throws PersistenceLayerException, URISyntaxException
...
 		return token;
 	}
 
-	@Test
-	public void testInvalidObject() throws PersistenceLayerException
-	{
-		assertFalse(token.parseToken(primaryContext, new Campaign(),
-				"MyVar|ADD|3").passed());
-	}
+//	@Test
```
karianna: Commented out code?

PR #3979, opened 2018-04-02
Convert ALLOW and ENABLE to String|Formula format

thpr:
Adds the String to be able to look up an INFO and provide info to a user

### DIFF
--- a/code/src/java/plugin/lsttokens/EnableLst.java
+++ b/code/src/java/plugin/lsttokens/EnableLst.java
@@ -1,13 +1,15 @@
 package plugin.lsttokens;
 
-import pcgen.base.formula.inst.NEPFormula;
-import pcgen.base.lang.StringUtil;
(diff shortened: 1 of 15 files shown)


code/src/java/plugin/lsttokens/EnableLst.java
```diff
@@ -33,25 +35,46 @@ public String getTokenName()
 	}
 
 	@Override
-	protected ParseResult parseNonEmptyToken(LoadContext context, CDOMObject obj,
+	protected ParseResult parseTokenWithSeparator(LoadContext context, CDOMObject obj,
 		String value)
 	{
 		return new ParseResult.Fail("Not supported since it is not monitored in an ongoing fashion");
+//		int pipeLoc = value.indexOf(Constants.PIPE);
```
karianna: lots of commented out code?

PR #3989, opened 2018-04-03
Move libs back to root

javydreamercsw:
As requested by @LegacyKing.


karianna:
I think this needs rebasing or something?  Seems to reapply some of Tom's previous work.

javydreamercsw:
I did rebase when I updated from upstream.

LegacyKing:
Doesn't look like it. If you rebased, then the only commits should be exclusively those on the branch, nothing else. The easiest fix for this, update master, branch from master, cherry pick your commits back in, then push.

javydreamercsw:
Will recreate.

PR #3990, opened 2018-04-04
Move libs to root as requested

javydreamercsw:


javydreamercsw:
This should be cleaner @LegacyKing.

javydreamercsw:
The previous failure was related to something installing perl that timed out. Since it's no longer needed after the last PR you merged I updated with upstream.

PR #3991, opened 2018-04-05
Replace gendatalist

javydreamercsw:
Remove perl scripts and replace the ones still used with a gradle script counterpart.


javydreamercsw:
Finally seems to work. @LegacyKing could you take the installer for a spin?

PR #3993, opened 2018-04-07
corrected spelling

net-diver:
http://paizo.com/pathfinderRPG/prd/ultimateEquipment/artifactsAndOthers/artifacts.html


mjmeans:
Won't this key change require a migration entry to avoid an error on load of a PCG file?

PR #4007, opened 2018-04-14
WIP [Pathfinder] young age

mjmeans:
This is my suggestion on how to deal with young characters and the Youth age category. It eliminates the optional rule to apply Youth stat mods, and adds an UNDEFINED age category that is any age younger than the minimum Youth age. This UNDEFINED age category applies no mods at all. I think this will keep all kits that do not set AGE and all characters that do not use AGE from breaking. However, it will break any existing character that has an age set in the Youth age range that did not apply the Youth specific age mods. A player that has such a character and does not want to have the penalties applied should set the optional rule that removes age related penalties.

I don't know if this breaks any of the existing test characters or whether any of them set an AGE other than 0. If Jenkins runs those test characters, then we should see.


mjmeans:
If at a later time, a new age category for Child or Baby get's defined those can be added in the same way, but UNDEFINED should remain at age 0.

PR #4014, opened 2018-04-15
GROUPING outside of MODIFYOTHER

thpr:
Add Generic Child Methods
Allow FACT to be grouped in MODIFY* as well


thpr:
Closing - will fail after Library 167 merge

PR #4015, opened 2018-04-15
Minimalist fix for Java9

thpr:
In Java8, the internal API of Java was loose on generics.  This was tightened in Java 9.  The minimalist fix (given we need to replace Swing anyway) is to go back to ignoring generics.


LegacyKing:
@thpr conflicted

karianna:
OK, gradle conflict now

PR #4020, opened 2018-04-15
GROUPING outside of MODIFYOTHER

thpr:
Add Generic Child Methods
Allow FACT to be grouped in MODIFY* as well


thpr:
Please ignore codacy for some of these issues - package is intentional, and Equipment has always been long.

PR #4027, opened 2018-04-16
Add BONUS:WEAPONPROF=x|STATDAMAGE

thpr:
Add a new BONUS for a fix required for Finesse to work properly


gameguy95:
Would this work for any stat? For example, the Kineticist in Pathfinder (https://www.d20pfsrd.com/occult-adventures/occult-classes/kineticist/) has an energy "kinetic blast", which is a "ranged touch attack that deals an amount of damage equal to 1d6 **+ 1/2 the kineticist’s Constitution modifier,** increasing by 1d6 for every 2 kineticist levels beyond 1st." 
It's bugged me for the longest time that that attack only shows as 1d6+0 on the PcGen character sheet, and I couldn't even create a custom version to fix the damage because I wasn't sure how to add the CON modifier to the damage. Would this allow me to just add `BONUS:WEAPONPROF=Kinetic Blast|(CON / 2)` for that particular case?

PR #4052, opened 2018-05-01
Equipment Rework - CR General and Magic Items

Llisandur:
Continuation of the rework that brings the rest of the Core Rulebook items in line with the changes to the weapons and armor.


code/testsuite/csheets/pf_Cleric.xml
```diff
@@ -1566,9 +1540,35 @@ BIO
...
+			</size>
+			<special_properties></special_properties>
+			<type>GOODS.TOOLS.SPELL COMPONENT.DIVINE FOCUS.HOLY SYMBOL</type>
+			<weight>1</weight>
+			<bonuslist></bonuslist>
+		</item>
 		<total>
 			<weight>59.5 lbs.</weight>
-			<value>926.6 gp</value>
+			<value>901.6 gp</value>
```
LegacyKing: Why did this value change on a test character?

code/testsuite/csheets/pf_Cleric.xml
```diff
@@ -1701,14 +1701,14 @@ BIO
 				<bonuslist></bonuslist>
 			</item>
 			<item>
-				<name>Blanket (Winter)</name>
+				<name>Blanket</name>
```
LegacyKing: Uh, Winter Blanket becomes Blanket?

code/testsuite/csheets/pf_Paladin.xml
```diff
@@ -1088,8 +1088,8 @@ BIO
 		<weapon>
 			<common>
 				<name>
-					<short>Oil (1 Pint Flask)</short>
-					<long>Oil (1 Pint Flask)</long>
+					<short>Oil</short>
+					<long>Oil</long>
```
LegacyKing: We no longer designating how much as per the CR?

code/testsuite/csheets/pf_Paladin.xml
```diff
@@ -1767,7 +1767,7 @@ BIO
 				<long>Medium</long>
 				<short>M</short>
 			</size>
-			<special_properties>+5 alchemical bonus to Fortitude saves against poison for 1 hour</special_properties>
+			<special_properties></special_properties>
```
LegacyKing: Is this intentional?

code/testsuite/csheets/pf_Paladin.xml
```diff
@@ -2346,7 +2346,7 @@ BIO
 		</item>
 		<total>
 			<weight>98 lbs.</weight>
-			<value>7563.45 gp</value>
+			<value>7538.45 gp</value>
```
LegacyKing: Value changes?

code/testsuite/csheets/pf_Cleric.xml
Llisandur: The amount off is the value of a silver holy symbol (25 gp). I was attempting to reduce the holy symbol to a single mod-required item, but the migration file doesn't allow EQMODs to be added. I thought I had fixed it, but apparently not.

code/testsuite/csheets/pf_Paladin.xml
Llisandur: The amount off is the value of a silver holy symbol (25 gp). I was attempting to reduce the holy symbol to a single mod-required item, but the migration file doesn't allow EQMODs to be added. I thought I had fixed it, but apparently not.

code/testsuite/csheets/pf_Cleric.xml
Llisandur: It's listed as "winter blanket" in the CR as a leftover from 3.5, then there's a lighter weight and cheaper cost item simply called "blanket" in the APG. Then the item "blanket" is listed in UE with the original weight and cost as the "winter blanket" with the description from the APG.

code/testsuite/csheets/pf_Cleric.xml
```diff
@@ -1458,14 +1432,14 @@ BIO
 				<short>M</short>
 			</size>
 			<special_properties>Cures 1d8+1 points of damage</special_properties>
-			<type>MAGIC.POTION.CONSUMABLE</type>
+			<type>MAGIC.POTION.CONSUMABLE.ENHANCEMENT</type>
```
skylarhiebert: Are one time recovery items really enhancements, since they have no lasting effect?

Llisandur: That gets added by the EQMOD "Special Ability ~ Spell Effect / Single Use".

code/testsuite/csheets/pf_Cleric.xml
LegacyKing: Okay, and is this test unit pulling from the CR, the APG or the UE? If it's the CR, and the UE isn't loaded, then the CR name should be prominent.

Llisandur: Well the cleric test loads both Core Rulebook and Advanced Player's Guide. Which version is it supposed to have?

Llisandur: Also, it's a pain to nearly impossible to change the display name. There really should be a MOD tag to change it.

LegacyKing:
Please fix conflicts, and then I'll review this again. Thanks.

Llisandur:
I'm fairly sure the conflicts are because this is now a few commits behind. What would be the best way to resolve them?

PR #4071, opened 2018-05-15
defined and assigned value to missing variable

net-diver:
Variable "AasimarSpellLikeTimes" was never defined or given a value despite being needed for SLA resulting in it being left at 0 in both spell-like ability listing and ability descriptor


LegacyKing:
Not sure if leaving it untyped is a good idea... I know I had a bug relating to this recently, but I don't have the time to investigate...

PR #4080, opened 2018-05-27
CODE-3257: fix pcgen.sh a bit

grimreaper:
- We already cd to correct directory at the top of the file. This one is
extraneous and has incorrect quoting
- remove broken '2> /dev/null' in conditional
- add quotes

### DIFF
--- a/code/pcgen.sh
+++ b/code/pcgen.sh
@@ -4,7 +4,7 @@ if command git rev-parse >/dev/null 2>&1
 then
   cd "$(git rev-parse --show-toplevel)/output"
 else
-  cd "$(dirname $0)"
+  cd "$(dirname "$0")"
 fi
 
 available_memory="unknown"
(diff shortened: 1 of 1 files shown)


code/pcgen.sh
```diff
@@ -82,5 +82,5 @@ done
 #     -Dpcgen.filter=/path/to/filter.ini
 #     -Dpcgen.options=/path/to/options.ini
 
-cd $(dirname $(readlink -f $0))
+# shellcheck disable=SC2086
```
LegacyKing: Why is this commented out? Doesn't follow the principles of clean code.

grimreaper: This is a "lint supression". It is supposed to be a comment. It tells the tool 'shellcheck' not to report `$javaargs` as unquoted. we want it as such in order to expand.

LegacyKing: Gotcha.

LegacyKing: I should install a linux build to properly test these linux only issues...

PR #4084, opened 2018-06-02
gradlew with JDK10 has issues with spotbug version '1.6.1'

tildar:
, update to '1.6.2'

JDK9 is end of life. Building with gradle and Java10 had issues with spotbug version 1.6.1
Also adjusted toolVersion from 3.1.2 to 3.1.3 in two places as this seems to be current version.

Build with ./gradlew fullZip produced usable build.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ plugins {
     id 'application'
     id 'build-dashboard'
     id 'checkstyle'
-    id "com.github.spotbugs" version '1.6.1'
+    id "com.github.spotbugs" version '1.6.2'
     id 'java'
     id 'maven'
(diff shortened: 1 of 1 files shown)


LegacyKing:
Just an FYI, the official release is currently locked at JDK8. @PCGen/code-team-reviewers please chime in if we are okay to upgrade to JDK9 or JDK10

grimreaper:
JDK 9 for an official release is likely reasonable at this point. I'm unsure about JDK 10

tildar:
JDK 9 reached end of life (and end of public support) in March 2018.

I am not suggesting a change from official support at JDK8.

PR #4086, opened 2018-06-05
fix for bug in getNumAttacks() in PlayerCharacter.java

tildar:


LegacyKing:
Interesting. The output ends up correct though, is this due to the sheet handling the formula then?  If so, then we will need to deactivate the sheet logic across the spectrum to avoid bad output, yes?

tildar:
I'll see what I can do about unit tests.

mertonmonk:
This is conceptually the same thing.  One way is max 4 attacks but no fewer than 1.  The other is min 1 attack but no more than 4.  Tomayto Tomahto.  The results should be identical.  Is this a bug or just a code tool complaint?

tildar:
It seems I'm over my head trying to do unit tests. It may take me quite a while to figure out how to do.
Perhaps just drop this pull request until i can get that done?

karianna:
You've actually partly written the unit tests in your comment :-).  See if you can find the PlayerCharacterTest class and see if there's any tests for `getNumAttacks()`.  If not then I think you can add one something like

```
@Test
public void baseAttackBonusOf3GivesYou3Attacks() {
    PlayerCharacter pc = new PlayerCharacter();
    pc.setBaB(3);
    assertEquals(3, pc.getNumAttacks());
}
```

tildar:
@karianna , Thanks for the re:

I have been looking at adding the test to  code/src/test/pcgen/core/PlayerCharacterTest.java since the request for unit tests was made.
My problem has been that there does not seem to be a simple pc.setBaB() or equivalent, and that I have been failing to set up a character correctly enough (with for example class fighter, and then incrementing its lvl as needed) that the baseAttackBonus() call in getNumAttacks() returns useful numbers. I had hoped to find a test of baseAttackBonus() that I could extend on, but there is not such a test in that file.

I guess I will keep working at it until i figure it out!

And to add to the issue:
While I know that getNumAttacks() returning 1 no matter what baseAttackBonus() returns is ABSOLUTELY NOT CORRECT, I also am not 100% sure that my changes meshes up with my understanding of  table 3-1 in the 3.5 PHB either (if that's even the correct place to be referencing?). Nor have I been able to find an official definitive formula of how the calculation from BAB to APR is supposed to be done, if such a formula even exists.

tildar:
Okay, I finally have results that match up with 3-1 from PHB. Turns out there was some int math issues, and that the BaB/5 needing to be rounded up to make everything happy.  See https://ideone.com/zM88wd

Still fighting with the unit test.

LegacyKing:
The progression is simple.  min(floor((BAB-1)/5),3) - That is how many additional attacks will be displayed. As long as it actually comes out right, I do not think anyone will care about the formula. 

Attacks progress like so: 
BAB 0-5 = 1 Attack
BAB 6-10 = 2 Attacks 
BAB 11-15 = 3 Attacks 
BAB 16+ = 4 Attacks

tildar:
I've added unit tests. Please review at your convenience.

PR #4089, opened 2018-06-08
Fix bug: CODE-3259

milesoc:
This is a proposed fix to the abovementioned JIRA issue. It appears that sometimes the spell level list in question is not sorted. 

I'd like review from those more familiar with this class to ensure the assumption that the spell list is sorted is not relied upon elsewhere, which would indicate the bug comes from a different source.

### DIFF
--- a/code/src/java/pcgen/core/analysis/SpellLevel.java
+++ b/code/src/java/pcgen/core/analysis/SpellLevel.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.Collections;
 import pcgen.base.util.HashMapToList;
 import pcgen.cdom.base.AssociatedPrereqObject;
 import pcgen.cdom.base.CDOMList;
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/core/analysis/SpellLevel.java
```diff
@@ -100,8 +100,17 @@ public static int getFirstLvlForKey(Spell sp, CDOMList<Spell> list,
 			List<Integer> levelList = wLevelInfo.getListFor(list);
 			if (levelList != null)
 			{
-				// We assume those calling this method know what they are doing!
-				return levelList.get(0);
+				// In specific situations, this list may not be sorted, so we won't assume it is
+				int lowestLevel = levelList.get(0);
+				for (int i = 1; i < levelList.size(); i++)
```
karianna: you can use a stream / filter here.

milesoc:
@grimreaper Agreed - it appeared that at some point this being sorted was assumed, so I'm not sure where the disconnect is; nonetheless it feels like best practice to assume no sorting since the method doesn't explicitly take a sorted list.

PR #4106, opened 2018-06-25
PFS season 8 updates

mjmeans:
This PR requires my previous PR to be merged first.


LegacyKing:
Mark - https://github.com/PCGen/pcgen/pull/4106 and https://github.com/PCGen/pcgen/pull/4113 and https://github.com/PCGen/pcgen/pull/4113 ALL contain the exact same commits! Please clean these branches up.

mjmeans:
I cleaned up the conflicts. I have no idea how to remove the commits that were already pulled into master. I was assuming that git would be intelligent and see that the otherwise duplicate commits were already done and only show the changes. I dislike git.

LegacyKing:
I'm not going to go into git being smart, I just know that you should not just build upon previous non-master branches. If you want a single big branch, then so be it, but different branches should always be based upon stable master.

PR #4115, opened 2018-06-29
[Pathfinder] implement piranha strike tempbonus

mjmeans:


mjmeans:
I should have based this on master. Will resubmit.

Issue #4109, opened 2018-06-26
Travis-ci is having issues with wget and network failures


tildar:
We seem to have travis fail to build fairly often due to this issue.

I'm trying to find a way to make the wget for dl of zlib128-dll.zip to be more robust, and will try testing changes on my own travis first for now.

If anyone has idea's please speak up.

tildar:
As I work on this, I have been curious what we do with /tmp/zlib.zip after its downloaded?

tildar:
I eventually tracked the jun 26th  failed downloads to some support notices at Sourceforge re: ddos issues, which were fixed later in the day.  I'm still going to keep looking at ways to avoid the build failing out in such cases. Maybe find a second source to dl from if the first fails? -OR- Maybe we don't even need to it to dl, I have not found were it gets used yet.

LegacyKing:
Can we close this issue?  Was this Pull Requested?

PR #4119, opened 2018-07-05
Remove the download of zlib128-dll.zip #4109

tildar:
I can not find any reason for doing the download.
It does not seem to be used in any way, therefore I suggest we
stop downloading it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,12 @@
 language: java
 dist: trusty
 
+#run in Fully-virtualized environment.
+#sudo: required
+
+#run in a Container-based environment.
+sudo: false
+
 matrix:
   include:
     - os: linux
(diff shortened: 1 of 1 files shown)


LegacyKing:
@javydreamercsw these are changes you made. Can you answer what the file is used for?

javydreamercsw:
If I remember correctly this is used by the installer building process.

tildar:
If it's necessary for the building process, why did travis build successful without it in this case?

karianna:
The travis build isn't a full windows deployment build right?

tildar:
@karianna I'm not sure weather or not travis being a "windows deployment build" is even an issues here. 

The issues as I see it is that we currently have TRAVIS configured to dl the zlib file, which seems to me to indicate that it should be needed by the build ON TRAVIS is some way. There are times when this DL fails to work for various reasons, causing the build to fail prematurely. Removing the DL did not cause the build to fail, and removes the failure point that can cause an issue. Thuse I have to question why travis is configured to DL the zlib?

javydreamercsw:
This is because that library is needed for NSIS to work. Travis tries to build the installer as well due to other nightmares in the past when a release was crippled with the installer not working for whatever reason.

javydreamercsw:
I vote against merging this unless the team is willing to not check the installer works on a regular basis which was requested not long ago resulting in part on the changes being removed.

tildar:
I'll defer to javydreamercsw o this then. still seems fishy to me, but not worth the hassle to keep digging.

PR #4168, opened 2018-08-16
CODE-3274: Fix PCGen hanging on the splash screen on MacOS with Java 10

Zaister:
PCGen only tests for Java version 9 to skip the non-functioning MacGUIHandler.initialize(). Changed this to be future-proof for higher versions than 9..


Zaister:
Found a better solution.

PR #4169, opened 2018-08-16
CODE-3274: Fix the MacGUIHandler to run with Java 9+

Zaister:
MacGUIHandler was no longer usable with Java 9, but was skipped for exactly Java version 9 only. Rewrote MacGUIHandler class to work with Java 9+, so it can be used again.

### DIFF
--- a/code/src/java/pcgen/gui2/PCGenUIManager.java
+++ b/code/src/java/pcgen/gui2/PCGenUIManager.java
@@ -46,8 +46,7 @@ public final class PCGenUIManager
 
 	public static void initializeGUI()
 	{
-		// Don't start macOS GUI if we're on Java 9 since it doesn't work.
-		if (SystemUtils.IS_OS_MAC_OSX && (System.getProperty("java.runtime.version").charAt(0) != '9'))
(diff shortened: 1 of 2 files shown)


Zaister:
As far as I know PCGen doesn't run on Java 8 anymore?

Zaister:
This might be a problem. The code that can run on Java 8 does not compile on Java 9+, and the code that can run on Java 9+ does not compile on Java 8. As far as I can see, it might not be possible to compile a version that can run on Java 8 **and** 9+. Any ideas?

Zaister:
But I'm not sure we're actually still supposed to compile unter Java 8. If I try that on MacOS I'm getting this error:

Could not target platform: 'Java SE 9' using tool chain: 'JDK 8 (1.8)'.

Zaister:
build.gradle has:

sourceCompatibility = "1.9"
targetCompatibility = "1.9"

I think that means PCGen is no longer intended to compile or run on Java 8, right?

karianna:
Yeah I think you're right :-).

Cheers,
Martijn

PR #4189, opened 2018-08-20
Fix bug: DATA-3699

LegacyKing:
Blood of Fiends: Tiefling Racial Trait Errors


mjmeans:
Not merged yet? I think the broken previous DATA-3699's are causing later PRs to fail. I'm assuming that after this PR resolves that, I will have to rebase my PRs so they will pass.

PR #4213, opened 2018-09-10
[build] remove gradle lint

grimreaper:


karianna:
Build failed because you still reference lintGradle in the reporting build script

grimreaper:
ah, right, I had this fixed as two separate commits. Fixing now.

PR #4220, opened 2018-09-11
[build] remove gradle lint

grimreaper:


karianna:
Why are we removing this?

grimreaper:
> Why are we removing this?

It was in the commit, but not the PR description. Updated

PR #4222, opened 2018-09-11
[io] Use modern sort API

grimreaper:


grimreaper:
FTR I often already have commits with fixes that just didn't make into this branch. Give me a few days and it'll usually be fixed.

PR #4224, opened 2018-09-11
multiple: replace merge with merge

grimreaper:


code/src/java/pcgen/cdom/facet/analysis/VariableFacet.java
```diff
@@ -90,13 +90,8 @@ public void dataRemoved(DataFacetChangeEvent<CharID, CDOMObject> dfce)
...
 	{
 		Map<VariableKey, Map<Formula, Set<CDOMObject>>> map = getConstructingCachedMap(id);
-		Map<Formula, Set<CDOMObject>> subMap = map.get(vk);
-		if (subMap == null)
-		{
-			subMap = new HashMap<>();
-			map.put(vk, subMap);
-		}
-		Set<CDOMObject> sources = subMap.get(formula);
+        Map<Formula, Set<CDOMObject>> subMap = map.computeIfAbsent(vk, k -> new HashMap<>());
```
karianna: Whitespace not consistent

karianna: Would be nice to expand `vk` to `variableKey` as well.  And `map` to a more meaningful name (what is it a map of?), same with `subMap` etc.  Part of the challenge of the CDOM framework code is that it's very abstract, so variable names that describe the semantics in PCGen terms helps.

code/src/java/plugin/exporttokens/HitDiceToken.java
```diff
@@ -143,15 +143,7 @@ public static String getMediumToken(PlayerCharacter pc)
...
-					Integer num = hdMap.get(hitDie);
-					if (num == null)
-					{
-						hdMap.put(hitDie, 1);
-					}
-					else
-					{
-						hdMap.put(hitDie, num.intValue() + 1);
-					}
+					hdMap.merge(hitDie, 1, (a, b) -> a.intValue() + b);
```
karianna: again, perhaps rename a and b to something more meaningful

PR #4230, opened 2018-09-11
[gmgen] Fix errors with gmgen on macOS

grimreaper:
Instead of changing OSXAdapter to modern code just use MacGUIHandler as
it would be nearly identical. It even adds one additional handler.


code/src/java/pcgen/gui2/PCGenUIManager.java
```diff
@@ -17,17 +17,16 @@
...
-
 import gmgen.GMGenSystem;
+import org.apache.commons.lang3.SystemUtils;
 import pcgen.gui2.dialog.PreferencesDialog;
 import pcgen.gui2.facade.GMGenMessageHandler;
 import pcgen.gui2.plaf.MacGUIHandler;
 import pcgen.pluginmgr.PluginManager;
 import pcgen.system.Main;
 
+import javax.swing.*;
```
karianna: prefer explicit imports

PR #4234, opened 2018-09-12
[*] Use char instead of escape code

grimreaper:


grimreaper:
This is unicode, just not the escape spelled out. Either way: I'll close.

PR #4239, opened 2018-09-13
[code] use List.copyOf over Collections.unmodifiableList

grimreaper:


code/src/java/pcgen/cdom/base/ConcretePrereqObject.java
```diff
@@ -66,7 +66,7 @@
 		 * 5.14 - to be changed once 5.14 code is gone and this can be changed
 		 * to return Collection or Set (or perhaps ListSet?)
 		 */
-		return Collections.unmodifiableList(new ArrayList<>(thePrereqs));
+		return List.copyOf(thePrereqs);
```
karianna: If you're making this change the comment above should be removed yes?  Also Travis fails because of Collections unused import.

PR #4241, opened 2018-09-13
[test] Enable using junit 5

grimreaper:
This runs all existing tests with vintage junit 3/4 runners but new
tests should be junit 5.


grimreaper:
I had this passing locally.Maybe I missed a commit. Will re-open.

PR #4242, opened 2018-09-14
[PlayerCharacter] found is always true

grimreaper:


grimreaper:
Probably. I did look but didn't see any obvious flaw.

PR #4243, opened 2018-09-14
[PlayerChar] return void

grimreaper:


grimreaper:
> There's still a return false; in this method

Ugh. I'm annoyed this compiles.

PR #4247, opened 2018-09-15
[build] Move tool versions into common location

grimreaper:


grimreaper:
This does split the versioning information into two different files, but overall simplifies the number of places that information about the reporting tools is stored.

grimreaper:
Ugh, didn't realize the version numbers were different. Fixing.

PR #4249, opened 2018-09-15
[notes] Move proxied method into the caller

grimreaper:


grimreaper:
This error seems spotbugs related and I don't understand it.

PR #4251, opened 2018-09-16
update pmd to 6.7.0

grimreaper:
Depends on #4247


grimreaper:
There is a slight bug where PMD ignores `@link` for unused imports. This only affects two classes, is known, and will be fixed in the future, so I just remove them for now.

PR #4253, opened 2018-09-18
[gui2] Merge DesktopBrowserLauncher into Utilities

grimreaper:
The utilities code appears to be a pass-through into the desktop code,
so just merge them all together and make the logic slightly less
complex.

Another option is to keep DesktopBrowserLauncher but drop the Utilities
code but this felt a bit better. Not sure why.


grimreaper:
actually, looking at other code, I want to go the other direction :\

PR #4277, opened 2018-09-29
[settings] garbage collect legacy settings

grimreaper:
There appear to be no further settings in legacy.ini, so rm it.


grimreaper:
I lied: twas looking in the wrong place.

PR #4285, opened 2018-09-30
[gui2] replace Hashtable with HashMap; simplify

grimreaper:


code/src/java/pcgen/gui2/tabs/TabTitle.java
```diff
@@ -33,15 +34,13 @@
...
+	static final String ENABLED = "enabled"; //$NON-NLS-1$
 	public static final String TITLE = "title"; //$NON-NLS-1$
-	public static final String ICON = "icon"; //$NON-NLS-1$
-	public static final String TOOLTIP = "tooltip"; //$NON-NLS-1$
+	static final String ICON = "icon"; //$NON-NLS-1$
+	static final String TOOLTIP = "tooltip"; //$NON-NLS-1$
 	public static final String TAB = "tab"; //$NON-NLS-1$
 	private final PropertyChangeSupport support;
-	private final Hashtable<String, Object> table;
+	private final Map<String, Object> table;
```
karianna: If it's truly a map, why are we calling it a table?  What's it a map of?  Use that as the variable name

PR #4298, opened 2018-10-07
Feature: DATA-3549  [Starfinder] Android Upgrade Slot

LegacyKing:


BagelOrb:
For anybody randomly ending up here like me:

You can find it at 
Feats and Abilities > Character > Anderoid Upgrade Slot (right panel)

PR #4309, opened 2018-10-14
NEWSOURCE-330  Ultimate Wilderness

LegacyKing:


Zaister:
The source should also be remove from PCGen/pcgen-newsources to avoid duplication errors on load.

PR #4311, opened 2018-10-15
Windows installer fixes

javydreamercsw:
Here are my changes to create installer on Java 10 with pcgenRelease. No need to upgrade Gradle to 5.X.

Also it bundles Java 11.

Installer seems to be half the size since Java 11 has no binaries for 32 bits so all that it's dropped.


LegacyKing:
LGTM

PR #4326, opened 2018-10-20
Mend mending stats

BagelOrb:
`Mending` heals 1d4, rather than 5d6.
`Make whole` heals 5d6, perhaps this got confused somehow.

See Starfinder core cule book p. 364
or https://starfinder.dragonlash.com/spells/mending/
or https://www.d20pfsrd.com/magic/all-spells/m/mending/


BagelOrb:
Oh and line 140, because the stats for `Make whole` were also wrong.

PR #4328, opened 2018-10-22
removed ring from MOD

net-diver:
Removed ring from MOD as it only appears in this book

http://aonprd.com/MagicRingsDisplay.aspx?FinalName=Ring%20of%20Delayed%20Doom


grimreaper:
@eslafgh what should the correct cost be?

eslafgh:
> @eslafgh what should the correct cost be?

As per the PRD (http://legacy.aonprd.com/advancedPlayersGuide/magicItems/rings.html#ring-of-delayed-doom) the cost should be 45000 in `apg_equip_magic_items.lst`.

As I mentioned, the cost displayed in pcgen is 0 because the item has no charges set, so in addition to changing `COST:5000` to `COST:45000` in `apg_equip_magic_items.lst` the equipment modifier needs to be changed to give the item a full set of charges i.e. change `EQMOD:CHARGED_ITEM_9` to `EQMOD:CHARGED_ITEM_9|CHARGES[9]`.

grimreaper:
> the equipment modifier needs to be changed to give the item a full set of charges i.e. change EQMOD:CHARGED_ITEM_9 to EQMOD:CHARGED_ITEM_9|CHARGES[9].

I don't see this, but changing the cost is an easy fix.

grimreaper:
oh, should `EQMOD:Special Ability ~ Charged Item / 9 Maximum` be replaced with `EQMOD:CHARGED_ITEM_9|CHARGES[9]` ?

eslafgh:
> oh, should `EQMOD:Special Ability ~ Charged Item / 9 Maximum` be replaced with `EQMOD:CHARGED_ITEM_9|CHARGES[9]` ?

That is correct. I run 6.06.01 and tested it on that version - someone changed the name of the EQMOD in 6.07.06.

PR #4329, opened 2018-10-23
Use the final GA URL

javydreamercsw:
This is the final GA URL for OpenJDK 11

### DIFF
--- a/code/gradle/release.gradle
+++ b/code/gradle/release.gradle
@@ -184,13 +184,13 @@ launch4j {
 task downloadJRE doLast {
     def major = 11
     def micro = 0
-    def update = 0
+    def update = 1
     def extension = 'zip'
     def jre64Dir = new File("${projectDir}/jre/jre_64")
     if(!jre64Dir.exists()){
         //Only download Windows libraries. Need to add more OS if we go multi OS at any time.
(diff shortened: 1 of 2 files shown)


code/gradle/release.gradle
```diff
@@ -184,13 +184,13 @@ launch4j {
...
-    def update = 0
+    def update = 1
     def extension = 'zip'
     def jre64Dir = new File("${projectDir}/jre/jre_64")
     if(!jre64Dir.exists()){
         //Only download Windows libraries. Need to add more OS if we go multi OS at any time.
-        def fileName = "openjdk-${major}_windows-x64_bin"
-        def jre64URL = "https://download.java.net/java/ga/jdk${major}/${fileName}.${extension}"
+        def fileName = "openjdk-${major}.${micro}.${update}_windows-x64_bin"
+        def jre64URL = "https://download.java.net/java/GA/jdk${major}/13/GPL/${fileName}.${extension}"
```
karianna: That 13 in the URL is that correct?

javydreamercsw: Yes. I know it looks weird but that's th url for that GA version. Made no sense to me.

Sadly it seems it might change with each version.

PR #4398, opened 2018-11-17
Cleanup and prepare various classes in preparation for TempBonus work

thpr:
NOTE: This should be a non-logic changing (non-functional) change and should be reviewed as such

### DIFF
--- a/code/src/java/pcgen/gui2/facade/TempBonusHelper.java
+++ b/code/src/java/pcgen/gui2/facade/TempBonusHelper.java
@@ -159,20 +159,15 @@ public class TempBonusHelper
 		{
 			return possibleEquipment;
 		}
-		boolean found = false;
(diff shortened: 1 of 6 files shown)


code/src/java/pcgen/gui2/facade/TempBonusHelper.java
```diff
@@ -159,20 +159,15 @@ public void commit()
 		{
 			return possibleEquipment;
 		}
-		boolean found = false;
 		theCharacter.setCalcEquipmentList(theCharacter.getUseTempMods());
-		for (Equipment aEq : charDisplay.getEquipmentSet())
+		FINDEQ: for (Equipment aEq : charDisplay.getEquipmentSet())
```
javydreamercsw: Nit: Not a big fan of labels for control. Maybe I'm missing something but the previous approach was equivalent and easier to follow.

PR #4405, opened 2018-11-18
Add PCSIZE and BASESIZE code controls

thpr:


code/src/java/pcgen/cdom/facet/analysis/LoadFacet.java
```diff
@@ -131,7 +131,7 @@ public Float getMaxLoad(CharID id, double mult)
 	 */
 	private double getLoadMultForSize(CharID id)
 	{
-		SizeAdjustment sadj = sizeFacet.get(id);
+		SizeAdjustment sadj = pcFacet.getPC(id).getSizeAdjustment();
```
javydreamercsw: Is there a guarantee of getPC not returning null? Otherwise this is a potential NPE.

I would suggest adding a check.

thpr: Guaranteed non null

grimreaper: Not worth a discussion here, but might be nice to pepper the source with `@Nullable`.

PR #4406, opened 2018-11-18
As much as possible, Isolate that Gender is an Enumeration

thpr:
Prepare for moving Gender to Data


thpr:
For me it depends on two things: (1) How permanent is the code and (2) How modular is the code?  In general most of the older code is not modular and thus very hard to isolate tests, and the code being used in transition is not permanent.  So I don't hold up items like this to get tests.

javydreamercsw:
Sounds fine. Just fix the conflicts then.

PR #4407, opened 2018-11-18
Isolate Handed to minimize awareness it's an Enumeration

thpr:
Prepare to place Handed in the data

### DIFF
--- a/code/src/java/pcgen/output/channel/compat/HandedCompat.java
+++ b/code/src/java/pcgen/output/channel/compat/HandedCompat.java
@@ -4,14 +4,33 @@ import java.util.Arrays;
 
 import pcgen.cdom.enumeration.Handed;
 
-public class HandedAdapter
+/**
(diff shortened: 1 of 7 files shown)


code/src/java/pcgen/output/channel/compat/HandedCompat.java
```diff
@@ -0,0 +1,56 @@
...
+
+	/**
+	 * Retrieve a Handed object to match the name ({@link #name()}) or localized name
+	 * (output by {@link #toString()}). The localized lookup is kept for legacy purpose
+	 * when the localized name was saved in the character files (instead of the
+	 * {@link #name()}).
+	 * 
+	 * @param name
+	 *            The localized display name of the Handed.
+	 * @return The matching Handed.
```
javydreamercsw: Looks like it could be null if nothing works. Maybe update the JavaDoc.

thpr:
To be replaced

PR #4417, opened 2018-11-18
Remove LanguageFacade

thpr:
Prepare for more consistent communication method between new formula system & UI


karianna:
@thpr Some codacy failures.

PR #4431, opened 2018-11-20
Add new format - DIE

LegacyKing:


LegacyKing:
@thpr it appears that DIE is not a valid DEFAULT entry

thpr:
Yea, it doesn't have a formatmanager today, I will need to build one.  This will need to wait for a bit

PR #4433, opened 2018-11-20
Formula System - Set up ArmorType and a few misc tags to examine best approach of loca…

LegacyKing:
…l to output.


thpr:
I'm not sure this is really appropriate for master at this point.  Shouldn't specific implementations be in branches appropriate to that specific "conversion task"?

PR #4434, opened 2018-11-20
Formula System - Basic Dynamic Scope of Movement

LegacyKing:


thpr:
I'm not sure this is really appropriate for master at this point.  Shouldn't specific implementations be in branches appropriate to that specific "conversion task"?

PR #4448, opened 2018-11-21
[Icons] migrate away from deprecated code

grimreaper:
Also switch hashmap to EnumMap


thpr:
Unfortunately, this breaks our full tests, with the Icon test failing.  So either this needs to be backed out, or the issues need to be corrected.

karianna:
Hmm, Travis doesn't run our full tests?

thpr:
No, Travis runs the quicker tests as a litmus test.  The full tests in Travis would take >1 hour due to the CPU that is allocated.  We have relied in the past on having Jenkins doing the full build.

PR #4449, opened 2018-11-21
[PreferencesPluginsPanel] remove option to disable plugins

grimreaper:
We never test in that configuration, and there isn't any reason to do
so. Plugins are primarily an internal method of code organization.


karianna:
Waiting for @thpr to approve this.

thpr:
Not sure this is entirely true that plugins are only organization.  This is true for the LST plugins, but for the Encounter plugins, etc. they may be disabled.  I'm ambivalent on that particular topic, but some folks may use this capability... is there a specific reason to remove it (is it headed somewhere else?)

PR #4453, opened 2018-11-21
Fix bug: DATA-3644 - Bonuses of the same type should not stack

LegacyKing:


MeChANiClySM:
Still seeing this as a bug; using Starfinder RPG Core Supplements.

Leve 5 Operative with Ghost
Acrobatics +17 = +5[RANK] +6[Acrobatics] +2[TYPE=INSIGHT|Default] +4[STAT] 

The +6 in Acrobatics should be split into +3 Acrobatics and +3 Insight and the default +2 Insight bonus from Operative's Edge should be removed since the Insight bonuses don't stack.

PR #4472, opened 2018-11-26
[itest] Convert much of itest from junit 3 to junit 4

grimreaper:


karianna:
@grimreaper Codacy fixes please.

grimreaper:
these errors are not really new and not introduced by by this diff, thus safe to ignore.

LegacyKing:
I think @thpr is concerned about compatibility with Eclipse IDE

karianna:
> these errors are not really new and not introduced by by this diff, thus safe to ignore.

Last Boy/Girl Scout and all that ;-).

karianna:
JUnit 4 is fine in Eclipse - it's the move to JUnit 5 that I'm waiting on confirmation from Tom

code/src/itest/pcgen/io/testsupport/AbstractGlobalTargetedSaveRestoreTest.java
```diff
@@ -45,6 +48,10 @@
 import plugin.lsttokens.testsupport.BuildUtilities;
 import plugin.lsttokens.testsupport.TokenRegistration;
 
+import org.junit.Assert;
+import org.junit.Test;
+import org.springframework.lang.Nullable;
+
```
thpr: I"m concerned this import order is not consistent with the rest of our code and will cause churn.  Where did we settle previously on import order?  Have we lost that to the limits of Slack?

grimreaper: I think we did, and I possibly lost it for unrelated reasons. Ideally we'll encode this in checkstyle, otherwise we'll continue to have this conversation multiple tims

(this may have also come from a manual merge conflict)

PR #4484, opened 2018-11-27
[gui2] Remove unused code from UIPropertyContext

grimreaper:


code/src/java/pcgen/gui2/util/AbstractRadioListMenu.java
```diff
@@ -33,10 +33,10 @@
 {
 
 	private final ButtonGroup group = new ButtonGroup();
-	private Map<E, RadioMenuItem> menuMap = new HashMap<>();
-	private E selectedItem = null;
+	private final Map<E, RadioMenuItem> menuMap = new HashMap<>();
+	private final E selectedItem = null;
```
thpr: This is a red flag that something is wrong... we need a broader review of this PR

grimreaper: I think you're right. There is a bug here which needs to be fixed separately.

PR #4493, opened 2018-11-28
Scope is optional, so need a null check before .equals

thpr:


grimreaper:
LGTM but we really should change the API to be a proper `Optional`

PR #4494, opened 2018-11-28
[gui2] Make use of selectedItem in AbstractRadioListMenu

grimreaper:
This was noticed, but not broken in #4484. Previously, selectedItem was
always null: now it changes. The impact is not clear to me (I test a few
usages of it and found no difference) but it seems this is what was
intended.


karianna:
I wonder if you passed E as Opposed to Object into the RadioMeniuItem whether this would make things clearer...

grimreaper:
You mean replace 
```
		private final Object item;

		private RadioMenuItem(Object item, boolean selected, ItemListener listener)
...
```

with
```
		private final E item;

		private RadioMenuItem(E item, boolean selected, ItemListener listener)
...
```
?

grimreaper:
modified. IDE is fine. Will let travis be the compiler.

PR #4495, opened 2018-11-28
Reduce use of System.arrayCopy

thpr:
e.g. leverage Arrays.copyOf


thpr:
Anyone - thoughts?  Is there an underlying out of memory issue or something with Travis?  The errors make no sense given the changes, and I can't reproduce locally.

PR #4499, opened 2018-11-28
Add common delegate/helper object to implement VarHolder interface

thpr:


karianna:
@thpr Merge conflicts.

PR #4510, opened 2018-11-29
Eliminate Race Facade

thpr:


code/src/java/pcgen/core/Deity.java
```diff
@@ -35,15 +34,4 @@
 		wpl.setName("*Domains");
 		DOMAINLIST = CDOMDirectSingleRef.getRef(wpl);
 	}
-
```
LegacyKing: Cross-pollution?  Looks like you're pulling Deity to a Race Facade PR

PR #4516, opened 2018-11-30
Share isUnselected in CDOMObject behavior

thpr:


code/src/java/pcgen/cdom/base/CDOMObject.java
```diff
@@ -1340,4 +1340,15 @@ public PCGenScoped getLocalChild(String childType, String childName)
...
+
+	/**
+	 * Indicates if this is the "UNSELECTED" item an object type for the loaded GameMode.
+	 * 
+	 * @return true if this is the "Unselected" item; false otherwise
+	 */
+	public final boolean isUnselected()
+	{
+		return getSafeListFor(ListKey.GROUP).stream()
+			.filter(s -> "Unselected".equalsIgnoreCase(s)).findFirst().isPresent();
```
grimreaper: nit: chop down. I wonder if we could automate that check
```SUGGESTION
			.filter(s -> "Unselected".equalsIgnoreCase(s))
			.findFirst()
			.isPresent();
```

PR #4536, opened 2018-12-02
Standardize new Code Control behaviors

thpr:
Avoids changes to SourceFileLoader for each Code Control in the future


thpr:
This is covered by the full integration tests

PR #4540, opened 2018-12-02
[6.8] CODE-3302 - attach variables to global PC scope instead of truly global scope

thpr:


code/src/java/pcgen/cdom/grouping/GroupingScopeFilter.java
```diff
@@ -79,7 +79,8 @@ public String getInstructions()
 	@Override
 	public void process(PCGenScoped owner, Consumer<PCGenScoped> consumer)
 	{
-		if (owner.getLocalScopeName().equalsIgnoreCase(scopeName))
+		String localScopeName = owner.getLocalScopeName();
```
grimreaper: Is it legal for localScopeName to ever be null here? I'd rather have it warn than ignored.

thpr: It is legal.  Today, one could have MODIFYOTHER:RACE|BASESIZE=M|...

Race has no localScopeName at this time.

Please note this is dependent upon items in the formula library and the interfaces there.  Given your comments earlier  about Optional, I have working branches back there to change the interface to use Optional not null, and thus this will eventually be changed to looking for optionals to make the intent of the code clearer.

PR #4541, opened 2018-12-02
CODE-3302 Each PC MUST have it's own FormulaManager

thpr:


LegacyKing:
Perhaps future companion support?

PR #4548, opened 2018-12-02
Eliminate need for ComparableComparator

grimreaper:


code/src/java/pcgen/core/GameMode.java
```diff
@@ -1535,7 +1537,7 @@ public void addPointBuyStatCost(final PointBuyCost pbc)
 		{
 			// Sort NUMERICALLY, not alphabetically!
 			// CONSIDER Huh? The natural order of Integer IS numerically... - thpr 10/20/06
-			pointBuyStatCosts = new TreeMap<>(new ComparableComparator<>());
+			pointBuyStatCosts = new TreeMap<>(Comparator.comparing(Function.identity()));
```
thpr: Based on the CONSIDER I put in many years ago, I'm wondering if this even needs a Comparator passed into TreeMap.  By comparing against the natural order anyway, it's just injecting an unnecessary object into the TreeMap, right?

grimreaper: I think you are correct: we're comparing against Function.identity anyways. Let me see if injecting the object is avoidable.

PR #4549, opened 2018-12-02
[*] Simplify if/else chains

grimreaper:


code/src/java/pcgen/io/exporttoken/WeaponToken.java
```diff
@@ -2937,7 +2939,7 @@ private static String getUnarmedAttackString(PlayerCharacter pc, int bonus, int
 	 */
 	private static boolean appendSeperator(Equipment eq)
 	{
-		return eq.isType("Natural") && (eq.isType("Both") || eq.isType("Melee") || eq.isType("Ranged"));
+		return eq.isType("Natural") && (Stream.of("Both", "Melee", "Ranged").anyMatch(eq::isType));
```
thpr: Thinking out loud: I'm not sure this is adding sufficient value.  I'm not opposed to the stream concept, I'm opposed to changing a list of 3 items to a list of 3 items.  There seems to be magic here not unlike a magical number within the code... Why did someone use 25?  Well, why did someone use just those types?  Especially true since the same 3 items appear as part of the group below naming them as 3 of 4 of "standard" types??

code/src/java/pcgen/io/exporttoken/WeaponToken.java
```diff
@@ -1947,9 +1948,10 @@ else if (hitMode == HITMODE_OHHIT)
 		 but for some reason they can (e.g. Monkey Grip) then check
 		 for TOHIT modifiers */
 
-		if (eq.getLocation() == EquipmentLocation.EQUIPPED_PRIMARY
-			|| eq.getLocation() == EquipmentLocation.EQUIPPED_SECONDARY
-			|| eq.getLocation() == EquipmentLocation.EQUIPPED_TWO_HANDS)
+		if (Stream.of(EquipmentLocation.EQUIPPED_PRIMARY,
+				EquipmentLocation.EQUIPPED_SECONDARY,
+				EquipmentLocation.EQUIPPED_TWO_HANDS)
+		          .anyMatch(equipmentLocation -> eq.getLocation() == equipmentLocation))
```
thpr: What do these equipped locations share?  See below for having this semantically useful

grimreaper:
alright. makes sense

PR #4557, opened 2018-12-02
Remove dead code in AbstractReferenceContext

grimreaper:


thpr:
There are plans for many of these, so I'd rather leave them for now

PR #4560, opened 2018-12-02
[*] Remove need for SortKeyComparator

grimreaper:


thpr:
Is it needed elsewhere or can a file be deleted?

Also, this is reordering imports, so let's close that issue before we merge this

PR #4564, opened 2018-12-02
WeightedCollection: annotate overrides like superclass

grimreaper:


code/src/java/pcgen/io/exporttoken/WeaponToken.java
```diff
@@ -21,7 +21,6 @@
 package pcgen.io.exporttoken;
 
 import java.util.ArrayList;
-import java.util.Iterator;
```
thpr: This causes breakage

PR #4566, opened 2018-12-02
[*] Use Arrays.setAll

grimreaper:


thpr:
Approved once we figure out import ordering

PR #4604, opened 2018-12-06
[CR] Fix Equipment to Sort Base Items at Bottom of List

Llisandur:
Remove INFO:OutputName, since it won't do anything properly for custom weapons.


LegacyKing:
Please update the broken tests.

PR #4611, opened 2018-12-07
[do not merge] demonstrate lots of unused code

grimreaper:
This is almost all mechnically produced deletions that show we have a lot of 'unused' logic for some of our tokens. I need to go through this, but wanted to start a conversation and/or get input as to what's happening here.

### DIFF
--- a/code/src/java/pcgen/core/GameMode.java
+++ b/code/src/java/pcgen/core/GameMode.java
@@ -18,7 +18,6 @@
 package pcgen.core;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -28,8 +27,6 @@ import java.util.Set;
 import java.util.SortedMap;
 import java.util.TreeMap;
 import java.util.TreeSet;
-
-import org.apache.commons.lang3.StringUtils;
 
 import pcgen.base.util.HashMapToList;
 import pcgen.cdom.base.CDOMReference;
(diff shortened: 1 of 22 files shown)


karianna:
How are you determining that they're not called?

thpr:
Kar - I looked at one of them, and the variable simply isn't used anymore in PCGen.  This is apparently common as GameMode never got a scrub like the main tokens did.  Changed less often / built less often meant it wasn't the top of my cleanup list.

The spirit here is right, but the implementation is definitely not the way to do this - bulk change is not the answer here, nor is leaving tokens that just "return true" as their only action.  We also need to keep the rebuild to streams to another PR, that is separate from the feature removal.  One thing at a time, especially when we are making changes to data that we need to clearly document for anyone with a home-brew game system.

Each variable should be removed from GameMode, with the resulting useless token(s) removed from PCGen as well, the pluginbuild.xml file updated so that token no longer is built into a JAR, and the data changed so it no longer calls the removed token (and thus won't produce errors).  One GameMode variable per PR, with hopefully one token, but that isn't always possible.  The data should be in a separate commit for the PR so it's easier for the data team to review.  See PR 4534 for an example of data changing in parallel to code.

grimreaper:
> How are you determining that they're not called?
Tests pass, and these were mechnically generated: the compiler proved that they are not used.

> The spirit here is right, but the implementation is definitely not the way to do this
As the title says, I never intended for these to be merged: I just wanted confirmation the spirit is correct and an idea of how to complete thes.

Thanks!

PR #4553, opened 2018-12-02
Simplify AbstractReferenceContext.resolveReferences

grimreaper:

PR #4614, opened 2018-12-08
Revert "Simplify AbstractReferenceContext.resolveReferences"

thpr:
Reverts PCGen/pcgen#4553

This patch causes build failures - it is not logically correct


code/src/java/pcgen/rules/context/AbstractReferenceContext.java
```diff
@@ -147,8 +148,12 @@ public void initialize()
 
 	public boolean validate(UnconstructedValidator validator)
 	{
-		return getAllManufacturers().stream()
-                            .allMatch(ref -> ref.validate(validator));
```
thpr: This is NOT logically identical because the stream does NOT guarantee it evaluates all elements.  We MUST do that to avoid crashes in the core

PR #4625, opened 2018-12-08
[DO NOT MERGE] DieSizes: remove MIN= and MAX= token from system/

grimreaper:


thpr:
Why did you take out [DO NOT MERGE]

This has found a bug, not something to be removed, right?

grimreaper:
> Why did you take out [DO NOT MERGE]

I meant to close these until I could figure out the right thing to do here.

PR #4627, opened 2018-12-08
remove CURRENCYUNIT from 'system' data

grimreaper:


thpr:
I'm skeptical of this one - I see CURRENCYUNITABBREV used in real code.  Also, this has no corresponding code PR, which is why it needed the detailed review

system/gameModes/35e/miscinfo.lst
```diff
@@ -27,8 +27,6 @@ BONUSSTACKS:Defense.Dodge.Circumstance.NotRanged.NotFlatFooted
 #
 #misc. currency specific items
 #
-CURRENCYUNIT:Gold
-CURRENCYUNITABBREV:gp
```
thpr: Abbrev is used in real code

grimreaper: ah, I see my problem. I intended to only remove `CURRENCYUNIT`, not `CURRENCYUNITABBREV`. Will fix.

PR #4628, opened 2018-12-08
[DEP 4660] Remove 'SHOWCLASSDEFENSE' from code

grimreaper:


code/src/java/plugin/lsttokens/gamemode/ShowclassdefenseToken.java
```diff
@@ -1,26 +0,0 @@
-package plugin.lsttokens.gamemode;
```
thpr: Removes token without removing the entry from pluginbuild.xml

PR #4630, opened 2018-12-08
[DO NOT MERGE] DieSizes: remove documentation of MIN and MAX variables

grimreaper:


thpr:
This sets off red flags because %up and %Hup are both used in data

grimreaper:
> This sets off red flags because %up and %Hup are both used in data

ah! I removed the wrong documentation. The code is correctly removable though.

PR #4636, opened 2018-12-08
Make method static where possible

grimreaper:


code/src/java/pcgen/core/GameMode.java
```diff
@@ -18,7 +18,6 @@
 package pcgen.core;
 
 import java.util.ArrayList;
-import java.util.Arrays;
```
thpr: Umm, this is sketchy in this PR?

code/src/java/pcgen/core/GameMode.java
```diff
@@ -2307,7 +2306,7 @@ public void resolveInto(AbstractReferenceContext referenceContext)
 		}
 	}
 
-	private AbstractReferenceContext getRefContext()
+	private static AbstractReferenceContext getRefContext()
```
thpr: FWIW, the whole "method should be static because it doesn't use a field" thing is really one of those things where tools say "it can be" and I say "so what" - just because it can be doesn't mean it should be... since this is not something we've had in code standards this isn't really high on my list of cleanup I think adds value.

PR #4651, opened 2018-12-08
Bug Fix - needs to handle default

thpr:


grimreaper:
accepting, but ideally we should add a test for this case.

PR #4653, opened 2018-12-08
Remove ClassFacade

thpr:


thpr:
Note: Codacy is likely to go crazy on this PR because so many undocumented items are changed from ClassFacade to Class.  The key is whether the new items are documented correctly, they are.  Others may get sunset with channels anyway, so I am limited in my desire to write new docs I will delete shortly

PR #4656, opened 2018-12-08
unbreak character creation

grimreaper:
This was broken in 832c919 ("Cleanup - mostly eliminated unnecessary
casts") which removed a hidden null check. While here sprinkle a bit of
assertions around to avoid this type of issue in the future.


code/src/java/pcgen/gui2/facade/CharacterLevelsFacadeImpl.java
```diff
@@ -386,28 +392,35 @@ public boolean isClassSkillForMaxRanks(CharacterLevelFacade level, Skill skill)
...
+		Objects.requireNonNull(skill);
 		return SkillModifier.modifier(skill, theCharacter);
 	}
 
 	@Override
-	public float getSkillRanks(CharacterLevelFacade level, Skill skill)
+	public float getSkillRanks(@Nullable CharacterLevelFacade level, @Nullable Skill skill)
 	{
-		// TODO Ranks aren't stored by level - have compromised by returning the total. Further discussion needed 
+		if (skill == null)
```
grimreaper: This is the bugfix.

PR #4670, opened 2018-12-09
added class skills

net-diver:
added class skills gained from being a poppet familiar

https://www.d20pfsrd.com/feats/general-feats/poppet-familiar/


LegacyKing:
I'm not exactly sure what this is meant to accomplish. You list a few skills (groups), and add nothing to them. If the intention is to make them 'race' class skills or otherwise known as Monster Class Skills, then adding 0 ranks to them in a Kit will not accomplish your goal. 

MCSKILL tag on the race is required to achieve that result.

Unless the code has changed drastically, this is just a meaningless change that adds lines of code to be processed but accomplishes nothing.

PR #4671, opened 2018-12-09
removed mindless for familiars added wis score

net-diver:
Removed mindless trait as poppet familiars gain an intelligence score and changed WIS score to equal INT as part of being a poppet familiar

https://www.d20pfsrd.com/feats/general-feats/poppet-familiar/


net-diver:
The description for Mindless is a "A creature with no Intelligence score. "

By becoming a familiar the poppet essentially becomes an awakened construct as it gaining it an INT score and the ability to speak.

PR #4672, opened 2018-12-09
[core] remove ShowMessageConsoleObserver

grimreaper:
Remove ShowMessageConsoleObserver. This class is unused, and implements
the Observer method which is deprecated as of Java 9. Instead of
migrating, remove.


LegacyKing:
@grimreaper  Please add reviewers so they get notified this is ready. :) Either a specific person or you can do a generic team blast using the Code Content Reviewers group I set up.

LegacyKing:
LGTM for what its worth.

PR #4673, opened 2018-12-09
Remove unused code in ResourceChecker and related

grimreaper:


code/src/java/pcgen/util/SkinLFResourceChecker.java
```diff
@@ -46,31 +37,16 @@ private SkinLFResourceChecker()
...
-
-		return "";
-	}
-
 	private static void checkResource()
 	{
 		if (!ResourceChecker.hasResource("com.l2fprod.gui.plaf.skin.SkinLookAndFeel", "skinlf.jar", resourceBuffer))
 		{
-			++missingResourceCount;
+			missingResourceCount = true;
```
LegacyKing: Terminology matters to me. Why are we calling a Boolean a "Count"?  Count means a value of an Integer or at the very least a number. I'm happy you're reusing and refactoring the code, but can we rename the var to match its new purpose?  Above the function call is hasMissingResources, why not name the Boolean hasMissingResource which matches name to intent.

thpr:
Agree with @LegacyKing  - the variable name should be updated

grimreaper:
resolved.

PR #4684, opened 2018-12-12
CharacterDisplay: cleanup and modernize

grimreaper:
- remove several dead functions
- use streams
- add final

### DIFF
--- a/code/src/java/pcgen/core/display/CharacterDisplay.java
+++ b/code/src/java/pcgen/core/display/CharacterDisplay.java
@@ -29,6 +29,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 import pcgen.base.formula.Formula;
 import pcgen.base.util.NamedValue;
@@ -79,7 +80,6 @@ import pcgen.cdom.facet.analysis.AgeSetFacet;
 import pcgen.cdom.facet.analysis.ArmorClassFacet;
 import pcgen.cdom.facet.analysis.BaseMovementFacet;
 import pcgen.cdom.facet.analysis.ChallengeRatingFacet;
-import pcgen.cdom.facet.analysis.ChangeProfFacet;
 import pcgen.cdom.facet.analysis.FavoredClassFacet;
 import pcgen.cdom.facet.analysis.FollowerOptionFacet;
 import pcgen.cdom.facet.analysis.HasAnyFavoredClassFacet;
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/core/display/CharacterDisplay.java
```diff
@@ -480,13 +478,12 @@ public int getTemplateSR(PCTemplate pct, int level, int hitdice)
 
 	private List<PCTemplate> getVisibleToTemplateList(View v)
 	{
-		List<PCTemplate> tl = new ArrayList<>();
 
 		Collection<PCTemplate> treeSet = new TreeSet<>(CDOMObjectUtilities::compareKeys);
-		templateFacet.getSet(id).stream().filter(template -> template.getSafe(ObjectKey.VISIBILITY).isVisibleTo(v))
-			.forEach(treeSet::add);
```
thpr: Wrapping? Shouldn't this be filter/foreach on their own lines?

code/src/java/pcgen/core/display/CharacterDisplay.java
```diff
@@ -1116,12 +1101,12 @@ public double getStatBonusTo(String aType, String aName)
 		return statBonusFacet.getStatBonusTo(id, aType, aName);
 	}
 
-	public String getUDamForRace()
+	String getUDamForRace()
```
thpr: This is a negative change, IMHO.  The method itself can be used publicly, reducing the visibility (1) sends the wrong message because it implies there is some reason it needs package protection and (2) causes someone to have to change it back to public if it is needed publicly.  Semantics matter, and this is one of those areas where I think the automated tools that say "X could be Y" don't understand "Y" is a signaling mechanism and overusing it detracts from the code base.  The general purpose use of methods should be public, private, protected (when a subclass needs it in a template).  Package should be for things where it really is exposing internal behavior to the package for the purposes of keeping separate functions in different classes (e.g. avoiding inner classes)

code/src/java/pcgen/core/display/CharacterDisplay.java
```diff
@@ -1116,12 +1101,12 @@ public double getStatBonusTo(String aType, String aName)
...
 	{
 		return unarmedDamageFacet.getUDamForRace(id);
 	}
 
-	public Set<List<String>> getUnarmedDamage()
+	Set<List<String>> getUnarmedDamage()
```
thpr: Same as above

code/src/java/pcgen/core/display/CharacterDisplay.java
grimreaper: If this was a library I'd agree with you. Outside of library code all methods should be private by default and have minimal visibility outside of that. IMHO it is a language-mistake for package to be the default with no keyword (it should be `private`)

karianna:
@thpr Is this OK to merge now?

grimreaper:
I'm going to kill this for now. It is just general cleanup and not a blocker for my future goals:
- eliminating null
- moving to junit 4 (and eventually 5)
- eliminating dead code

PR #4687, opened 2018-12-14
Core_race_language_migration

evilpixie87:
Moved core race languages to templates


LegacyKing:
Travis stalled. Change is safe.

PR #4690, opened 2018-12-16
DATA-3790 - PcGen marks down my ability mods for more saves under a m…

LegacyKing:
…isc mod rather than ability mod, it also doesn't add gear boost or armor upgrade info to the sheet.

DATA-3787 - Vesk Overplate III cost= 0 credits, starfinder


LegacyKing:
Thanks for the vote. The two hidden spaces causing errors needed to be fixed first though. Any chance we'll get master building and running soon? Waiting on Travis to do the error checking is very slow and painful.

PR #4708, opened 2018-12-22
Remove dead code in StatToken

grimreaper:


grimreaper:
didn't intend to push this one to origin - just the wrapper regen. that said, not going to revert unless y'all don't like it.

LegacyKing:
Usually if we have a break on master we place a freeze until we isolate and fix the break. I'm not @thpr and can't say one way or another, but I think we need to focus on fixing the master before we introduce anymore variables to the master.

PR #4713, opened 2018-12-22
Revert changes in "Core_race_language_migration" and related

grimreaper:
This unbreaks pfinsttest

   This reverts commit 8991a05fd9b6003f894449f0037353e134ca1c5b.
   This reverts commit dc3c2d7f4aea9b34f338ed27694c54a09ad15d97.
   This reverts commit e247a41b698fe7b4691108409bdfb90a760dc5f9.


LegacyKing:
Okay, this is passing for the BUILDS - I don't know WHY, because the LST is reasonable and shouldn't break the tests. Considering this is for the TRAVIS issue only, I don't believe this is the root cause for the Master breaking with the gradlew run/gradle run stuff.

PR #4723, opened 2018-12-24
Sprinkle Optional in places

grimreaper:
I found some null pointer exceptions in skill editing code. Sprinkle
Optional around to try and get the error to happen a bit earlier (and
more obviously).

### DIFF
--- a/code/src/java/pcgen/cdom/facet/LevelInfoFacet.java
+++ b/code/src/java/pcgen/cdom/facet/LevelInfoFacet.java
@@ -20,6 +20,7 @@ package pcgen.cdom.facet;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 
 import pcgen.cdom.enumeration.CharID;
 import pcgen.cdom.facet.base.AbstractListFacet;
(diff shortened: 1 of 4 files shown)


karianna:
@thpr - I think you'd better look at this and judge whether that's how you want the Facade to behave.

grimreaper:
FWIW my goal here is to eliminate `null` from everywhere eventually. This particular diff gives me a way to find a bug though, so begin sprinkling it around.

grimreaper:
> This will still error out if the result is empty, it just won't be a NPE
That's the point: it will also error out earlier, and the fix will be to continue to apply `Optional` earlier and earlier in the chain

code/src/java/pcgen/core/display/CharacterDisplay.java
```diff
@@ -1045,7 +1045,7 @@ public int getECL()
 
 	public PCLevelInfo getLevelInfo(int index)
 	{
-		return levelInfoFacet.get(id, index);
+		return levelInfoFacet.get(id, index).get();
```
thpr: This makes no sense.  Some callers of getLevelInfo are gating the result on null, why does the get appear in this method?

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -3269,14 +3269,14 @@ public void setHasMadeKitSelectionForAgeSet(final int index, final boolean arg)
 
 	public PCLevelInfo getLevelInfo(int index)
 	{
-		return levelInfoFacet.get(id, index);
+		return levelInfoFacet.get(id, index).get();
```
thpr: This makes no sense.  Some callers of getLevelInfo are gating the result on null, why does the get appear in this method?

code/src/java/pcgen/cdom/facet/LevelInfoFacet.java
```diff
@@ -58,14 +59,14 @@
 	 * @return The object in this LevelInfoFacet for the Player Character
 	 *         represented by the given CharID and location.
 	 */
-	public PCLevelInfo get(CharID id, int location)
+	public Optional<PCLevelInfo> get(CharID id, int location)
```
thpr: This implementation leaves something to be desired.  It returns an Optional, telegraphing semantically that something is optional, only to have all uses immediately call "get()".  This makes no sense.  If it's required, then the return null below should throw an exception.

Having said that, this change is also not well implemented because some of the gets immediately return and then encounter a null check, so no matter what - this patch is either badly formed or logically incorrect.  Changes are required.

grimreaper:
You're correct on this one. Dropping it on the floor

PR #4726, opened 2018-12-24
rm DISPLAYVARIABLE#[NAME|TEXT]

grimreaper:


grimreaper:
yes. This removes it from code now.

PR #4727, opened 2018-12-24
Remove dead code in skill editing code

grimreaper:


grimreaper:
I have some certainty that this code isn't used, but it may be more correct to use it than remove it.

thpr:
I concur with Eitan on this one - there are other methods the existing UI uses to get the skill ranks/costs.  This looks like code that was built under one point of view, and then it was realized that another process was better and these were not cleaned up.

PR #4733, opened 2018-12-25
Teach Gradle about the application name on macOS

grimreaper:
The wrapper (pcgen.sh) also needs to be updated, but one thing at a time.


grimreaper:
@karianna we should separate the "better ico file" from this change. Other places use the same file (and conceptually it is a different change).

PR #4740, opened 2018-12-25
CODE-2772: Implement a 'remove equipment' button

grimreaper:
Sometimes equipment gets used rather than sold. Instead of having to switch
the sell mode each time, add a new button.

(this PR should be rebased; not squashed)

### DIFF
--- a/code/src/java/pcgen/gui2/tabs/PurchaseInfoTab.java
+++ b/code/src/java/pcgen/gui2/tabs/PurchaseInfoTab.java
@@ -118,6 +118,7 @@ public class PurchaseInfoTab extends FlippingSplitPane implements CharacterInfoT
 	private final JCheckBox autoResizeBox;
 	private final JButton addCustomButton;
 	private final JButton addEquipmentButton;
+	private final JButton sellEquipmentButton;
 	private final JButton removeEquipmentButton;
 	private final InfoPane infoPane;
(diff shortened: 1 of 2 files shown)


code/src/java/pcgen/gui2/tabs/PurchaseInfoTab.java
```diff
@@ -632,14 +621,14 @@ public void install()
...
-	private class RemoveAction extends AbstractAction
+	private final class SellAction extends AbstractAction
 	{
 
 		private final CharacterFacade character;
 
-		public RemoveAction(CharacterFacade character)
+		private SellAction(CharacterFacade character)
 		{
-			super(LanguageBundle.getString("in_ieRemEq")); //$NON-NLS-1$
```
javydreamercsw: Removing the comment seems not on purpose.

grimreaper: semi-intentional. I'd prefer we use `@SuppressWarnings` instead. That said, I'm separating the baseline cleanup from the adding of the button.

karianna:
@grimreaper So the title of this PR should really be "Implement a 'sell equipment' button"?

grimreaper:
@karianna no. We already have "sell equipment". This adds _remove_ which bypasses the current sell mode.

PR #4741, opened 2018-12-25
PCGenAction: remove always false check

grimreaper:
We no longer use mrj, so this is always false.


karianna:
@grimreaper How do you know this is always false?

grimreaper:
`mrj` is a macOS specific property. From wikipedia:
```
Since the transition to Mac OS X, Apple has discontinued MRJ and instead maintains and distributes a port of Oracle's HotSpot Java virtual machine.
```

PR #4743, opened 2018-12-26
Distribution: eliminate 'root' mode

grimreaper:
In an effort to eventually move us to building an uberjar and using
gradle's native runtime scripts, remove the mode that installs files to
the root. Instead all output will always be in `output`

Further, allow build's output to be cached by not depending on 'clean'

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       - os: osx
         osx_image: xcode9
  
-script: ./gradlew clean build installToRoot test compileSlowtest datatest pfinttest allReports buildDist buildNsis pcgenRelease
+script: ./gradlew clean build copyToOutput test compileSlowtest datatest pfinttest allReports buildDist buildNsis pcgenRelease
 # Do not use itest, requires release notes
 
 before_install:
--- a/build.gradle
+++ b/build.gradle
@@ -251,34 +251,8 @@ task copyToOutput(type: Copy, dependsOn: [createExe, copyToLibs, jar,
     rename "(.+)-$version(.+)", '$1$2'
 }
 
-task syncLibsToRoot(type: Sync, dependsOn: [copyToLibs, jar, converterJar]) {
-    outputs.dir("$projectDir/libs")
-    inputs.files(fileTree("$buildDir/libs"))
-
-    from "$buildDir/libs"
-    into "$projectDir/libs"
-    exclude "pcgen-*.jar"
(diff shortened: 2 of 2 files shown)


karianna:
@grimreaper and if you invoke the gradle target to run it runs this correctly?

grimreaper:
yeah. It runs perfectly

javydreamercsw:
Well it seems the installer built correctly. I just suggest testing the actual installer to see if the exe works. Also the metainf in the jar shouldn't list the dependencies in the lib folder.

karianna:
@javydreamercsw Are you able to create a PR with @grimreaper to also adjust the installer?

javydreamercsw:
Is he done with his changes so I can fork his branch and work there?

thpr:
Javier, you can branch off this, but before we touch ANYTHING else in Gradle, we need a stable build in Gradle.  We don't have that today.  I can't get a build to actually complete anymore.  It's hanging on various tests (both in Jenkins and locally).  Travis here can't build without failing 33%+ of the time.  Something over the last 2 or so months has put our build process on a very unstable platform, and we need to get that fixed... quickly.

javydreamercsw:
I'll wait then.

karianna:
@javydreamercsw We're back to a stable build, so please go ahead.

javydreamercsw:
I'll get to it ASAP.

karianna:
@javydreamercsw Are you happy this can land now?

javydreamercsw:
I was unable to make the installer works with this changes. The installer builds, etc. but just doesn't work.

Was unable to make it work and not much time to debug.

karianna:
> I was unable to make the installer works with these changes. The installer builds, etc. but just doesn't work.
> 
> Was unable to make it work and not much time to debug.

OK thanks will need to keep this open for now then

LegacyKing:
@grimreaper @karianna @thpr  What are we waiting on for this?

thpr:
Addressing @javydreamercsw 's concern, I think

javydreamercsw:
I was unable to make installer work with the changes. Something about the Uber jar doesn't work well with the installer and was unable to figure out what.

javydreamercsw:
Finally was able to test after all the changes and it seems to work well after my tests.
This has my blessing to go forward.

grimreaper:
Thanks for testing. No worries about the time.

PR #4744, opened 2018-12-26
Distribution: use uberjar

grimreaper:
This allows us to avoid needing to copy libraries around, installing
them, and so on. Will eventually allow us to rely on Gradle's startup
scripts rather than our own.

Depends on #4743


karianna:
@grimreaper this contains the same commits as #4743 - so I think that one can be closed...

karianna:
Same question about this working out of the box applies

grimreaper:
This was supposed to be a followup, but I'll double check.

karianna:
@grimreaper Does this need to be tested alongside @javydreamercsw installer?

grimreaper:
> @grimreaper Does this need to be tested alongside @javydreamercsw installer?

probably.

javydreamercsw:
Yes. In theory it should work.

The manifest needs to be empry of a path for example. Also the installer copies jars etc. Those might fail if those are not there, etc.

karianna:
@javydreamercsw Let us know when you have some time to do the installer checking on this and #4743

javydreamercsw:
See comment on #4743

karianna:
> See comment on #4743

@javydreamercsw #4743 has now landed - safe to land this?

grimreaper:
Confirmed on slack. Merging.

PR #4754, opened 2018-12-26
Convert several non-abstract classes to junit 4

grimreaper:


karianna:
@grimreaper forgot some imports I think

```
symbol:   method assertEquals(String,String)
  location: class LookupFunctionTest
/home/travis/build/PCGen/pcgen/code/src/utest/plugin/function/LookupFunctionTest.java:770: error: cannot find symbol
		assertEquals(formula, rv.toString());
```

code/src/utest/pcgen/cdom/facet/fact/XPFacetTest.java
```diff
@@ -22,25 +22,26 @@
...
 	private CharID id;
 	private CharID altid;
 	private XPFacet facet;
 	private int adjustment;
 	private Map<Integer, Integer> minXP;
 
-	@Override
-	protected void setUp() throws Exception
+	@Before
+	public void setUp() throws Exception
```
thpr: This has knock-on side effects that break the build

PR #4758, opened 2018-12-27
PreferencesDialog: remove dead code: there is no work done in resetOptionValues

grimreaper:


grimreaper:
Ah. Forget this: LookAndFeelPanel implments this, but I did not notice this before.

PR #4763, opened 2018-12-28
Use Optional where possible in SizeUtilities.getDefaultSizeAdjustment

grimreaper:


thpr:
This should not be implemented as such.  It is optionalizing something that is not optional.

Please see: 
public boolean process(LoadContext context, Collection<? extends SizeAdjustment> obj)
in plugin.lsttokens.sizeadjustment.IsdefaultsizeToken

This prevents a load from completing unless we KNOW there is a default size.

code/src/java/pcgen/core/analysis/SizeUtilities.java
```diff
@@ -31,17 +33,14 @@ private SizeUtilities()
...
-			{
-				return s;
-			}
-		}
+		return Globals.getContext()
+		              .getReferenceContext()
+		              .getConstructedCDOMObjects(SizeAdjustment.class)
+		              .stream()
+		              .filter(s -> s.getSafe(ObjectKey.IS_DEFAULT_SIZE))
+		              .findFirst();
```
thpr: This should call findFirst() then get() since failure is not an option per my previous comment :)

grimreaper:
This was a debugging technique. Closing

PR #4765, opened 2018-12-28
CODE-3200: add Locale.ENGLISH to "everywhere"

grimreaper:
In an effort to help with with locale issues: assume _everywhere_ is
English.  If this solves the issue, great. If not, it is still-a-prereq
to solving the issue. Additionally, being explicit is good. :)

### DIFF
--- a/code/src/java/pcgen/cdom/base/BasicClassIdentity.java
+++ b/code/src/java/pcgen/cdom/base/BasicClassIdentity.java
@@ -17,6 +17,7 @@
  */
 package pcgen.cdom.base;
 
+import java.util.Locale;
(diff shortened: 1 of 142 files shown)


karianna:
@grimreaper I suspect this may badly hamper internationlisation down the line.

@Zaister - thoughts?

grimreaper:
@karianna I'll split this up for better review, but this almost only affects tokens and computery things. It doesn't affect user reading code.  It may even help i18n by allowing turkish users to use pcgen (see CODE ticket for details)

grimreaper:
I'm going to split this up to about 10 files per review to make it easier to review.

PR #4766, opened 2018-12-28
SettingsHander: remove showImagePreview

grimreaper:
This setting is never used, so just remove it.


karianna:
@grimreaper How do you know it's never used?

grimreaper:
grep.

thpr:
This is dead code.

PR #4774, opened 2018-12-29
CODE-3200: add Locale.ENGLISH to computer-read values (id: 7:18244)

grimreaper:


code/src/java/plugin/pretokens/test/PreArmorTypeTester.java
```diff
@@ -56,7 +57,7 @@ public int passes(final Prerequisite prereq, final CharacterDisplay display, CDO
 				{
 
 					String stripped = desiredType.substring(Constants.SUBSTRING_LENGTH_FIVE);
-					StringTokenizer tok = new StringTokenizer(stripped.toUpperCase(), ".");
+					StringTokenizer tok = new StringTokenizer(stripped.toUpperCase(Locale.ENGLISH), ".");
```
thpr: Types are data controlled
...I think in this case we need to check if caps is reqlly necessary?  Is eq.isType already case insensitive?  Should it be?

code/src/java/plugin/pretokens/test/PreArmorTypeTester.java
```diff
@@ -83,13 +84,13 @@ public int passes(final Prerequisite prereq, final CharacterDisplay display, CDO
 				}
 				else
 				{ //not a TYPE string
-					final String eqName = eq.getName().toUpperCase();
+					final String eqName = eq.getName().toUpperCase(Locale.ENGLISH);
```
thpr: This is certainly data controlled

grimreaper: Are data types type case sensitive? Even though this is data controlled, it is an essentially a variable name and not something that should follow the default locale.

thpr: Not entirely.  The else on line 99/100 is definitely an internal usage that is okay to use Locale.English.  The if on line 93/94 and the else if on line 107/108 are not, and would impact data not written in English.  So in effect, this needs to be properly rewritten to handle that mix of locality (yes, the code is not great today, but to avoid putting other locales in a bad position in the future, let's fix it here)

code/src/java/plugin/pretokens/test/PreEquipTester.java
```diff
@@ -70,7 +71,7 @@ public int passes(final Prerequisite prereq, final PlayerCharacter character, CD
 				}
 				else if (targetEquip.startsWith("TYPE=") || targetEquip.startsWith("TYPE.")) //$NON-NLS-1$ //$NON-NLS-2$
 				{
-					StringTokenizer tok = new StringTokenizer(targetEquip.substring(5).toUpperCase(), ".");
+					StringTokenizer tok = new StringTokenizer(targetEquip.substring(5).toUpperCase(Locale.ENGLISH), ".");
```
thpr: This is also data dependent, as may be items lower down in this file and later in this PR.

grimreaper:
ack. I was unaware of this data dependence. Dropping this PR.

PR #4777, opened 2018-12-29
CODE-3200: add Locale.ENGLISH to computer-read values (id: 1:9275)

grimreaper:


code/src/java/pcgen/cdom/facet/analysis/MovementResultFacet.java
```diff
@@ -371,7 +372,8 @@ public double movement(CharID id, int moveIdx)
 			double moveInFeet = getMovement(moveIdx);
 
 			// First get the MOVEADD bonus
-			moveInFeet += bonusCheckingFacet.getBonus(id, "MOVEADD", "TYPE." + getMovementType(moveIdx).toUpperCase());
+			moveInFeet += bonusCheckingFacet.getBonus(id, "MOVEADD", "TYPE." + getMovementType(moveIdx).toUpperCase(
+					Locale.ENGLISH));
```
thpr: This is an example that makes me nervous.  This string is technically under control of the data, not a built in magical word.  So I think this is a bad assumption

PR #4780, opened 2018-12-29
CODE-3200: add Locale.ENGLISH to computer-read values (id: 6:18815)

grimreaper:


code/src/java/plugin/exporttokens/TextToken.java
```diff
@@ -97,11 +98,11 @@ public String getToken(String tokenSource, PlayerCharacter pc, ExportHandler eh)
 
 		if (action.equalsIgnoreCase("UPPER") || action.equalsIgnoreCase("UPPERCASE"))
 		{
-			retString = retString.toUpperCase();
+			retString = retString.toUpperCase(Locale.ENGLISH);
```
thpr: This is controlling what is in an output sheet - why do we believe it would be English?

grimreaper: you're correct. I had intended for this to be part of my next round of adding explicit default locale calls. Fixed.

PR #4781, opened 2018-12-29
CODE-3200: add Locale.ENGLISH to computer-read values (id: 4:8213)

grimreaper:


code/src/java/pcgen/io/ExportHandler.java
```diff
@@ -853,7 +853,7 @@ else if (tokenCount != 2)
 			{
 				return leftString.equals(rightString.substring(1));
 			}
-			return leftString.toUpperCase().contains(rightString.toUpperCase());
+			return leftString.toUpperCase(Locale.ENGLISH).contains(rightString.toUpperCase(Locale.ENGLISH));
```
thpr: This is fairly powerful - and in code I'm not familiar with... how much has this been tested?

code/src/java/pcgen/io/PCGFile.java
```diff
@@ -167,6 +168,6 @@ public static boolean isPCGenCharacterOrPartyFile(final File file)
 	 */
 	private static String getWindowsSafeFilename(final File file)
 	{
-		return file.getName().toLowerCase();
+		return file.getName().toLowerCase(Locale.ENGLISH);
```
thpr: Isn't this almost certainly dangerous?  File names are user controlled...

grimreaper: fixed. I originally was thinking that system files should use `Locale.English` but re-reading it makes more sense to explicitly use the default locale.

PR #4782, opened 2018-12-29
CODE-3200: add Locale.ENGLISH to computer-read values (id: 5:19271)

grimreaper:


code/src/java/plugin/converter/ChooseConvertPlugin.java
```diff
@@ -93,7 +94,7 @@ private static void processSpellList(TokenProcessEvent tpe)
 		String decision = tpe.getDecider()
 			.getConversionInput("Please provide class spell list which " + tpe.getObjectName() + " modifies").trim();
 		String stat = spelllistAnswered.computeIfAbsent(decision, d -> tpe.getDecider()
-			.getConversionInput("Please provide SPELLSTAT (abbreviation) for Class " + d).trim().toUpperCase());
+			.getConversionInput("Please provide SPELLSTAT (abbreviation) for Class " + d).trim().toUpperCase(Locale.ENGLISH));
```
thpr: Because this uses class name... :/

grimreaper: Is that a problem though? Are class names localized?

thpr: They could be... this goes back to Kar's original comment on the merged PR that was split into this and others - we should be designing these changes in a way that is compatible with future direction as well as current state - otherwise we may make things worse...

PR #4783, opened 2018-12-29
CODE-3200: add Locale.ENGLISH to computer-read values (id: 3:9074)

grimreaper:


code/src/java/pcgen/core/prereq/PrerequisiteUtilities.java
```diff
@@ -389,7 +390,7 @@ private static int countSubKeyWildcardMatch(final PlayerCharacter character, fin
 
 		for (String assoc : character.getConsolidatedAssociationList(ability))
 		{
-			final String fString = assoc.toUpperCase();
+			final String fString = assoc.toUpperCase(Locale.ENGLISH);
```
thpr: Associations are from data

PR #4787, opened 2018-12-29
Remove dead code in TaskStrategyListener

grimreaper:
These methods are never called and the contents of the methods don't
matter. Just remove them.

### DIFF
--- a/code/src/java/pcgen/gui2/converter/ConvertPanel.java
+++ b/code/src/java/pcgen/gui2/converter/ConvertPanel.java
@@ -73,36 +73,7 @@ public class ConvertPanel extends JPanel
 	{
 		super(new BorderLayout());
 		statusLabel = new JLabel();
-		TaskStrategyListener tsl = new TaskStrategyListener()
-		{
-			private String status;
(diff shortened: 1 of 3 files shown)


code/src/java/pcgen/gui2/converter/ConvertPanel.java
```diff
@@ -73,36 +73,7 @@ public ConvertPanel(List<ConvertSubPanel> bq)
...
-			{
-				long currentTime = System.currentTimeMillis();
-				if ((currentTime - time) > 100)
-				{
-					statusLabel.setText(status + " [" + string + "]");
-					time = currentTime;
-				}
-			}
-		};
+		TaskStrategyListener tsl = (source, string) -> statusLabel.setText(string);
```
thpr: This is simply not a valid reduction - not a Functional Interface - it has 3 methods...

grimreaper: I think this PR depends on other changes I have: the other two methods are unused.

PR #4799, opened 2018-12-30
Replace Cache class with Map

grimreaper:
This class doesn't provide much on type of a normal HashMap. It may have
done more in the past (prior to java 8?) but currently is trivially
replaced.


grimreaper:
wrong version

PR #4801, opened 2018-12-30
ItemFacet: return an Optional from `get`

grimreaper:
This makes it clear whether or not we're handling `null` properly. There 
are several cases here that look like bugs. I've attempted to change
nothing about the semantics. If you find bugs of the sort "we should add a
check there", but there wasn't one before, please comment, but I'll fix
them in a followup.

In a couple of cases we have some odd APIs (e.g., `.get().get()`) but IMHO
the tradeoffs are worth it.

### DIFF
--- a/code/src/java/pcgen/cdom/base/ItemFacet.java
+++ b/code/src/java/pcgen/cdom/base/ItemFacet.java
@@ -17,6 +17,8 @@
  */
 package pcgen.cdom.base;
 
+import java.util.Optional;
+
 /**
  * An ItemFacet is a Facet that stores a single item per identifier.
  * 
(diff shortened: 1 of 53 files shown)


thpr:
I would like this to stay pending for a bit to evaluate whether this impacts merging with a significant number of local branches

grimreaper:
> Also, this code unconditionally calls .get() meaning the optional is rather useless. So this leads a developer to think they are optional, which is a problem given actual behavior.
The point is that this makes it clear when this happens. Every unconditional `.get` is a place where a bug likely exists.

thpr:
But again, that's confusing.  If you are concerned about nulls, do requireNotNull, not turning the underlying into Optional when they aren't truly optional.  The semantics this has are deceptive.

grimreaper:
> But again, that's confusing. If you are concerned about nulls, do requireNotNull, not turning the underlying into Optional when they aren't truly optional. The semantics this has are deceptive.

It is my understanding that this is actually optional in many cases, but only required in a few. Is that not the case?

grimreaper:
We're eliminating Facet anyways, so closing this. Once we have a newer system I'll take a closer look at what should `Optional` vs not.

PR #4802, opened 2018-12-30
[tests] run tests in parallel

grimreaper:
Use number of cores as a rough estimate for how many to run at a time.
This should help with running tests on weaker machines as well.

This changes running `utest` from ~8:29 to ~5:21 on my machine under
normal load.


thpr:
I think we need a better method here.  I really am not after absolute speed - so we need different options for different people.  I am after "I still need to be able to do things in parallel to a compile" which is getting hard and reducing my productivity on PCGen.

grimreaper:
> I think we need a better method here. I really am not after absolute speed - so we need different options for different people. I am after "I still need to be able to do things in parallel to a compile" which is getting hard and reducing my productivity on PCGen.

I actually did this to help you while still maintaining some kind of greater speed. I'm fine deleting the parallel path entirely.

karianna:
> > I think we need a better method here. I really am not after absolute speed - so we need different options for different people. I am after "I still need to be able to do things in parallel to a compile" which is getting hard and reducing my productivity on PCGen.
> 
> I actually did this to help you while still maintaining some kind of greater speed. I'm fine deleting the parallel path entirely.

It's better to work off the number of available processors for sure, but I'd add a -1 (with a min of 1 obviously) so that there's always 1 hardware thread for other tasks.

thpr:
I was wondering more - rather than using processors is there an industry best practice whereby I can alter a local file / environment variable / eclipse setting of some sort on my machine to override the value - that way it can be set to a default that works for most (perhaps what you are doing in this case) but also overridden so it doesn't need to be a debate (nor do we need to account for how Intel vs AMD actually report threads back to software)

karianna:
Is there a local developer only (not checked in / git ignored) gradle properties file?

grimreaper:
> Is there a local developer only (not checked in / git ignored) gradle properties file?

`~/.gradle/gradle.properties`. I'm not sure about per project.

PR #4814, opened 2018-12-31
Update installers/win-installer/pcgen.nsi to modern standards

mjmeans:
This is an attempt to correct the following issues:

1. Windows 8 and later do not suport multiple level start menu folders. i.e. In Windows 7 the start menu might have PCGen > ver1 >pcgen.lnk and PCGen > ver2 > pcgen.lnk what you will see in Windows 8/10 is two copies of PCGen > pcgen.lnk. The folders ver1 and ver2 will not shown. The user will not know which link is which since the name is the same. So each version of PCGen has a collection of links that are mostly unique to that version, the top menu name needs to include the version number such as PCGen-ver1\pcgen.lnk and PCGen-ver2\pcgen.lnk. Windows will then place the two versions' links in separate folders in the start menu.

2. The uninstall registry entry DisplayName property needs to be labeled exactly the same as the PCGen program link name so that the control add/remove program items entry shows the correct icon and actual installation size. And since PCGen allows multiple versions to be installed, both the uninstall registry entry DisplayName and the the PCGen link name needs to also include the version number. Therefore, PCGen should use "PCGen ver1" for both the DisplayName and the start menu program link name.

I have no way to test this, so I will leave that task up to someone else. This PR may need to be tweaked to make it right.

### DIFF
--- a/installers/win-installer/pcgen.nsi
+++ b/installers/win-installer/pcgen.nsi
@@ -9,8 +9,8 @@
 !include "x64.nsh"
 
 ; Define constants
-!define APPNAME "PCGen"
-!define APPNAMEANDVERSION "${APPNAME} ${LONGVER}"
+!define APP "PCGen"
+!define APPNAME "$[APP] ${LONGVER}"
 !define APPDIR "${LONGVER}"
 !define TargetVer "1.10"
 !define OverVer "1.11"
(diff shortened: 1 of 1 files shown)


mjmeans:
It looks like someone changed the installer since I made this. Both problems I mention still exist in RC3.

PR #4821, opened 2019-01-01
experience plugin: avoid deprecated code and friends

grimreaper:


code/src/java/plugin/encounter/EncounterPlugin.java
```diff
@@ -38,15 +38,6 @@
...
 
-import gmgen.GMGenSystem;
-import gmgen.GMGenSystemView;
-import gmgen.io.ReadXML;
-import gmgen.io.VectorTable;
-import gmgen.plugin.InitHolderList;
-import gmgen.plugin.PcgCombatant;
-import gmgen.plugin.dice.Dice;
-import gmgen.pluginmgr.messages.AddMenuItemToGMGenToolsMenuMessage;
-import gmgen.pluginmgr.messages.RequestAddTabToGMGenMessage;
```
thpr: This is an imports ordering problem.  gmgen is part of pcgen and should be in the pcgen block.

code/src/java/plugin/experience/ExperienceAdjusterModel.java
```diff
@@ -19,16 +19,18 @@
 
 import java.io.File;
 import java.util.Observable;
+import java.util.stream.IntStream;
+
+import pcgen.core.SettingsHandler;
+import pcgen.util.Logging;
+import plugin.experience.gui.PreferencesExperiencePanel;
```
thpr: Same as above

code/src/java/plugin/experience/ExperienceAdjusterPlugin.java
```diff
@@ -31,15 +32,6 @@
...
 
-import gmgen.GMGenSystem;
-import gmgen.GMGenSystemView;
-import gmgen.plugin.Combatant;
-import gmgen.plugin.InitHolderList;
-import gmgen.pluginmgr.messages.AddMenuItemToGMGenToolsMenuMessage;
-import gmgen.pluginmgr.messages.CombatHasBeenInitiatedMessage;
-import gmgen.pluginmgr.messages.FileMenuSaveMessage;
-import gmgen.pluginmgr.messages.RequestAddPreferencesPanelMessage;
-import gmgen.pluginmgr.messages.RequestAddTabToGMGenMessage;
```
thpr: Same as above

code/src/java/plugin/experience/gui/ExperienceAdjusterView.java
```diff
@@ -497,12 +497,12 @@ public void valueChanged(ListSelectionEvent e)
 		add(jPanel6);
 	}
 
-	private double calculateRealValue(int i)
+	private static double calculateRealValue(int i)
```
thpr: Either this method is unused and should be deleted, or you are now violating the access rules for static methods by accessing them from an instance.

grimreaper:
My settings file had `gmgem` instead of `gmgen. Fixing. thanks.

code/src/java/plugin/experience/gui/ExperienceAdjusterView.java
grimreaper: It was being accessed without an explicit name, so it was allowed. Added `ExperienceAdjusterView.` as a prefix.

PR #4823, opened 2019-01-01
SettingsHandler: remove two unused settings (grimHPMode, grittyACMode)

grimreaper:


karianna:
Can HOmebrews not make use of this?

PR #4825, opened 2019-01-01
Settings: remove 'rog' setting

grimreaper:
This setting was initially added CVS in 1.175 of SettingsHandler.java.
At present there is no UI to modifying the setting, though in theory
it is honored if it was set in the past.

It doesn't seem useful to keep it around, is undocumented, and adds a
mild amount of complexity to some tokens.

### DIFF
--- a/code/src/java/plugin/exporttokens/SpecialAbilityToken.java
+++ b/code/src/java/plugin/exporttokens/SpecialAbilityToken.java
@@ -20,9 +20,7 @@ package plugin.exporttokens;
 
 import java.util.StringTokenizer;
 
-import pcgen.cdom.base.Constants;
 import pcgen.core.PlayerCharacter;
-import pcgen.core.SettingsHandler;
 import pcgen.io.ExportHandler;
 import pcgen.io.exporttoken.Token;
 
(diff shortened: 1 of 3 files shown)


code/src/java/plugin/exporttokens/SpecialAbilityToken.java
```diff
@@ -50,52 +48,15 @@ public String getToken(String tokenSource, PlayerCharacter pc, ExportHandler eh)
 			i = Integer.parseInt(aTok.nextToken());
 		}
 
-		if (aTok.hasMoreTokens())
-		{
-			String subToken = aTok.nextToken();
-
-			if ("DESCRIPTION".equals(subToken))
-			{
-				return getDescriptionToken(pc, i);
```
thpr: Technically, this is not logically equivalent, so it needs data (or perhaps OS) approval for a change.  Logical equivalence would return "" in the case of DESCRIPTION.

LegacyKing: Yeah, gonna say no here. 

Unless I'm mistaken, that is referencing DESC which is a multi-tag allowed token/tag in all our data. (Token has next, which means it appends, not replaces, unlike ASPECT/QUALITY/INFO), two other tags mimic that behavior - BENEFIT & SPROP.

If you can prove it's not going to break anything, then I might say yes - I have no idea about ROG, but export tokens, we should be very very clear before we start pulling anything out that might be used for outside support - old CMP/Homebrew situations.

grimreaper: As far as I could tell this only affects `DESCRIPTION` if `isROG` is ever true. This can only happen if people set this manually.

grimreaper: Restored the "" return

PR #4828, opened 2019-01-01
SettingsHandler: remove dead variable pcgenPreviewDir

grimreaper:


karianna:
@grimreaper How do you know this isn't used?

thpr:
@karianna There are a number of preferences where in our rebuild it looks like they never made it into the new UI.  Given that we have made SOO many incompatible changes that the CMP and other data would be unusable (and no one has reported this feature missing), and given we recommend people install new vs overwriting a previous version (so the old prefs wouldn't be used), that to me means these sections of dead code are effectively unused; or enough so we should take the risk and force folks to speak up if someone really is altering a preference file by hand.

PR #4829, opened 2019-01-01
Code 3328

javydreamercsw:
Some of the other proposed changes broke things on the installer. I suggest not messing with it any further unless someone has time to test it throughly.

I won't have any time soon as I resume work tomorrow.


grimreaper:
LGTM. Annoyingly I want to spend some more time on the build but don't really have a way to test the installer.

javydreamercsw:
I use a VM to test it. Maybe this helps: https://developer.microsoft.com/en-us/windows/downloads/virtual-machines

PR #4834, opened 2019-01-02
TextToken: be more explicit about localization issues

grimreaper:


code/src/java/plugin/exporttokens/TextToken.java
```diff
@@ -167,9 +166,9 @@ else if (replaceType.equalsIgnoreCase("FIRST"))
 	 * @param value The value to be modified. 
 	 * @return The value in sentence case.
 	 */
-	private String changeToSentenceCase(String value)
+	private static String changeToSentenceCase(String value)
 	{
-		String temp = value.toLowerCase();
+		String temp = value.toLowerCase(Locale.ENGLISH);
```
thpr: Why are the others getDefault and this is ENGLISH? (and similar below)

grimreaper: cause I put the wrong version up for review :(

fixing.

PR #4841, opened 2019-01-03
PlayerCharacterOutput: `cr` is never fractional (remove dead code)

grimreaper:


thpr:
This is a bold claim that cr is never fractional.  What supports that claim?

grimreaper:
> This is a bold claim that cr is never fractional. What supports that claim?
```
		Integer calcCR = display.calcCR(); // calcCR is an integer. 
		float cr = (calcCR == null) ? -1 : calcCR; // cr is either -1 (an integer) or calcCR (an integer)
		String retString = "";
```

Further: `CharacterDisplay.calcCR()` returns an `Integer`. and `ChallengeRatingFacet.getCR` returns an `Integer`.

This mostly looks wrong from a rules PoV (CR can be fractional in D&D or pathfinder) but correct from a code point of view.

LegacyKing:
So, we should rectify that, because @Zaister set up specific code to support Fractional CRs.

code/src/java/gmgen/plugin/PlayerCharacterOutput.java
```diff
@@ -106,15 +105,7 @@ public String getCR()
 		float cr = (calcCR == null) ? -1 : calcCR;
 		String retString = "";
 
-		// If the CR is a fractional CR then we convert to a 1/x format
```
thpr: This code as shown is a bug, and we need to fix not remove.  Please see plugin.exporttokens.CRToken for how fractions are handled (negative numbers are fractions)

PR #4848, opened 2019-01-05
DATA-3812

mjmeans:
Partial fix,


grimreaper:
I'm having some trouble seeing the change. Is this basically a removal of `TEMPBONUS:EQ` ?

mjmeans:
No. I've added an "Off-Hand" choice and removed the "Natural Primary - Simple" choice.

I've also fixed the formula for damage by replacing them with the value PiranhaStrikeDamageModifier and PiranhaStrikeDamageModifier/2 as appropriate. Basically it was calculating 50% bonus of the BAB 1 total damage (2 becoming 1) and then adding 2 for each BAB 4 over that. This is wrong. It should have taken the total BAB 1 +2 + the BAB adjusted additional +2s before adjusting -50% for off-hand (copied form power attack)  and natural secondary damage. i.e. BAB 8 should have a damage bonus of +3, not +5.

I think Power attack has the same problem, bit I didn't touch that.

PR #4862, opened 2019-01-06
Equip node interface removal

thpr:
Removes an interface
As a side effect, this is also some early preparation for fixing the kit-can't-equip-items bug


code/src/java/pcgen/gui2/facade/EquipNode.java
```diff
@@ -0,0 +1,242 @@
...
+	
+		/** This is a node representing an equipped item of equipment. */
+		EQUIPMENT;
+	}
+
+	private static final NumberFormat FMT = new DecimalFormat("00");
+	private final NodeType nodeType;
+	private final BodyStructure bodyStructure;
+	private Equipment equipment;
+	EquipNode parent;
```
grimreaper: why are some things not private here?

grimreaper:
overall LGTM

code/src/java/pcgen/gui2/facade/EquipNode.java
thpr: This is a minimal change and that's how they were before

PR #4879, opened 2019-01-06
Fix bug: DATA-3805 - Swashbuckler Finesse bonuses similar to Weapon F…

LegacyKing:
…inesse doubling Penalties.


grimreaper:
LGTM but I'm no data monkey

PR #4882, opened 2019-01-06
DataLoadTest:remove some dead stores

grimreaper:


grimreaper:
hrm... wasn't as dead as I thought.

PR #4889, opened 2019-01-06
One more round of removing is()

grimreaper:


raihle:
LGTM. 

(EvaluatorFactoryTest actually looks horrible, but at least _better_)

PR #4890, opened 2019-01-06
CompareEqualInt: eliminate

grimreaper:


raihle:
Looks good, except that `eq(int)` still has usages.

grimreaper:
> Looks good, except that `eq(int)` still has usages.

not anymore :)

PR #4926, opened 2019-01-07
RequiredTokenTest: convert to JUnit 5

grimreaper:


code/src/utest/plugin/lsttokens/datacontrol/RequiredTokenTest.java
```diff
@@ -112,7 +115,7 @@ public void testValidStringNo()
 		assertNull(cd.getRequired());
 		assertTrue(token.parseToken(context, cd, "NO").passed());
 		assertNotNull(cd.getRequired());
-		assertFalse(cd.getRequired().booleanValue());
+		Assertions.assertFalse(cd.getRequired().booleanValue());
```
thpr: This feels inconsistent - assertNotNull is statically imported, but this is directly used?

grimreaper: likely just missed it in the migration.

PR #4940, opened 2019-01-07
This fixes a race condition on creation of formats for CDOM classes

thpr:
This fixes - in hopefully a brute-force enough fashion that it's "final" - the problem of how we access a format like "SKILL".  The historical problem has been "SKILL" as a format isn't defined until the first skill is loaded, which is very late.  A previous system had tried to hack through that and say "if you requested something we recognize as a CDOM format, we'll just create it on the fly", but it has a bit of a loophole that will fail.  Asking for an Array of a format before the base format will cause the logic to fail.  This forces a pre-build of all of the "known basic formats" so that this shouldn't occur (for at least those base formats - which is most of what we will use in the formula system to start)

### DIFF
--- a/code/src/java/pcgen/rules/context/RuntimeReferenceContext.java
+++ b/code/src/java/pcgen/rules/context/RuntimeReferenceContext.java
@@ -57,7 +57,8 @@ public class RuntimeReferenceContext extends AbstractReferenceContext
 	public void initialize()
 	{
 		super.initialize();
-		StringPClassUtil.getBaseClasses().forEach(c -> importCDOMToFormat(c));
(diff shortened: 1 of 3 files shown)


code/src/java/pcgen/rules/context/RuntimeReferenceContext.java
```diff
@@ -52,6 +53,13 @@ protected RuntimeReferenceContext()
 	{
 	}
 
+	@Override
+	public void initialize()
+	{
+		super.initialize();
+		StringPClassUtil.getBaseClasses().forEach(c -> importCDOMToFormat(c));
```
karianna: what's a `c`?

PR #4944, opened 2019-01-07
AbstractListFacetTest+friends: convert to JUnuit 5

grimreaper:


code/src/utest/pcgen/cdom/testsupport/AbstractListFacetTest.java
```diff
@@ -427,7 +431,7 @@ public void testRemoveAllNull()
 		try
 		{
 			getFacet().removeAll(id, null);
-			fail();
+			fail()
```
thpr: One of your scripts is going bad - this is another example of this problem.

PR #4952, opened 2019-01-08
Moves sorting of certain elements out to the UI

thpr:
Lowers dependency on the core to do work for the UI


grimreaper:
not sure I'd rename them all, but the core looks fine.

PR #4959, opened 2019-01-08
PrerequisiteWriterTest: convert to JUnit 5

grimreaper:


code/src/test/pcgen/persistence/lst/output/prereq/PrerequisiteWriterTest.java
```diff
@@ -456,127 +422,81 @@ protected void setUp() throws Exception
...
-		{
-			if (!bExceptionThrown)
+			else
 			{
-				if (prereq != null)
-				{
-					System.out.println(prereq.toString());
-				}
-				fail("exception expected but not thrown");
+				assertNotNull(PreParserFactory.getInstance().parse(aPreString));
```
thpr: This is flawed - it needs to assign prereq

code/src/test/pcgen/persistence/lst/output/prereq/PrerequisiteWriterTest.java
```diff
@@ -456,127 +422,81 @@ protected void setUp() throws Exception
...
-			fail("Could not parse prereq: '" + aPreString + "'");
-		}
 
-		StringWriter sw = new StringWriter();
-		PrerequisiteWriter writer = new PrerequisiteWriter();
-		try
-		{
+			StringWriter sw = new StringWriter();
+			PrerequisiteWriter writer = new PrerequisiteWriter();
 			writer.write(sw, prereq);
```
thpr: Given current code, this seems to unconditionally write a null object?

code/src/test/pcgen/persistence/lst/output/prereq/PrerequisiteWriterTest.java
```diff
@@ -456,127 +422,83 @@ protected void setUp() throws Exception
...
```
thpr: We still have a problem.  The return statement that used to be on line 492 used to abort when an exception was expected.  I believe after line 433 assertThrows you may need a continue; or some such to ensure that the same logic is followed.  The main problem is prereq can still be null here, and thus we can still try to write a null prereq.

grimreaper:
This is this now conflicting in a still-mergeable way. Closing and will recreate.

PR #4973, opened 2019-01-09
APG - Can't have SERVESAS and ABILITY with the same name on the same line

Llisandur:
It does strange things.


Llisandur:
* What went wrong:

Execution failed for task ':datatest'.

> Process 'Gradle Test Executor 5' finished with non-zero exit value 1

  This problem might be caused by incorrect test process configuration.

  Please refer to the test execution section in the user guide at https://docs.gradle.org/5.1/userguide/java_testing.html#sec:test_execution

PR #4975, opened 2019-01-10
[FOR FREEZE] Remove test (for now) that causes a transient error

thpr:
This is one part of lifting the current code freeze - this eliminates temporarily a transient error being caused by weak references / unpredictable garbage collection.

Longer term, the associated tokens need better behavior (they should store the references in the object, actually, even if unused,  so that errors are guaranteed to be displayed)

### DIFF
--- a/code/src/utest/plugin/lsttokens/testsupport/AbstractListInputTokenTestCase.java
+++ b/code/src/utest/plugin/lsttokens/testsupport/AbstractListInputTokenTestCase.java
@@ -591,16 +591,21 @@ public abstract class AbstractListInputTokenTestCase<T extends CDOMObject, TC ex
 		}
 	}
 
-	@Test
-	public void testInputInvalidClearDot()
-	{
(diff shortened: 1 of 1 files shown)


thpr:
"The attribute enabled is undefined for the annotation type Test"

I think this is potentially still a JUnit 3 test class, so I don't think it has that yet... I suspect we want to do the JUnit 3->4->5 test updates separately

thpr:
As best I can tell, because of how this is being invoked (via TestCase class), this is still running under JUnit 3 rules.  Adding both @Disabled and @Ignored has no effect on whether the test is run.  @grimreaper Can clean this up (uncomment it) when he gets to updating this class hierarchy to JUnit 5

grimreaper:
yeah, this is a Junit 3 class. Annoyingly we had to keep some mixed classes. Additionally, any of the `.gc` based testcases are likely to fail.

Going to merge this and then look to see what else fails.

PR #4979, opened 2019-01-10
[FOR FREEZE] Move around to have finishLoad() in CharacterTestCase

thpr:
This is another portion of fixing the problems we have in master and eventually lifting the code freeze.

This changes how certain tests behave. Due to some enhancements deployed over the last few days, a number of features have interacted, causing a few badly behaved unit tests to be exposed in their less than stellar behavior. Some of these unit tests required additional setup for a test before they "finish the load" and resolve references. In order to accommodate that, a fairly significant change to AbstractCharacterTestCase was required. This will fix a number of unit tests that are currently failing (3 ish?)


grimreaper:
which tests are failing without this change?

PR #4989, opened 2019-01-12
OA_spell_cleanup_1_12_19

evilpixie87:
Moved Psychic Spells from ARG to support to prevent unwarranted error messages.
Added TYPE:Psychic to all Psychic class spells


grimreaper:
positive review from data, and test was a false negative.

PR #4991, opened 2019-01-13
[FOR FREEZE] Eliminate Parallelism

thpr:
(Travis is dying of what appear to be memory issues - likely due to
parallelism)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,2 +1,1 @@
 version=6.09.00-DEV
-org.gradle.parallel=false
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,2 +1,2 @@
 version=6.09.00-DEV
-org.gradle.parallel=true
\ No newline at end of file
+org.gradle.parallel=false
```
grimreaper: remove this line rather than set it to false.

grimreaper:
acc' to https://docs.travis-ci.com/user/reference/overview/ travis has 7.5G of memory which is roughly what I have. Surprising how it fails.

grimreaper:
(just realized this is more than one commit. In the future I'll try and squash)

PR #4992, opened 2019-01-13
Add @AfterEach and @AfterAll as appropriate for JUnit5 tests

thpr:
Also makes some things private
Also moves variables to the top of a class


grimreaper:
I don't understand why this is required. Does JUnit hold on to a reference causing a lack of GC?

karianna:
> I don't understand why this is required. Does JUnit hold on to a reference causing a lack of GC?

Yes, JUnit runs all tests in a single JVM unless you fork per JVM (*very* expensive).  JUnit tests should be idempotent.

PR #4995, opened 2019-01-13
[build] modernize our gradle a bit.

grimreaper:
Gradle has a habit of constantly requiring new syntax or features.
Follow along.

-  switch from 'compile' to 'implementation' and 'api' (we have none in
this project)
- use newer getters/setters
- avoid some cross-configuration setting
- update to gradle 5.1.1 (possible dup of existing PR)


grimreaper:
This is probably wrong and conflicts with other work. Splitting it out into better changes.

PR #5006, opened 2019-01-13
Library195

thpr:
Updates Libraries from 166/176 to 170/195

Mainly this 
(a) Uses another Optional<> rather than having a null
(b) Changes how variables are initialized so that derived formats like ARRAY[NUMBER] don't need data intervention to specify a ton of basically useless default values.
There is one minor bug fix included (COLUMN didn't properly report its format)


system/gameModes/default/compatibilityDataControl.lst
```diff
@@ -36,4 +36,5 @@ DEFAULTVARIABLEVALUE:BOOLEAN|False
 DEFAULTVARIABLEVALUE:STRING|
 DEFAULTVARIABLEVALUE:DICE|0
 DEFAULTVARIABLEVALUE:ALIGNMENT|None
+DEFAULTVARIABLEVALUE:HANDED|Right
```
grimreaper: does this require data approval?

thpr: Not particularly, it's extracting it out of the existing code (hardcoded)

PR #5010, opened 2019-01-13
Remove duplicate if conditions

grimreaper:


code/src/java/pcgen/io/exporttoken/WeaponToken.java
```diff
@@ -1887,8 +1887,8 @@ else if (hitMode == HITMODE_OHHIT)
 				baseBonus += pc.getOffHandLightBonus();
 			}
 
-			if ((hitMode == HITMODE_TWOHIT) || (hitMode == HITMODE_OHHIT) || (hitMode == HITMODE_TWFOHL)
-				|| (hitMode == HITMODE_TWFOHH))
+			if ((hitMode == HITMODE_OHHIT) || (hitMode == HITMODE_TWFOHL)
+					|| (hitMode == HITMODE_TWFOHH))
```
thpr: Doesn't this strike you more as a bug than a potential optimization?

karianna: I don't see the duplicate either?

grimreaper: I don't think its a bug, but you're right that this is weird. better fix applied.

code/src/java/pcgen/io/exporttoken/WeaponToken.java
```diff
@@ -1550,7 +1548,7 @@ else if (display.isSecondaryWeapon(eq) && display.hasPrimaryWeapons())
 			else
 			{
 				// offhand heavy
-				hitMode = HITMODE_TWFOHH;
+				hitMode = HITMODE_TWOHIT;
```
LegacyKing: Looks like we have logic changes here.  We had Two Weapon Fighting Off Hand Heavy {TWFOHH} changed to TWOHIT.

karianna:
So basically even though the two constants have the same value, it's still useful to have both for readability..

LegacyKing:
It looks like we're losing an entire HitCategory (We have 4 conditions dropped to 3). I don't feel good about this change. For some reason we've completely dumped the Two Weapon Off Hand Heavy group.  What exactly is this supposed to be fixing or addressing? (Mind you, EQUIPMENT is one of the most heavily magic'd code in pcgen that I'm aware of, and these values are displayed on the character sheet).

karianna:
I'm actually going to close this PR.

PR #5019, opened 2019-01-14
[PFS] Blood fo the Beast

mjmeans:
Enable PFS options for Blood of the Beast


mjmeans:
Hold off on this I have to run another check.

mjmeans:
Okay. Apparently travis doesn't check for errors with game mode Pathfinder_PFS and the Pathfinder Society Core Assumption which broke (more) with this PR. There appears to be an issue with the core assumption and a different error (unconstructed error with the skill, craft tattoos). I will have to fix that before Pathfinder Core Assumption can be added back to CI. But that fix will be in a separate PR. This PR should work now.

PR #5071, opened 2019-01-21
Mac DMG build support

treaves:
- Updated the .gitignore file to ignore the Mac .DS_Store files
- Added a Mac-installer directory with the resources needed for a Mac app and DMG
- Updated the build.gradle file to build the DMG
- Tested on Ubuntu; if Mac is not the OS, then the DMG target is not run.


build.gradle
```diff
@@ -179,8 +174,8 @@ dependencies {
     compile group: 'org.mozilla', name: 'rhino', version:'1.7.10'
 
     compile group: 'org.xmlunit', name: 'xmlunit-core', version:'2.6.2'
-    implementation group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.170'
-    implementation group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.195'
+    compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.170'
+    compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.195'
```
LegacyKing: Was this change intentional?

treaves: Yes.  The issue is that with them being marked implementation, they do not get copied over for inclusion into the package.  So ClassNotFound exceptions when running the app. (incidentally, with the resources not being in src/main/resource, they do not get copied over either, and I had to add steps to do so).

I'm not a gradle expert, and their documentation leaves a lot to be desired. The documentation talks about exposing the dependencies to consumer classpaths; I've been wringing Java for a couple of decades, and have no idea what they mean by that.  With maven, things can be marked for inclusion to the product without being added to the classpath, but, that doesn't seem to be what this means.

Anyway, another option would be going with a fat jar file, so that everything ends up in one jar; maybe then the api & implementation vs. compile would make more sense?

grimreaper: yes please, though note #5072. I've been (slowly) working on improving our build. It is also a moving target as gradle makes changes.

build.gradle
```diff
@@ -220,6 +215,57 @@ jar {
     }
 }
 
+macAppBundle {
+    mainClassName	= "pcgen.system.Main"
+    appName			= "PCGen"
+    appStyle		= "universalJavaApplicationStub"
+    icon 			= "installers/mac-installer/PCGenApp.icns"
+    bundleJRE 		= false
```
grimreaper: IMHO we should bundle it: many users only have Java 8.

build.gradle
```diff
@@ -220,6 +215,57 @@ jar {
...
+    backgroundImage = "installers/mac-installer/PCGen_folder_background.png"
+    jvmVersion		= "+11.0"
+    appIconX		= "243"
+    appIconY		= "270"
+    appFolderX		= "370"
+    appFolderY		= "270"
+    javaProperties.put("apple.laf.useScreenMenuBar", "true")
```
grimreaper: We also have. I don't know how much this all matters.
```
        public static void configurePlatformUI()
        {
                System.setProperty("com.apple.macos.useScreenMenuBar", "true");
                System.setProperty("apple.laf.useScreenMenuBar", "true");
                System.setProperty("com.apple.macos.use-file-dialog-packages", "true");
                System.setProperty("com.apple.mrj.application.growbox.intrudes", "false");
                System.setProperty("com.apple.mrj.application.live-resize", "true");
                System.setProperty("apple.awt.brushMetalLook", "true");
        }
```

build.gradle
```diff
@@ -323,6 +368,8 @@ clean {
     it.dependsOn 'cleanRoot'
 }
 
+
```
grimreaper: remove this.

build.gradle
grimreaper: actually, I need to understand the implication a little better: the documentation for these two is primarily around android. In the meantime leave them alone as `compile.`

karianna:
@treaves Apologies I merged in some other build fixes so you'll need to rebase

build.gradle
treaves: Neither Java 10 nor Java 11 have a JRE; that is not more.  What needs to happen with Java 11 is a custom runtime be built using the new modular approach.  The DMG plugin does not support that, and neither does PCGen. The DMG plugin doesn't verify that it's a JRE, so, a custom Mac runtime can be created and bundled.

In the meantime, if a user does not have Java 11, they will git a dialog first asking to allow system notifications, and then informing them that Java 11 needs to be installed.

build.gradle
treaves: Removed in 9258f11e90ded8d2193db5c2e4b1ee4edaa4ee2c

javydreamercsw:
Should this be part of build Release, or something like that?

karianna:
> Should this be part of build Release, or something like that?

I think so yes.

treaves:
Guys, don't know what's going on here, but, this PR seems to have stalled.

javydreamercsw:
We could either add it to build Release target or open a new issue to cover that part and link it to close this one.

mertonmonk:
Has the build.gradle conflict been resolved?  Otherwise it looks fine to me.

LegacyKing:
@treaves I agree with @javydreamercsw - this should be on the release.gradle. Also, branch conflicts. You able to resolve?

treaves:
@LegacyKing @javydreamercsw Done.

build.gradle
LegacyKing: Actually look deeper, we are bundling a Java 11 JRE from AdoptOpenJDK.

LegacyKing:
Awesome, check out the comment I made about JRE. And I can't merge via the rebase option there is conflicts... Bad practice to PR from master.

treaves:
- The JRE is only created for the Windows build.  The gradle file may or may not be able to be modified to create JRE's for Linux and Mac.  That isn't related to createing a DMG, and can be looked at in a separate change.
- As a general statement, it's not bad practice to PR from master.  Master is a branch like any other. It may have particular properties ascribed to it based on a groups desired branching model.  It may not be this groups norms, but I did look over the WIKI's documentation.  If the preference is to create a branch - such as Story Based Branching - that's fine, I'll do that in the future.
- I pushed the vastest merges.  The details section is currently showing no conflicts.

javydreamercsw:
Adding other JRE shouldn't be too hard. Just needs different parameters when using their API.

javydreamercsw:
Were you able to make it embed the Mac Java with my changes then?

LegacyKing:
Still conflicted?  

<<<<<<< master

    compile group: 'org.xmlunit', name: 'xmlunit-core', version:'2.6.2'
    compile group: 'net.sourceforge.pcgen', name: 'PCGen-base', version:'1.0.170'
    compile group: 'net.sourceforge.pcgen', name: 'PCGen-Formula', version:'1.0.195'
    compile group: 'xml-apis', name: 'xml-apis', version:'1.3.04'
=======
    compile group: 'org.xmlunit', name: 'xmlunit-core', version:'2.6.2'
>>>>>>> master

treaves:
> Were you able to make it embed the Mac Java with my changes then?

No; there are issues with both the way the bundler copies over the JRE and with the way the stub script searches for them.  I've filed defects.

PR #5088, opened 2019-01-22
Add AbstractReferenceFacade to share event code

thpr:


code/src/java/pcgen/facade/util/AbstractReferenceFacade.java
```diff
@@ -0,0 +1,80 @@
...
+	 * @param old
+	 *            The old value of the reference underlying this AbstractReferenceFacade
+	 * @param newer
+	 *            The new value of the reference underlying this AbstractReferenceFacade
+	 */
+	protected void fireReferenceChangedEvent(Object source, T old, T newer)
+	{
+		Object[] listeners = listenerList.getListenerList();
+		ReferenceEvent<T> e = null;
+		for (int i = listeners.length - 2; i >= 0; i -= 2)
```
karianna: Why `-2`? Would you not miss the last element in the array?  And why are we using an Array instead of a Type safe coellction?

thpr: I'm leveraging javax.swing.EventListenerList since that is a Java object to leverage rather than a custom object.

The listeners are stored in one array, with the class of the listener type in location i and the object itself in location i+1.  length -2 is required so you query the class (does it match?) before you grab the object from location i+1.  Note also the for loop decrement is -2, so it jumps from class to class.  It's done backwards, so the last object added receives the event first.

This is "boilerplate" code that is throughout the Java JDK/JRE code base.

karianna:
@thpr merge conflicts.

thpr:
Fixed

PR #5107, opened 2019-01-24
Add Integration Test for Modify

thpr:
Clean up misc errors associated with the test/improve test cleanup

### DIFF
--- a/code/src/itest/tokencontent/GlobalModifyTest.java
+++ b/code/src/itest/tokencontent/GlobalModifyTest.java
@@ -65,7 +65,7 @@ public class GlobalModifyTest extends AbstractContentTokenTest
 		if (result != ParseResult.SUCCESS)
 		{
 			result.printMessages(TestURI.getURI());
-			fail("Test Setup Failed");
+			assertEquals("Test Setup Failed", ParseResult.SUCCESS, result);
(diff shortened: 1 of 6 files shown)


thpr:
@grimreaper What time frame is that?  In addition to a test, this has code fixes that are gate other PRs

grimreaper:
> @grimreaper What time frame is that? In addition to a test, this has code fixes that gate other PRs

I was basing my reply on the title and brief read of the diff. Will review in full.

code/src/itest/tokencontent/GlobalModifyTest.java
```diff
@@ -0,0 +1,132 @@
...
+		super.setUp();
+		TokenRegistration.register(SMF);
+		context.getVariableContext().assertLegalVariableID("MyVar",
+			context.getActiveScope(), FormatUtilities.NUMBER_MANAGER);
+	}
+
+	@Override
+	public void processToken(CDOMObject source)
+	{
+		ParseResult result = token.parseToken(context, source, "MyVar|SET|30");
```
grimreaper: can we use `assertTrue` here?  It might be worth us writing our own matchers eventually, but trying to move away from use of `fail`

PR #5145, opened 2019-01-25
UE - Fix Masterwork Tool skill bonus

Llisandur:


Llisandur:
Please squash commits.

PR #5147, opened 2019-01-26
Eliminate Custom Bio Fields

thpr:
The custom fields produce some really nasty indirection problems
for refactoring, so this eliminates them (by making them always
visible).  It is likely they will be restored in the future, albeit with
more flexibility.


grimreaper:
I used to use these but I'm not convinced it worth keeping.

PR #5174, opened 2019-01-28
Remove a few more dead classes

grimreaper:


grimreaper:
looks like this depends on another not-yet-submitted PR. sorry

PR #5178, opened 2019-01-28
Hide Race Size Selection when it's not used

Llisandur:


LegacyKing:
I'm not sure this is the best method to achieve this... Let's discuss tomorrow when we're both available on Slack.

Llisandur:
This is only used by aasimar and tiefling, and it shows up as a useless category on all of the other races.

PR #5256, opened 2019-02-02
[build] remove vintage test runner.

grimreaper:
We're done!


karianna:
Not yet we're not! :-)

PR #5276, opened 2019-02-09
Fix Biosettings files

Llisandur:
I think there was a recent change that made the SOURCEx tags at the top of the file no longer legal for biosettings.


thpr:
This should be opened as a code bug not fixed in data

Llisandur:
I wasn't sure, since only 3 files have this.

LegacyKing:
I don't think these files ever possessed SOURCExxx pages. Since the biosetting file is dual homed - GameMode and Data. There would never be a call to trace the sourcepage for it. Ergo, going to have the Data Fix merge. If we want to enable the SOURCExxx on said files, great.

PR #4759, opened 2018-12-27
Set up Releases to work

LegacyKing:

PR #5311, opened 2019-03-03
Revert "Set up Releases to work (#4759)"

grimreaper:
This reverts commit ca077bf0deba1cd62424d60b089ef52ceb802a8d.


grimreaper:
This is only for if #5312 doesn't work.

PR #5312, opened 2019-03-03
Fix .travis.yml

grimreaper:
Without this, all travis checks are broken.


karianna:
@grimreaper travis build failed

grimreaper:
> @grimreaper travis build failed

I was watching, fixing, and restarted. Merged cause now the check is required.

PR #5334, opened 2019-03-26
[5e] Fix Custom Rules Selection

Llisandur:


grimreaper:
Seems reasonable to me - though I'm not sure if this is indicative of requiring something else fixed.

Llisandur:
It's the same way Pathfinder does it. Otherwise you can't select them unless something puts points into the pool.

PR #5339, opened 2019-03-31
DATA-3866: correct weight for Mirthal shirt

grimreaper:
Set weight to 20 since something later divides the value by 2. the
correct weight that we'd like to retain is 10.


grimreaper:
This is quite possibly wrong practice. I read through as much as I can (given its past 1:00am for me). If you could point me where the division by 2 is happening (data or code) there is likely a better way to fix this.

PR #5354, opened 2019-04-10
[pathfinder data] change source to be more specific

grimreaper:
This avoids duplicative reports of the 'Mage Hand' spell. Pick the more
specific one to make it clear where things are coming from.


LegacyKing:
@grimreaper can you fix the conflicts please?

PR #5355, opened 2019-04-10
Bonus: cleanup

grimreaper:
- add Locale since Bonus is always internal name, not translatable text
- use modern newClass
- don't swallow exception if unable to build bonus class
- avoid explicit StringBuilder. Java does this automatically if needed.

### DIFF
--- a/code/src/java/pcgen/core/bonus/Bonus.java
+++ b/code/src/java/pcgen/core/bonus/Bonus.java
@@ -17,6 +17,8 @@
  */
 package pcgen.core.bonus;
 
+import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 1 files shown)


grimreaper:
> Hard coding LOCALE english will break i18n

how so? This is for a BONUS and the locale doesn't affect this. I.e., "NaturalArmorBonus" or whatnot

karianna:
> > Hard coding LOCALE english will break i18n
> 
> how so? This is for a BONUS and the locale doesn't affect this. I.e., "NaturalArmorBonus" or whatnot

@thpr are these Strings immutable or can they be translated?

PR #5358, opened 2019-04-10
[UC] Added TEMPBONUS tags to Archaeologist's Luck ability

GrimaceJD:
Now it can be added to the character in the PCGen UI


grimreaper:
LGTM but I'm no data monkey

PR #5361, opened 2019-04-12
[core rules] replace 'magic item' with specific magic item.

grimreaper:
This makes the output sheet a little nicer and clearer (at the expense
of being a little longer).


grimreaper:
ref #5354

PR #5366, opened 2019-04-13
[menu] Enable temporary bonus menu in "edit"

grimreaper:
This looks like it'd be useful. Also remove a dead menu item.

### DIFF
--- a/code/src/java/pcgen/gui2/PCGenFrame.java
+++ b/code/src/java/pcgen/gui2/PCGenFrame.java
@@ -26,12 +26,8 @@ import java.awt.GridBagConstraints;
 import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.awt.Rectangle;
-import java.awt.event.ActionEvent;
-import java.awt.event.ActionListener;
 import java.awt.event.ComponentAdapter;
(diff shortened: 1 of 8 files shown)


code/src/java/pcgen/gui2/PCGenFrame.java
```diff
@@ -125,12 +122,13 @@
...
  * global UI functions such as message dialogs. 
  */
-public final class PCGenFrame extends JFrame implements UIDelegate
+public final class PCGenFrame extends JFrame implements UIDelegate, CharacterSelectionListener
 {
 
 	private final PCGenActionMap actionMap;
 	private final CharacterTabs characterTabs;
 	private final PCGenStatusBar statusBar;
+	private final PCGenMenuBar pcGenMenuBar;
```
karianna: I find it odd that we are adding the menubar here for the first time.  Something seems wrong here.

grimreaper: We're just keeping a referencence. The menu bar was always created in this class. See line 180 of the "old" side.

karianna: Ah I see now thanks

PR #5371, opened 2019-04-15
[global] replace SwingWorker V3 with native SwingWorker

grimreaper:
SwingWorker v3 was a backwards compatible port for pre JDK-6. This is
using the current version of the API. Further work could allow properly
handling the status bar using setProgess/publish, or moving some
work onto a backgroudn thread rather than starting and waiting for the
thread right away.


karianna:
@grimreaper merge conflicts

grimreaper:
resolved

PR #5372, opened 2019-04-15
PCGenActionMenu: delete unsupported actions

grimreaper:
I'd be happy to work on or implement some of these in the future, but
its confusing when you accidentally press a key to see an exception.
Also, it makes it seem like there is more implemented than actually
exists.

This also adds missing tooltips and missing localization to menu items.


karianna:
@grimreaper merge conflicts

grimreaper:
resolved

PR #5376, opened 2019-04-16
DATA-3890 [UE] Fixed price of Four-Mirror armor

GrimaceJD:
Price was incorrect in UE files, now fixed.


grimreaper:
LGTM but I'm no data monkey.

PR #5386, opened 2019-04-16
[HtmlSheetSupport] vastly simplify

grimreaper:
- As far as I could tell from spelunking source code, the properties
being set no longer matter.
- Further, testing shows no functional impact nor any obvious
performance impact.

Also modernize the code a bit.


grimreaper:
hrm. Nevermind. I may have broken something with this.

PR #5397, opened 2019-04-20
Patch 2019 04 19

LegacyKing:


outputsheets/d20/5e/pdf/common_sheet/block_spells_condensed.xslt
```diff
@@ -30,10 +30,10 @@
 					<xsl:if test="warlock &gt; 0">
 						<xsl:apply-templates select="spell_slots_warlock"/>
 					</xsl:if>
-					<xsl:apply-templates select="spells_innate/racial_innate"/>	
-					<xsl:apply-templates select="spells_innate/class_innate"/>	
+<!--					<xsl:apply-templates select="spells_innate/racial_innate"/>			-->
```
grimreaper: Is this a workaround for a code bug? Not showing racial innate spells seems odd

LegacyKing: Yes, it's a hack til we figure out why this is causing an NPE.

PR #5408, opened 2019-04-21
[directional - do not not merge] replace lobobrowser with JavaFX WebView

grimreaper:
I am submitting this as a PR primary to start some high level discussion. The code is incomplete/buggy and should **not** be merged. 

At minimum it presents a bug where opening a new character when one is already open either
- locks up pcgen (if on charsheet)
- stop showing the character sheet preview entirely (if off of charsheet).

That said, before I get too far, I'd like some feedback on the overall direction I'm taking here.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ plugins {
 }
 
 javafx {
-    version = "12"
+    version = "12.0.1"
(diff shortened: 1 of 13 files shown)


grimreaper:
We decided on JavaFX (by virtue of no one disagreeing :))

PR #5438, opened 2019-05-02
CODE-3248: store language choosers in CharacterFacadeImpl

cmhulbert:
Store the language choosers list in the CharacterFacadeImpl after creation, so we can add to it later, when changes are triggered.

Mainly, we want to ensure the bonus language chooser is not available on a new character if there is no race, but this will also handle instances of bonus languages being added by intelligence increase, or other general triggers that call `refreshLanguageList`.

It will also remove the bonus chooser if it exists, and bonus languages remaining has been set to 0, or race was removed.

Coincidentally, this also removes the bonus language todo's if a race is not selected, or is removed.

### DIFF
--- a/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
+++ b/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
@@ -2166,7 +2166,11 @@ public class CharacterFacadeImpl
 				currBonusLangs.add(lang);
 			}
 		}
-		int bonusLangRemain = bonusLangMax - currBonusLangs.size();
+		int bonusLangRemain = 0;
+		if (!theCharacter.getRace().getDisplayName().equals("<none selected>"))
+		{
+			bonusLangRemain = bonusLangMax - currBonusLangs.size();
+		}
 		if (!allowBonusLangAfterFirst && !atFirstLvl)
 		{
 			bonusLangRemain = 0;
--- a/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
+++ b/code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
@@ -2167,7 +2167,7 @@ public class CharacterFacadeImpl
 			}
 		}
 		int bonusLangRemain = 0;
-		if (!theCharacter.getRace().getDisplayName().equals("<none selected>"))
+		if (theCharacter.getRace().isUnselected())
 		{
 			bonusLangRemain = bonusLangMax - currBonusLangs.size();
 		}
@@ -2218,13 +2218,15 @@ public class CharacterFacadeImpl
 
 			/* Ensure the bonus language chooser is removed, if it exists.*/
 			Iterator<LanguageChooserFacade> itr = langChoosersList.iterator();
-			while (itr.hasNext())
+			if (itr.hasNext())
 			{
-				LanguageChooserFacade chooser = itr.next();
-				/* If we find an add bonus chooser, remove it.*/
-				if (chooser.getName().equals(LanguageBundle.getString("in_sumLangBonus")))
+				for (LanguageChooserFacade chooser = itr.next();  itr.hasNext();)
 				{
-					itr.remove();
+					/* If we find an add bonus chooser, remove it.*/
+					if (chooser.getName().equals(LanguageBundle.getString("in_sumLangBonus")))
+					{
+						itr.remove();
+					}
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
```diff
@@ -2164,14 +2166,39 @@ void refreshLanguageList()
 				currBonusLangs.add(lang);
 			}
 		}
-		int bonusLangRemain = bonusLangMax - currBonusLangs.size();
+		int bonusLangRemain = 0;
+		if (!theCharacter.getRace().getDisplayName().equals("<none selected>"))
```
grimreaper: We generally try and avoid using display name as a method of checking in code. 
Perhaps use `theCharacter.getRace().isUnselected()` ?

code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
```diff
@@ -2164,14 +2166,39 @@ void refreshLanguageList()
...
 		numBonusLang.set(bonusLangRemain);
 		if (bonusLangRemain > 0)
 		{
+
+			boolean containsAddBonus = false;
+			/* Check to see if the chooserList already contains an "add bonus" chooser*/
+			for (LanguageChooserFacade chooser : langChoosersList)
+			{
+				/* If we find one, break, as we don't need to add another.*/
+				if (chooser.getName().equals(LanguageBundle.getString("in_sumLangBonus")))
```
grimreaper: Same here, checking the name against the translated string seems odd. Is there another way of doing this?

code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
```diff
@@ -2188,6 +2215,18 @@ void refreshLanguageList()
 		{
 			todoManager.removeTodo("in_sumTodoBonusLanguage");
 			todoManager.removeTodo("in_sumTodoBonusLanguageFirstOnly");
+
+			/* Ensure the bonus language chooser is removed, if it exists.*/
+			Iterator<LanguageChooserFacade> itr = langChoosersList.iterator();
+			while (itr.hasNext())
```
grimreaper: Let's use a enhanced forloop here?

cmhulbert: The only reason I didn't is because using an iterator allows for removing the choosers in-place. I don't know of a way to do this with an enhanced for loop. I could use a normal for loop with the iterator, but it would be of the form:

```java
for (LanguageChooserFacade chooser = itr.next();  itr.hasNext();)
{
	/* If we find an add bonus chooser, remove it.*/
	if (chooser.getName().equals(LanguageBundle.getString("in_sumLangBonus")))
	{
		itr.remove();
	}
}
```

Is that reasonable?

grimreaper: Works for me other than the existing comment about using the Key or other non translated attribute

code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
cmhulbert: So far as I can tell, there isn't another way that I have been able to find. Do you have any suggestions?

code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
cmhulbert: I agree. I didn't know of another way to check the race, but this is much more robust. Updated

code/src/java/pcgen/gui2/facade/CharacterFacadeImpl.java
grimreaper: welp. Even reading later code in the file does the same thing and I don't see a better way. Tried modifying the interface for LanguageChooserFacade but that doesn't give us much.

grimreaper:
Merging. If we find a better way always possible to change later. Thanks for the update!

cmhulbert:
So this might be for a later change, as you said, but I did have a few ideas on how do this without string matching on the translated attribute:

For one, so far as I can tell there will only ever be one bonus language chooser (if it gets removed, but there is still a bonus language remaining, I believe a new one will be added, so it will look like there is still only ever one). So one way to do it would be to store a reference in to the current bonus chooser, and just delete it when there are no bonus languages remaining.

The other option is to use the language bundle property key, instead of the string. The key is not translated, so that could be possible. The issue is that the LanguageChooserFacadeImpl does not store the key for the property, only the translated text, so if we went this route, we would need to allow the chooser itself to store the property as well.

grimreaper:
Of those two, I prefer the first approach. It keeps the translation and localization information separate from the chooser info. If you'd like to implement that it'd be great!

grimreaper:
hey @cmhulbert just letting you know that this diff causes a freezing bug reproducible by @LegacyKing  (though I don't see it). As a result we've reverted it. That said, please do consider contributing some more!

PR #5442, opened 2019-05-03
[PCGenFrame] Use JavaFx browser for "simple" HTML content.

grimreaper:


code/src/java/pcgen/gui3/SimpleHtmlPanel.java
```diff
@@ -0,0 +1,28 @@
+package pcgen.gui3;
```
thpr: Needs a copyright header

build.gradle
```diff
@@ -180,6 +186,7 @@ dependencies {
 
     compileOnly group: 'org.jetbrains', name: 'annotations', version:'17.0.0'
     compileOnly group: 'com.yuvimasory', name: 'orange-extensions', version: '1.3.0'
+    compileOnly group: 'com.github.spotbugs', name: 'spotbugs-annotations', version: '3.1.12'
```
thpr: This seems out of place in a browser swap...

grimreaper: I was going for "all new gui3 code will use the new rules..." but yeah, this should be a different, followup, change.

grimreaper:
> Also, can we drop any of the libraries with this change?

With the changes in #5408, yes.

PR #5445, opened 2019-05-04
[CDomObject] make CDomObject SortKeyAware

grimreaper:
This eliminates the need to special case CDomObject when sorting
objects.


grimreaper:
Per @thpr SKA was supposed to be a tag that says "I have a sort key that can override my default sorting mechanism" and not more general sorting. Closing to keep this

PR #5447, opened 2019-05-04
PreFactTester: convert Iterable to Collection

grimreaper:
While I generally prefer weaker types, only Collection is easily
convertible to a Stream. By using streams we can simplify
PreFactSetTester, PreFactTester, and several others.

### DIFF
--- a/code/src/java/plugin/pretokens/test/PreFactSetTester.java
+++ b/code/src/java/plugin/pretokens/test/PreFactSetTester.java
@@ -17,9 +17,8 @@
  */
 package plugin.pretokens.test;
 
-import java.util.List;
+import java.util.Collection;
 
-import pcgen.base.util.Indirect;
 import pcgen.cdom.base.CDOMObject;
 import pcgen.cdom.base.Reducible;
 import pcgen.cdom.enumeration.FactSetKey;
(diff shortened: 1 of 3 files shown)


code/src/java/plugin/pretokens/test/PreFactSetTester.java
```diff
@@ -72,26 +71,17 @@ public int passes(final Prerequisite prereq, final PlayerCharacter aPC, CDOMObje
...
-				if (indirect.get().equals(targetVal))
-				{
-					runningTotal++;
-					continue CDO;
-				}
-			}
-		}
+		int runningTotal = (int) objModel.stream()
+		                                 .map(r -> r.getCDOMObject().getSetFor(fk))
+		                                 .filter(sets -> sets.stream()
```
thpr: Wow, stream in a stream - is there anything here that can be broken out into a separate primitive or method to make it clearer?

code/src/java/plugin/pretokens/test/PreFactSetTester.java
```diff
@@ -72,26 +71,17 @@ public int passes(final Prerequisite prereq, final PlayerCharacter aPC, CDOMObje
...
-			{
-				if (indirect.get().equals(targetVal))
-				{
-					runningTotal++;
-					continue CDO;
-				}
-			}
-		}
+		int runningTotal = (int) objModel.stream()
+		                                 .map(r -> r.getCDOMObject().getSetFor(fk))
```
karianna: `r` and `fk` can we get more descriptive variable names?

grimreaper:
low importance PR with open issues.

PR #5451, opened 2019-05-04
Equipment: remove dead code

grimreaper:


code/src/java/pcgen/core/Equipment.java
```diff
@@ -6258,15 +6071,15 @@ public CDOMObject getLocalChild(String childType, String childName)
...
+		return Collections.singletonList("EQUIPMENT.PART");
 	}
 
 	@Override
 	public List<PCGenScoped> getChildren(String childType)
 	{
 		if ("EQUIPMENT.PART".equals(childType))
 		{
-			return new ArrayList<>(heads);
+			return Collections.unmodifiableList(heads);
```
thpr: This is a functional change - and we need to be careful with these... there are places where items return ArrayList because it's transferring ownership of the returned list

PR #5453, opened 2019-05-05
NEWTAG-627: Add getDatasetID to PCGenIdentifier interface

grimreaper:
This is in preparation for  ResultFacet and VariableStoreFacet taking a
DatasetID instead of a CharID

### DIFF
--- a/code/src/java/pcgen/cdom/base/PCGenIdentifier.java
+++ b/code/src/java/pcgen/cdom/base/PCGenIdentifier.java
@@ -17,6 +17,8 @@
  */
 package pcgen.cdom.base;
 
+import pcgen.cdom.enumeration.DataSetID;
+
 /**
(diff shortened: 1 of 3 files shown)


code/src/java/pcgen/cdom/base/PCGenIdentifier.java
```diff
@@ -17,12 +17,14 @@
...
+
 /**
  * This interface represents an identifier (like a CharID) so that certain classes can use
  * Generics to have a minimal compile requirement of using an expected identifier rather
  * than an arbitrary object.
  */
 public interface PCGenIdentifier
 {
-	//Intentionally Empty Interface
+	public DataSetID getDataSetID();
```
thpr: Method needs docs :)

grimreaper: better?

PR #5455, opened 2019-05-05
npcgen: remove it

grimreaper:
- this code is never used or called.
- it uses JEP which will need to be modified eventually


thpr:
This needs discussion

karianna:
> * this code is never used or called.
> * it uses JEP which will need to be modified eventually

You sure it's not being used?

grimreaper:
> > * this code is never used or called.
> > * it uses JEP which will need to be modified eventually
> 
> You sure it's not being used?

It has no UI, and removing the code has no impact. How else can it be used?

@thpr re discussion: sure. I could start a thread on the ML

grimreaper:
sent a mail to the mailing list

LegacyKing:
Approved - As far as my opinion counts

grimreaper:
https://pcgen.groups.io/g/pcgendevelopers/topic/removal_of_npcgen/31513007?p=,,,100,0,0,0::recentpostdate%2Fsticky,,,100,2,0,31513007

I'll send an email to users as well

PR #5456, opened 2019-05-05
reduce reliance on commons-io

grimreaper:
This is useful since most of what we're using commos-io
for is actually availble in native jdk

### DIFF
--- a/code/src/java/pcgen/io/PCGIOHandler.java
+++ b/code/src/java/pcgen/io/PCGIOHandler.java
@@ -24,12 +24,14 @@ import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileOutputStream;
+import java.io.FileReader;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.math.BigDecimal;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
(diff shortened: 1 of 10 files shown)


code/src/java/pcgen/io/PCGIOHandler.java
```diff
@@ -633,7 +615,7 @@ private SourceSelectionFacade internalReadSources(InputStream in)
 		// Verify it is ver2
 		boolean isPCGVersion2 = isPCGCersion2(lines);
 
-		final String[] pcgLines = lines.toArray(new String[lines.size()]);
+		final String[] pcgLines = lines.toArray(new String[0]);
```
thpr: Why lose the known length information and force 2 array creations rather than one?

code/src/java/pcgen/system/BatchExporter.java
```diff
@@ -390,13 +381,7 @@ public static boolean exportPartyToNonPDF(PartyFacade party, File outFile, File
...
 			party.export(new ExportHandler(templateFile), bw);
 			return true;
-		}
-		catch (final UnsupportedEncodingException e)
-		{
-			Logging.errorPrint("Unable to create output file " + outFile.getAbsolutePath(), e);
-			return false;
-		}
-		catch (final IOException e)
+		} catch (final IOException e)
```
thpr: Curly brace standard violation

grimreaper: we really need to add this to some automated tool

code/src/java/pcgen/io/PCGIOHandler.java
grimreaper: In pre-Java-6 versions using a presized array was better, as the reflection call  to properly size thw array slow. However since Kava 6 this call was intrinsified, making the performance of the empty array version the same and sometimes even better

PR #5472, opened 2019-05-05
Working PR for SF Race/Template Sizes

LegacyKing:


LegacyKing:
This requires the RACE Scope to be merged first.

PR #5478, opened 2019-05-06
[build] update comment for broken spotbugs config

grimreaper:
The given link probably gives enough information to fix, but I'm tired
so start by just linking it :)


grimreaper:
going to just fix directly

PR #5479, opened 2019-05-06
[build] move gradle includes to buildSrc

grimreaper:
These have a slightly different semantic (getting included directly
rather than with an isolated classpath). That said, it seems closer to
what we were actually expecting.


grimreaper:
I forgot `git add` on some files. Will fix and refile

PR #5482, opened 2019-05-07
CODE-3388 Fix: Semantics based on context of a formula

thpr:
Fix: Load Process (MODIFY/MODIFYOTHER) does not set up for built-in local functions


grimreaper:
I read through the code and see nothing obvious wrong. That said, this part of the code i have the last understanding of. Merging since it seems "not wrong" and you're really the best reviewer, but hey - progress!

PR #5483, opened 2019-05-07
patch to allow non-number endings, i.e. RCx

LegacyKing:


grimreaper:
@LegacyKing I modified this PR to also fix some syntax. That said, LGTM.

LegacyKing:
@grimreaper the original code WORKED for the RC5 release. I don't understand the syntax, you'd have to ask @javydreamercsw as he supplied it. I'm just putting it here so we can include it.

I cannot merge it if you've altered the syntax and no time to test if it breaks the release.

grimreaper:
I believe it worked for RC5. I'm worried it'll break once the version stops being "RC".
Perhaps @javydreamercsw could comment?

javydreamercsw:
It'll basically work for anything after the number.

Basically it tokenizes with the period. So 6.00.00XX

Turns into 6 00 00XX. For the last piece it does a substring of the first 2 characters.

code/gradle/release.gradle
```diff
@@ -215,7 +214,6 @@ task downloadJRE doLast {
                         fileTree(dir: "${projectDir}/jre/${os}/").include("*.${extension}").each { simLib ->
                             copy {
                                 println "Unzip $simLib.name"
-                                def name = simLib.name - '.$extension'
```
javydreamercsw: I thought this was used somewhere else. Maybe it was just an output no longer there.

This should be easy to test. Just delete the JRE folder and call this task.

code/gradle/release.gradle
```diff
@@ -110,8 +109,8 @@ ext {
     }
 
 	List result = plainVerNum.replaceAll('-DEV', '').tokenize('.')
-	installerVerNum = result.get(0).toInteger() + "." + result.get(1).toInteger() + "." +  result.get(2).toInteger() + ".0"
-}
+	installerVerNum = result[0].toInteger() + "." + result[1].toInteger() + "." + result[2].take(2).toInteger() + ".0"
+	}
```
javydreamercsw: This should be still correct.

PR #5484, opened 2019-05-07
[gui] make new gui3 code stricter w.r.t. code standards

grimreaper:
- Make non-null the default


build.gradle
```diff
@@ -186,6 +186,7 @@ dependencies {
 
     compileOnly group: 'org.jetbrains', name: 'annotations', version:'17.0.0'
     compileOnly group: 'com.yuvimasory', name: 'orange-extensions', version: '1.3.0'
+    compileOnly group: 'com.github.spotbugs', name: 'spotbugs-annotations', version: '3.1.12'
```
LegacyKing: Is spotbugs required?

grimreaper: Its a compile only dependencies - users won't see it.

PR #5485, opened 2019-05-07
[build] remove now unneeded dependencies

grimreaper:
With the conversion to JavaFX we longer need cobra or rhino.


grimreaper:
depends on #5486

PR #5487, opened 2019-05-07
[*] Fix some typing & generics issues

grimreaper:
- Add types to some places where it'd help
- Enforce immutability in some places where it'd help.


grimreaper:
will be replaced with a different branch

PR #5489, opened 2019-05-07
[gui3] use FXML instead of creating nodes in code

grimreaper:
This continues to more fully decouple code from UI. Further it makes the
notion of a controller "real" in pcgen.

Long term we may even have people who edit the UI that don't need code
permissions or info.

While this is a trivial example, I'm trying to figure out the right
pattern for UI in pcgen.

### DIFF
--- a/code/src/java/pcgen/gui2/PCGenFrame.java
+++ b/code/src/java/pcgen/gui2/PCGenFrame.java
@@ -101,7 +101,8 @@ import pcgen.gui2.tools.CharacterSelectionListener;
 import pcgen.gui2.tools.Icons;
 import pcgen.gui2.tools.Utility;
 import pcgen.gui2.util.ShowMessageGuiObserver;
-import pcgen.gui3.SimpleHtmlPanel;
+import pcgen.gui3.JFXPanelFromResource;
+import pcgen.gui3.SimpleHtmlPanelController;
 import pcgen.io.PCGFile;
 import pcgen.persistence.SourceFileLoader;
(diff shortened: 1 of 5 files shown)


code/src/java/pcgen/gui2/PCGenFrame.java
```diff
@@ -1827,7 +1829,7 @@ private void showMatureDialog(String text)
 
 		aFrame.getContentPane().setLayout(new BorderLayout());
 		aFrame.getContentPane().add(jPanel1, BorderLayout.NORTH);
-		aFrame.getContentPane().add(htmlPanel, BorderLayout.CENTER);
+//		aFrame.getContentPane().add(htmlPanel, BorderLayout.CENTER);
```
thpr: Should we remove this rather than comment out?

code/src/resources/pcgen/gui3/SimpleHtmlPanel.fxml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thpr: Do we need (c) items in these files as well?

code/src/java/pcgen/gui2/PCGenFrame.java
grimreaper: Neither. It shouldn't have been commented out :)

grimreaper:
> This has duplication from the splash screen conversion, I foresee conflicts.

They are actually a chain of PRS - each one depends on the next.

PR #5490, opened 2019-05-07
[gui3] Convert JLabelPane to SimpleHtmlPanel

grimreaper:
It seems similar enough. This change loses
- some sizing information
- marking the tip as "hyperactive"

but both seem easily fixable.


thpr:
This seems to include PR #5489, so is very confusing - and has the same questions/issues

LegacyKing:
The Splash Screen PR#5491 is the same... all have the exact same files.

grimreaper:
> This seems to include PR #5489,

Yeah, it depends on #5489 and is likely better reviewed once rebased.

PR #5491, opened 2019-05-08
[gui3] SplashScreen: Convert splash screen to JavaFX

grimreaper:
This is starting to set a pattern w.r.t. how UI is going to work. This 
pattern seems like it will work well so long as we have mixed JavaFX and 
Swing code. Once we only have JavaFX we'll need to rejigger parts of the 
high level structure but the main code should Just Work


code/src/java/pcgen/gui2/dialog/TipOfTheDay.java
```diff
@@ -132,10 +135,9 @@ private void initUI()
 		FontManipulation.xxlarge(lblDidYouKnow);
 		lblDidYouKnow.setOpaque(true);
 
-		tipText = new JLabelPane();
 		tipText.setBorder(null);
 		tipText.setFocusable(false);
-		tipText.addHyperlinkListener(new Hyperactive());
+//		tipText.addHyperlinkListener(new Hyperactive());
```
LegacyKing: We plan on using this? Otherwise commented out code is a best practice no-no and just accrues code debt.

code/src/java/pcgen/gui3/JFXPanelFromResource.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Displays HTML content as a "panel".
+ */
+public final class JFXPanelFromResource<T> extends JFXPanel
+{
+
+	private final FXMLLoader fxmlLoader = new FXMLLoader();
+
+	public JFXPanelFromResource(Class<T> klass, String resourceName)
```
thpr: How permanent are these classes?  How much Javadoc should we be doing for them as a result of the first question?

code/src/java/pcgen/gui3/preloader/PCGenPreloader.java
```diff
@@ -0,0 +1,93 @@
...
+import pcgen.system.PCGenTaskListener;
+import pcgen.system.ProgressContainer;
+import pcgen.util.Logging;
+
+import javafx.application.Platform;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.Scene;
+import javafx.stage.Stage;
+
+public class PCGenPreloader implements PCGenTaskListener
```
thpr: Same on Javadoc

code/src/resources/pcgen/gui3/SimpleHtmlPanel.fxml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
thpr: Same on (c) as other PRs

code/src/java/pcgen/gui2/dialog/TipOfTheDay.java
grimreaper: This is actually more of a TODO. Agreed that commented code is bad, but in this case its not unreasonable.

code/src/java/pcgen/gui3/JFXPanelFromResource.java
grimreaper: Fairly permanent. They'd only go away once 90-100% of the code is converted. Will add docs.

PR #5511, opened 2019-05-10
Replace some JFileChooser with FileChooser (swing -> javafx)

grimreaper:
This doesn't change all JFileChooser usage, but many of them. In the 
meantime we'll have an inconsistent UX for selecting files.


thpr:
#5491 should be merged first

PR #5515, opened 2019-05-11
[resources] move existing resources to new resource directory

grimreaper:
This allows IDEs to view them as resources and have a proper resource directory.
It doesn't yet bring us away from absolute resource paths, but closer.


grimreaper:
delay this one for a bit, I need to fix gmgen

PR #5550, opened 2019-05-12
OptionsPathDialog: convert to JavaFX

grimreaper:


grimreaper:
... and right after submitting I see the problem: I'm on the event loop when creating and then showing the dialog

PR #5566, opened 2019-05-13
CODE-3401: Make Resizable Types Type-Safe

thpr:


code/src/java/pcgen/core/Globals.java
```diff
@@ -593,7 +594,7 @@ public static boolean canResizeHaveEffect(final Equipment aEq, List<String> type
 		}
 
 		final List<String> resizeTypeList = SettingsHandler.getGame().getResizableTypeList().stream()
-			.map(String::toUpperCase).collect(Collectors.toList());
+			.map(Type::toString).map(String::toUpperCase).collect(Collectors.toList());
```
grimreaper: nit: we were chomping down on streams calls right (new lines before each ".")

PR #5577, opened 2019-05-13
Remove unused code from several classes

grimreaper:
This PR should be rebase-merged and not squash-merged.


thpr:
This is almost guaranteed to break local branches that I have - some of
this is good, some needs to be kept for a bit and/or validated for purpose.

Tom

grimreaper:
I split this into many PRs.

thpr:
Yea, the problem being things that will break my branches are being
merged... so this is now going to churn my time fixing all of those
branches.

I was hoping to have these contents held up until I could give more
feedback on which are good and which are a problem.

As it is, they are being merged because the impact is not understood...
please expect me to revert a number of items... and for future reference
things in cdom. Or context. Should probably not be changed without
talking... they are rarely unused completely.

Tom

PR #5623, opened 2019-05-16
Remove unused from Type

grimreaper:


grimreaper:
this is actually broken

PR #5629, opened 2019-05-16
Eliminate "point pool" code

grimreaper:
This adds a fair amount of complexity, but appears to be completely unused.


karianna:
Isn't this where people use point allocation for their characters?

grimreaper:
I need to look closer: I thought this code wasn't used but need to more fully understand itl.

PR #5641, opened 2019-05-17
pmd: start checking OptimizableToArrayCall

grimreaper:


grimreaper:
This was was created nearly entirely automatically. 
The two exceptions are:
- the actual change to ruleset.xml
- the change to SortableTableRowSorter

since this will cause conflicts to other PRs and I don't want to stall, merging without review.

PR #5645, opened 2019-05-17
[*] remove default template comment on class

grimreaper:


grimreaper:
this one is simple enough to merge without review

PR #5664, opened 2019-05-18
ScanForUnusedIl8nKeys: re-enable test

grimreaper:
...


code/src/testResources/pcgen/lang/cleaned.properties
```diff
@@ -93,16 +90,14 @@ in_InfoPrepared=Prepared Spells
...
-
 in_Templates=Templates
 
 #Other Universal stuff
 in_sourceLabel=Source
 
 #Other Universal stuff
 
-in_template=Template
+in_condition=Condition
```
LegacyKing: Was two entries, and now it's one?

grimreaper: We don't have `in_template` at all anymore

PR #5672, opened 2019-05-18
HtmlSheetSupport: convert to SimpleHtmlPanel

grimreaper:


code/src/java/pcgen/gui2/InfoGuidePane.java
```diff
@@ -122,16 +122,16 @@ private void initComponents()
...
-			boolean show = e.getNewReference() == null;
-			if (show)
-			{
-				tipPane.getController().setHtml(
-						LanguageBundle.getFormattedString(
-								"in_si_tip",
-								TipOfTheDayHandler.getInstance().getNextTip()
-						));
-			}
+			frame.getSelectedCharacterRef().addReferenceListener(e -> {
```
thpr: Why the indentation change?

code/src/java/pcgen/gui2/tabs/models/HtmlSheetSupport.java
```diff
@@ -118,7 +105,7 @@ public void refresh()
 		}
 		if (refresher != null && !refresher.isDone())
 		{
-			refresher.cancel(false);
+			refresher.cancel(true);
```
thpr: Is this an embedded bug fix, or ?

grimreaper: yes. this is a bug fix. or at least a behavior change that I think is better. Will make it a separate PR for better future debugging.

code/src/java/pcgen/gui2/InfoGuidePane.java
grimreaper: bug. fixed.

grimreaper:
This needs to be recreated.

PR #5690, opened 2019-05-20
[*] Simplify stream callchains

grimreaper:


code/src/java/plugin/pretokens/test/PreSkillTester.java
```diff
@@ -167,10 +167,10 @@ else if (isType && !foundSkill)
 					else if ((percentageSignPosition >= 0))
 					{
 						List<Type> mockTypes = target.getTrueTypeList(true);
-						for (Type mockType : mockTypes)
+						for (Type ignored : mockTypes)
```
thpr: Is this is a bug now semantically being hidden due to an ignore?

karianna: Yeah the `mockType` or `ignored` being unused looks like a bug

grimreaper:
conflicts and not important PR

PR #5691, opened 2019-05-20
Add 2 unit tests for SF

LegacyKing:


LegacyKing:
I don't know why this fails. @grimreaper you able to decipher?

grimreaper:
looks like its passing now?

PR #5709, opened 2019-05-22
PMD: start checking UnusedLocalVariable

grimreaper:


code/src/java/pcgen/core/PCClass.java
```diff
@@ -434,10 +434,9 @@ public void setLevel(final int newLevel, final PlayerCharacter aPC)
 			SubstitutionClassApplication.checkForSubstitutionClass(this, newLevel, aPC);
 		}
 
-		for (PCClass pcClass : aPC.getClassSet())
-		{
-			aPC.calculateKnownSpellsForClassLevel(this);
-		}
+		aPC.getClassSet().stream()
```
thpr: This feels like it's hiding a bug.  ignored -> this??

karianna: I think the map can be removed?

grimreaper: oh, you're right.

LegacyKing:
Is the intent here you remove unused variables, along with enabling the rule?

grimreaper:
> Is the intent here you remove unused variables, along with enabling the rule?

Precisely: enable rule, fix reported issues, and fix any bugs caught as a result.

PR #5730, opened 2019-05-23
Main: construct dialog only in "use gui" case

grimreaper:
JFXPanel can't be constructed in the headless case. Avoid doing so.

This allows int tests to run, but exposes that they have been failing
for some time.


LegacyKing:
If this fixes the Unit tests, then I'm happy. We can fix breaking tests.

PR #5743, opened 2019-05-24
DesktopHandler: handle unsupported platforms

grimreaper:
- test for "supported" before using a feature
- avoid platform specific names
- move to gui3 since it is a "converted" class


LegacyKing:
Patch works. PCGen now loads correctly on Windows machines.

LegacyKing:
However, we have additional issues upon source load and UI distortion.

Issue #5762, opened 2019-05-26
Extra feat when Disenchanter Archetype is chosen for Warpriest 


Muzouka:
When the Disenchanter Archetype is chosen for the Warpriest a Disenchanter Bonus feat shows up in Feats & Abilities.
![image](https://user-images.githubusercontent.com/36632341/58377187-33677a00-7fad-11e9-91ce-84dfde024388.png)

LegacyKing:
Fixed. Opening JIRA to track.

PR #5763, opened 2019-05-26
Fix bug: DATA-3947 - Disenchanter Archetype for Warpriest granting in…

LegacyKing:
…correct Bonus Feat progression.


LegacyKing:
Fix for #5762

PR #5766, opened 2019-05-26
FlippingSplitPane: reduce differences between two versions

grimreaper:
We have two copies of FlippingSplitPane - both of which are nearly
identical but have different implementations (sometimes due to my
cleanup). In prep for combining them, and then converting them to
JavaFX, reduce the difference between the two.


grimreaper:
Resolved in #5771

PR #5778, opened 2019-05-27
Move pcgen frame alerts to JavaFX thread

grimreaper:
Long term we'll need to be better about ensuring functions get called on
the right thread. In the short-medium term we'll be crossing between
swing and Java fairly often,


grimreaper:
This is a simple enough change that unbreaks master. Happy to follow up on comments later

PR #5779, opened 2019-05-27
Remove "invalid option text" preference

grimreaper:
There is limited reason to make this a preference: it merely changes the
text that gets printed when an option is not available. Further, it
complicates some of the options dialog, so remove it.


grimreaper:
FYI @LegacyKing  this affects output sheets

LegacyKing:
I need this to be exposed to data so I can address this please.

PR #5790, opened 2019-05-27
AboutDialog: remove l2f license

grimreaper:
We no longer distribute L2F, so remove the license from the dialog


grimreaper:
need to fix this moere - and am adding a test

PR #5802, opened 2019-05-29
ExportDialog: pull file searching out of thread

grimreaper:
Finding templates isn't particularly expensive and makes following the
logic for how this dialog gets initialized confusing.

### DIFF
--- a/code/src/java/pcgen/gui2/dialog/ExportDialog.java
+++ b/code/src/java/pcgen/gui2/dialog/ExportDialog.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/gui2/dialog/ExportDialog.java
```diff
@@ -641,34 +630,20 @@ protected void done()
 			return Files.walk(dir.toPath())
 			            .filter(f -> !f.endsWith(".fo"))
 			            .map(Path::toFile)
-					    .collect(Collectors.toList());
-		}
-
-		@Override
-		protected void done()
+			            .collect(Collectors.toList());
+		} catch (IOException e)
```
thpr: Curly Brace on own line...

grimreaper: #5813 should resolve these types of comments.

PR #5809, opened 2019-05-29
WriteDirectoryPanel: use .walk instead of walking manually

grimreaper:


code/src/java/pcgen/gui2/converter/panel/WriteDirectoryPanel.java
```diff
@@ -191,42 +190,35 @@ private void showWarning()
...
-				{
-					findPCCFiles(file, existingFiles);
-					continue;
-				}
-				existingFiles.add(file);
-			}
+			return Files.walk(aDirectory.toPath())
+			            .filter(file -> file.getFileName().toString().endsWith("pcc"))
+			            .collect(Collectors.toUnmodifiableList());
+		} catch (IOException e)
```
thpr: Curly Brace gets its own line...

code/src/java/pcgen/gui2/converter/panel/WriteDirectoryPanel.java
```diff
@@ -191,42 +190,35 @@ private void showWarning()
...
-			{
-				if (file.isDirectory())
-				{
-					findPCCFiles(file, existingFiles);
-					continue;
-				}
-				existingFiles.add(file);
-			}
+			return Files.walk(aDirectory.toPath())
+			            .filter(file -> file.getFileName().toString().endsWith("pcc"))
```
thpr: One of the earlier file -> file.blah items also did .toLowerCase.... do we intend to be case sensitive here?

Also, it would be interesting to see if some of these filters could be shared at some point (shouldn't hold up this PR)

grimreaper: yeah, long term we should pull these out into a PCGenFileFilters holder class or something similar.

PR #5833, opened 2019-06-01
checkstyle: actually enforce things

grimreaper:
I had forgotten that we set the severity to "warning" in a previous
effort to clean up.


grimreaper:
Merging this since it unblocks some other PRs and the broad idea of building-up the erroring-out has consensus.

PR #5838, opened 2019-06-01
PCGen System: implement deadlock detector

grimreaper:


karianna:
VisualVM / Netbeans comes with this, why do we need to build one ourselves?

grimreaper:
Allows us to get reports from users/testers rather than require people to hook up a separate debugger or profiler.

PR #5852, opened 2019-06-03
OutputPanel: fix squashed display

grimreaper:
This broke when a setting was removed but constraints were not reset
correctly.


grimreaper:
minor UI bugfix. merging.

PR #5856, opened 2019-06-03
PreferencesDialog: downgrade to JSplitPane

grimreaper:
Much like in 8ea81b0b03f353887a0573fd4a66a25dac8daab8, while
FlippingSplitPane has some nice utility, it isn't required for a
preferences dialog.


grimreaper:
two liner and helpful for the next step: merging pref panes.

PR #5885, opened 2019-06-09
Fix bug: DATA-3956 - [STARFINDER] Feat "Fleet" not adding movement bo…

LegacyKing:
…nus.


LegacyKing:
@bigrin42 This is how a Master branch should appear in a review as an FYI.

PR #5902, opened 2019-06-13
refactored recursive code to not recreate the same object on each cal…

kypvalanx:
what i have done here is refactored this code a bit to remove unnessesary item creation from a recursive method and also remove an if check that can only be true on the first call in the recursive stack.

based on my measurements it reduces load time from 12 to 9 seconds

also note that i have named a method _findFiles.  I did this because i've seen it used to denote a private method that is only to be called by the method of the same name.  i can rename it.


grimreaper:
Nice performance improvement

PR #5918, opened 2019-06-22
Added GM Award to grant a bonus skill point per level.

GrimaceJD:
Just a new GM award option.


GrimaceJD:
Going to repull due to fork issues.

PR #5928, opened 2019-06-29
addressed Enum internal switching code smell

kypvalanx:
classes that switch on enums are possible candidates for converting to inheritance models.  i have done so here since this class in question appears to have 2 very different, independent, workflows.


kypvalanx:
fixing

PR #5960, opened 2019-07-23
DATA-4001: Fixed redundant Acrobatics/Jump situation

GrimaceJD:
Fixed outliers to use "Jump" situation for Acrobatics and commented out "When Jumping" situation.


LegacyKing:
@Llisandur is this backwards?  Shouldn't the "Jump" be the one to be removed?
@GrimaceJD what is the bonus called on the CR stuff?

Llisandur:
"When Jumping" actually matches most Bestiary entries. "Acrobatics (Jump)" is a remnant from the original dataset being ported from the 3,5e SRD.

PR #5962, opened 2019-07-23
[HH] Added descriptions and fixed nursing necklace issue due to misprint

GrimaceJD:
That is all.


LegacyKing:
@Nylanfs can you give this a quick glance? Merged, but just need to make sure this is still in license. Thanks!

PR #5972, opened 2019-08-01
Revert "PcgenFtlTestCase: stop deleting output file (#5842)"

Zaister:
This reverts commit 1025ae8c. The commit added syntactially incomplete and broke the build.

Signed-off-by: Stefan Radermacher <stefan@zaister.de>


grimreaper:
How is this broken? I'm not sure I understand

Zaister:
Check out the code from that commt, there seems to be something missing, the code is incomplete, not even that braces are balanced - the try block just ends abruptly at the brace closing the class.

LegacyKing:
@Zaister One of the character unit tests is failing - nodes dropped from 115 to 113

LegacyKing:
I think we can close as we have a fix?

starcraftman:
I assume so, no longer relevant.

PR #5973, opened 2019-08-01
Added temporary fix for build error 

TotallyCoded:
This pull request add `System.load("C:\\Windows\\System32\\WindowsCodecs.dll");` to 2 of the tests as a temporary fix to https://github.com/javafxports/openjdk-jfx/issues/66. This was causing `gradlew` builds to fail on my side with `EXCEPTION_ACCESS_VIOLATION` during running `Task :test`.

This commit need to be tested on linux and mac machines as it could cause builds to fail on them.

### DIFF
--- a/code/src/utest/pcgen/gui3/dialog/AboutDialogTest.java
+++ b/code/src/utest/pcgen/gui3/dialog/AboutDialogTest.java
@@ -41,6 +41,7 @@ class AboutDialogTest
 	@Start
 	private void Start(Stage stage) throws IOException
 	{
+		System.load("C:\\Windows\\System32\\WindowsCodecs.dll");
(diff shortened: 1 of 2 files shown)


TotallyCoded:
The travis build failed confirming what I suspected, that it would fail to build on linux machines. Not sure what action to go from here as it fixes some builds on windows but breaks other builds?

TotallyCoded:
This fix should be temporary until JavaFX 13 is released which is a permanent fix.

grimreaper:
almost there. mind adding a comment linking to the upstream bug (or at least back to this comment) so when JavaFX 13 comes out we'll know to remove it.

grimreaper:
Thank you!

PR #5977, opened 2019-08-05
Fixed clean running after build deleting build folder

TotallyCoded:
This ensures the clean never runs after build if called in the same dependsOn such as in prepareRelease. 

`task prepareRelease (dependsOn: ['clean', 'build']) {` Without the `shouldRunAfter` (this Pr addition) the build and clean could run in either order. Gradle just cared about them running before `prepareRelease` which often led to the situation of clean running after build and deleting the build folder.

~~I am using `shouldRunAfter` rather than `mustRunAfter` so that clean and build can be run independently as well.~~ - This was incorrect.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -472,3 +472,5 @@ apply from: 'code/gradle/distribution.gradle'
 apply from: 'code/gradle/reporting.gradle'
 apply from: 'code/gradle/release.gradle'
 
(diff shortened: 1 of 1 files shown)


TotallyCoded:
I misunderstood how mustRunAfter and shouldRunAfter worked and looking at it further I should be using mustRunAfter as it still allows for the tasks to be called independently but it both are run it forces the order which is what we want.
I'll push a new commit with this fixed soon.

PR #5987, opened 2019-08-17
Fix CODE-3451: Simply remove redundant dialog.

starcraftman:
https://pcgenorg.atlassian.net/browse/CODE-3451

The fix is simple, I raised some questions in the solution section of 3451 that may need to be discussed before or after pulling.


karianna:
Build fails...

starcraftman:
checkstyleMain does not seem to run with standard gradle check task. Removed unused imports.

PR #6000, opened 2019-09-03
[bio] remove show/hide checkboxes

grimreaper:
This adds complexity to the UI and is one of two special places where we
handle have-this-but-don't-show-it.


karianna:
@grimreaper merge conflicts

PR #6002, opened 2019-09-03
FIX CODE-3452: Staggering Fist feat error

starcraftman:
- Corrected BAB requirement to 2.

-----------------------------
Simple fix, of note this also affects the 6.08x branch. Cherry pick would be trivial.


starcraftman:
I see, should have used the move option on JIRA right? Will do in future. Makes sense.

PR #6010, opened 2019-09-09
FIX GITHUB-5990: Loading Source from Existing Character

### LINKED ISSUE #5990: Loading Source from Existing Character
Not sure if this is a bug, I'm on v6.08.00 RC7.
I can load character files, and the sources load with them, but they don't show up in the `Sources > Select Sources` Menu.
Basically, I want to add a new source that I missed.
Is this possible without repeating every one previously chosen manually?

starcraftman:
- When a character loaded and advanced source selections brough up,
always present the character's source selections. Allows easy editting
of sources.

--------------------
Easy feature, fixes #5990. Is fairly handy and took no effort. Not a huge priority for backporting but it would be convenient.


starcraftman:
Don't merge just yet, making some changes.

starcraftman:
Hmmmm, think I need to consider this a bit more. It occurs that this would always clobber any manual changes post load, will consider it and make PR again later.

PR #6076, opened 2019-11-04
Fix  DATA-4011    Psychic Detective uses Bard known spells.  

orangepeelbeef:
Like other spellcasters, a psychic detective can cast
only a certain number of spells of each spell level per
day. She knows the same number of spells and receives
the same number of spells slots per day as a bard of her
investigator level, and knows and uses 0-level knacks as a
bard uses cantrips. In addition, she receives bonus spells
per day if she has a high Intelligence score (see Table 1–3
on page 17 of the Core Rulebook).


data/pathfinder/paizo/roleplaying_game/occult_adventures/oa_classes.lst
```diff
@@ -192,23 +192,23 @@ CLASS:Homunculus Companion	STARTSKILLPTS:2
...
-12	CAST:0,5,5,4,3
-13	CAST:0,5,5,4,3,1
-14	CAST:0,5,5,4,4,2
-15	CAST:0,5,5,5,4,3
-16	CAST:0,5,5,5,4,3,1
-17	CAST:0,5,5,5,4,4,2
-18	CAST:0,5,5,5,5,4,3
-19	CAST:0,5,5,5,5,5,4
-20	CAST:0,5,5,5,5,5,5
+1	CAST:0,1            KNOWN: 4,2
```
LegacyKing: Space between KNOWN and the value is not good. Also, I believe you need a MEMORIZE:NO tag on the Class Line to make the KNOWN activate.

PR #6080, opened 2019-11-04
DATA-4042: Region settings not working for Goblins of Golarion.   

orangepeelbeef:
* PREREGION does not need 1,     
* Shackles region is actually The Shackles


orangepeelbeef:
This requires sources:
* Pathfinder core
* Advanced Players Guide
* Inner Sea World Guide 
* Goblins of Golarion

PR #6092, opened 2019-11-17
Create PCG_SAVE_PATH after selecting options.ini location

orangepeelbeef:
This is not the best fix, I just set it to open the current folder instead of the PCG_SAVE_PATH.  

It seems like we really need to actually create the preset directory paths during the initial load if they don't exist.   

When you first open the app it asks which directory will be your pcgen directory, then it looks for a characters directory in the path of whatever you selected, but nothing actually creates those directories.  So when you go to open the path, java throws a path not found behind the scenes but the user gets no indication.

### DIFF
--- a/code/src/java/pcgen/system/Main.java
+++ b/code/src/java/pcgen/system/Main.java
@@ -280,7 +280,7 @@ public final class Main
 	public static void loadProperties(boolean useGui)
 	{
 		if ((settingsDir == null)
-			&& (ConfigurationSettings.getSystemProperty(ConfigurationSettings.SETTINGS_FILES_PATH) == null))
+				&& (ConfigurationSettings.getSystemProperty(ConfigurationSettings.SETTINGS_FILES_PATH) == null))
 		{
 			if (!useGui)
 			{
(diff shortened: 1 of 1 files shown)


code/src/java/pcgen/system/Main.java
```diff
@@ -297,10 +297,26 @@ public static void loadProperties(boolean useGui)
...
+		if (!savepath_dir.exists() && !savepath_dir.isDirectory())
+		{
+			try
+			{
+				Logging.log(Level.INFO, "Making directory " + savepath_dir);
+				savepath_dir.mkdir();
+			}
+			catch (Exception e)
+			{
+				e.printStackTrace();
```
karianna: Prefer logging rather than printing a stack trace.  This could also use try with resources I suspect.

LegacyKing:
FYI @orangepeelbeef 

> Task :pmdMain
/home/travis/build/PCGen/pcgen/code/src/java/pcgen/system/Main.java:317:	Avoid printStackTrace(); use a logger call instead.
> Task :pmdMain FAILED

orangepeelbeef:
Unfortunately I don't really understand the try-with-resources syntax, but I did swap it over to a SEVERE error.

karianna:
For future reference - https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

PR #6099, opened 2019-11-24
[build] bump javaFX to 13.0.1

grimreaper:


LegacyKing:
Is there a reason we are updating JavaFX beyond the supported version of 11?

grimreaper:
> Is there a reason we are updating JavaFX beyond the supported version of 11?

Since Java 11, JavaFX and Java have different release and versioning schemes. this was part of the intent of splitting javafx instead its own component.

PR #6100, opened 2019-11-24
[build] bump gradle to 5.6.4

grimreaper:
This is not the latest version. Gradle 6.0 and 6.0.1 has issues
resolving jep.


LegacyKing:
We are having serious issues with the build, and I think the Gradle changes with various versions are not helping. I'm all for the keeping things current, but it's months before we get the complaints about Mac not working and Linux not working. I'd ask that we get better code test coverage that makes sure we are delivering working product/code instead of breaking the end-user experience. PCGen 6.09.x line is unusable for the community. Only the 6.08.x line is viable.

grimreaper:
>  I'm all for the keeping things current, but it's months before we get the complaints about Mac not working and Linux not working.

I am able to build reliably on macOS and on Linux. I can't test Windows though. Can you point me to build errors? That seems serious and I need to fix

PR #6102, opened 2019-11-24
[gui] fix debug dialog threading

grimreaper:
Currently loading debug dialog fails since it is on the wrong thread.
Fix this, and add more assertions to catch this earlier


code/src/java/pcgen/gui2/PCGenActionMap.java
```diff
@@ -248,11 +251,12 @@ private DebugAction()
 		@Override
 		public void actionPerformed(ActionEvent e)
 		{
+			GuiAssertions.assertIsNotJavaFXThread();
 			if (dialog == null)
 			{
-				dialog = new DebugDialog();
+				Platform.runLater(() -> dialog = new DebugDialog());
```
javydreamercsw: Why not all the if login on the same thread call? If null it ends up using two different threads.

grimreaper: good call. I'll fix this

PR #6117, opened 2019-11-30
Partially revert changes to enumerations

tfriedrichs:
The commit 5d56bdf7d3b32cbae141e826947c540ae1f3b4af changes the retrieval of enum entries by name to EnumUtils.getEnumIgnoreCase(). This does not match previous functionality where the given string is checked against the text variable. Therefore, pcgen crashes for enum entries where the name and text do not match.


grimreaper:
LGTM. Will land when tests pass. Thanks for the fix.

LegacyKing:
@grimreaper Tests passed.

PR #6128, opened 2019-12-05
General tidy up with autolint and refactoring

hyphz:
Clearing out cruft in auto refactor testing. Doesn't address any bug but improves code and resolves some strange issues that were lying around. Passes full test suite.


hyphz:
Ok, I'll try and break this up better. Is the "old wiki" actually the current one?

LegacyKing:
http://159.203.101.162/w/index.php/Main_Page <--- The current wiki, not much has changed.

PR #6137, opened 2019-12-16
Second round. Not sure why the first is still showing up here.

hyphz:


LegacyKing:
You reused a branch? Didn't merge in master with a rebase?

code/src/java/pcgen/cdom/facet/analysis/ChallengeRatingFacet.java
```diff
@@ -276,10 +276,7 @@ private Integer getClassRaceCRMod(CharID id, PCClass cl)
 			if (SettingsHandler.getGame().getClassTypeByName(classType) != null)
 			{
 				Integer crMod = raceFacet.get(id).get(MapKey.CRMOD, classType);
-				if (crMod != null)
-				{
-					return crMod;
-				}
+                return crMod;
```
LegacyKing: Not sure this is a safe change. The if statements usually got added when a null would cause an issue...

code/src/java/pcgen/cdom/helper/Aspect.java
```diff
@@ -107,7 +107,7 @@ public Aspect(final AspectName key, final String aString)
 	private void parseAspectString(final String aString)
 	{
 		int currentInd = 0;
-		int percentInd = -1;
+		int percentInd;
```
LegacyKing: Not sure this was useful. You've removed a qualifier value in place to prevent something... Was this tested?

hyphz: It ran through the automated tests fully, although I didn't know they weren't considered complete. percentInd is immediately reassigned by the condition on the while loop, so this -1 should never make any difference.

code/src/java/pcgen/cdom/facet/analysis/ChallengeRatingFacet.java
hyphz: In the original if crMod is NULL, the IF statement falls through, lines 285-301 have no effect because they are on the other side of an earlier IF, and then NULL is returned anyway by line 302. So this shouldn't change anything.

PR #6167, opened 2020-01-05
Added application icon, which is nice to have for GUI launchers

notklaatu:
The application icon (once located in the system/sponsors/pcgen dir) was deleted in commit b48176b68c887ec86647198861ece83d98d4e6c9. While it's not necessary to represent a sponsor, it is a nice icon to distribute with the application, so it can be used on desktops to represent the app itself. 

I did look elsewhere for the icon but couldn't find it anywhere. If I'm mistaken, then obviously this PR is moot.


grimreaper:
Thank you!

PR #6198, opened 2020-02-13
[build] use modern configuration names during build

grimreaper:


karianna:
@grimreaper - build fails, not sure what change in this file has caused that though

kypvalanx:
```
File: "/home/travis/build/PCGen/pcgen/build\nsis\PCGen_60903_opt\plugin\pdf\libs\*.*" -> no files found.

Usage: File [/nonfatal] [/a] ([/r] [/x filespec [...]] filespec [...] |

   /oname=outfile one_file_only)

Error in script "/home/travis/build/PCGen/pcgen/installers/win-installer/pcgen.nsi" on line 133 -- aborting creation process

```

looks like a file isn't being put where it's supposed to be?

BenCAllen:
I've ended up reproducing this error when I've tried updating the build config names separately. I'll work through and try to determine which of them is causing an issue 😃

BenCAllen:
After some experimentation, the line that makes it error out is:
`compile group: 'org.apache.xmlgraphics', name: 'fop', version:'2.4'`
Updating the `compile` command to `implementation` in line with the rest of the changes causes the build to fail.
Which is interesting, as that doesn't appear to be a change made in this PR.... 🤔

PR #6199, opened 2020-02-15
cleaned up Globals and changed a few callers to work with the changes

kypvalanx:
- removed the deprecated getGame method in favor of the supported version.
- EQ_SLOT_MAP was only every using integers, so now it uses integers
- canResizeHaveEffect was using a list, with a backup list, but the backup list was only ever populated with the firs list
- getSection15 was the only access to the underlying variable, and it's only modified and never queried.

### DIFF
--- a/code/src/java/pcgen/persistence/lst/CampaignLoader.java
+++ b/code/src/java/pcgen/persistence/lst/CampaignLoader.java
@@ -23,10 +23,8 @@ package pcgen.persistence.lst;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.List;
-
 import pcgen.cdom.base.Constants;
 import pcgen.cdom.enumeration.ListKey;
-import pcgen.cdom.enumeration.SourceFormat;
 import pcgen.core.Campaign;
 import pcgen.core.Globals;
 import pcgen.core.prereq.Prerequisite;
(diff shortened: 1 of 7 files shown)


code/src/java/pcgen/persistence/lst/CampaignLoader.java
```diff
@@ -157,19 +155,6 @@ public void loadCampaignLstFile(URI filePath) throws PersistenceLayerException
 		{
 			// Check the campaign's prerequisites, generating errors if any are not met but proceeding
 			validatePrereqs(campaign.getPrerequisiteList());
-			List<String> copyright = campaign.getListFor(ListKey.SECTION_15);
```
mertonmonk: Why is the copyright portion removed?

kypvalanx: it doesn't actually do anything.  the list is retrieved for adding and then never used.  according to my search at least.  if there is a way it is being accessed i need to change my techniques for searching.

mertonmonk: I don't know that it is used - I'll search as well.  If it isn't used - might as well remove dead code.

kypvalanx: i'm using intellij's search functions if it helps

mertonmonk: Okay - looks good.  Nice work!

PR #6243, opened 2020-04-26
Java Fx Cross Platform support

TimMcGilly:
I've started to implement support for cross compiling pcgen with javafx. This fixes https://pcgenorg.atlassian.net/browse/DATA-4149 and other issues due to javafx being platform specific. https://pcgenorg.atlassian.net/browse/CODE-3472
https://pcgenorg.atlassian.net/browse/CODE-3310

Steps to get javafx cross platform
1. Creating a jar file without any java fx platform specific modules bundled.
2. Download JDK for all platforms and add javafx jmods for each platform into the JDK
3. Use Badass Runtime Plugin (https://badass-runtime-plugin.beryx.org/) to generate a platform specific JRE with the javafx mods in it. This creates a folder for each platform. This was partly already implemented with modules listed so I just expanded it to include the javafx modules. This image is based on the installDir.
5. Download javafx .jars for person specific platform and run the tests with with 'add-modules' as the tests need javafx as well.
4. Copy pcgen.jar into directory of each image and copy the launch4j exe.
5. User runtimeZip to zip each platforms 

This should create a zip for each platform. 

Todo
- Testing on multiple platforms
- Check that no licences and other files are missing
- Get it working with developer workflow in intellij for example. I haven't got it properly setup myself so would need to know how they have it configured.
- Get mac builds working. I've not tried these at all yet.
- NSIS needs testing. libs folder now named lib.

Known issues
- ~pcgen.jar not properly copying to images~
- ~pcgen.exe not properly copying to images.~
- ~jlink breaking with different version of jdk. I believe this is the current issue with travis.~ Both the build version of java and the output version of java needs to be the same. I've chosen java 14 to match Travis but this could be easily changed. 
- ~lib folder in image contains javafx jars but doesn't use them as far as I can tell.~
- ~gradle run task broken. appears to be generating invalid jvm arguments~

I would love any help with testing and/or improving this.


grimreaper:
thank you so much for looking at this. Looking foward to coming out of draft!

karianna:
@TimMcGilly - If you rebase you'll pick up the Java 14 JavaFX updates :-)

TimMcGilly:
Just rebased :) Thanks for the suggestion.

karianna:
@TimMcGilly - Is there more to do here?

TimMcGilly:
I am currently trying to get this working with travis builds and the mac installer. I hope to get that done this weekend and then merge it into this pull request. Also before merging someone needs to test that this doesn't break their intellij or other IDE workflow as I have a suspicion it may depending on configuration. 

The generation of multi-platform zips should be working at this current stage so if we wanted to split get installer working in a separate task that could also be a option?

karianna:
> @karianna I am currently trying to get this working with travis builds and the mac installer. I hope to get that done this weekend and then merge it into this pull request. Also before merging someone needs to test that this doesn't break their intellij or other IDE workflow as I have a suspicion it may depending on configuration.
> 
> The generation of multi-platform zips should be working at this current stage so if we wanted to split get installer working in a separate task that could also be a option?

Happy to wait :-), I can test in my IntelliJ once you're done.  I also have eclipse and netbeans and VS code FWIW

karianna:
@TimMcGilly - Any luck with this?

TimMcGilly:
Sorry for the wait. I've got Mac installers working now with jpackage. Due to the lack of signing these installers will appear with a warning saying that they don't match secruity settings. This can be bypassed by right clicking then open the installer. Not sure where this should be documented so it appears with download?

I attempted to get linux installers as well but there appears to be issues with the creation of deb and rpm files. https://bugs.java.com/bugdatabase/view_bug.do?bug_id=JDK-8233449

I've split up the checksums due to the installers and zips possibly being generated on separate machines (different travis machine).

I've also added a workaround for pcgen.bat startscript where when modifying the output path of it that JAVA_HOME directory has a incorrect depth. There is a open issue for this in the gradle repo https://github.com/gradle/gradle/issues/10387

I've rebased against master and I think this is at ready for review now and needs some developer testing.

LegacyKing:
@karianna I think this is ready for review.

LegacyKing:
@mertonmonk can you take a look at this as well. :)

TimMcGilly:
A migration to JavaFX has been taking place for a while although I'm not certain of the reasons. (Someone else may be able to chime in.) This pull request is to get the JavaFX working with cross platform builds. It has the added advantage of bundling the jdk for each OS in their zip which removes the common issues people have installing java on their own machines.

The reason this is java 14 is because for the mac installer (and hopefully a linux one) I am building using jpackage which is new in java 14. This is due macappbundle (previously used) https://github.com/crotwell/gradle-macappbundlebeing broken for a long time, recently deprecated and I couldn't easily find a way to get it working with the bundled jre.

karianna:
> @mertonmonk A migration to JavaFX has been taking place for a while although I'm not certain of the reasons. (Someone else may be able to chime in.) This pull request is to get the JavaFX working with cross platform builds. It has the added advantage of bundling the jdk for each OS in their zip which removes the common issues people have installing java on their own machines.
> 
> The reason this is java 14 is because for the mac installer (and hopefully a linux one) I am building using jpackage which is new in java 14. This is due macappbundle (previously used) https://github.com/crotwell/gradle-macappbundlebeing broken for a long time, recently out of active development and I couldn't easily find a way to get it working with the bundled jre.

So to be clear, does this mean we're bundling Java 14 or expecting users to install Java 14?

TimMcGilly:
@karianna Devs would need to install Java 14 to build and compile. Users would have Java 14 bundled with pcgen.

TimMcGilly:
Just did another test with a different intellij configuration and discovered that the application configuration may fail to run due to missing javafx libraries. 

Sorry about missing this earlier. I've reverted the pull request and have pushed a fix for this which should fix any issues with intellij. If anyone is actively using netbeans or eclipse or another ide it would be worth testing whether there are any similar issues before merging again.

PR #6282, opened 2020-06-18
Implemented Feature where Inputs on a preview sheet tagged with a tar…

kypvalanx:
…get_var attribute are persistent between sessions

* only the pathfinder preview sheet currently has functional persistent inputs
* adding `"target_var"="{varname}"` to an input tag makes the tag persistent
* This should not reduce performance as it's tying listeners to input tags on load

### DIFF
--- a/code/src/java/pcgen/core/PlayerCharacter.java
+++ b/code/src/java/pcgen/core/PlayerCharacter.java
@@ -511,6 +511,7 @@ public class PlayerCharacter implements Cloneable, VariableContainer
 			CNAbilityFactory.getCNAbility(AbilityCategory.LANGBONUS, Nature.VIRTUAL, Globals.getContext()
 				.getReferenceContext().getManufacturerId(AbilityCategory.LANGBONUS).getActiveObject("*LANGBONUS"));
 	private final CodeControl controller;
+	private Map<String, String> previewSheetVars = new HashMap<>();
 
 	/**
 	 * Constructor.
@@ -10087,5 +10088,22 @@ public class PlayerCharacter implements Cloneable, VariableContainer
 	public Gender getGenderObject()
 	{
 		return genderFacet.getGender(id);
+	}
+
+
+	public void addPreviewSheetVar(String key, String value)
+	{
+		setDirty(true);
+		previewSheetVars.put(key, value);
+	}
+
+	public String getPreviewSheetVar(String key)
+	{
+		return previewSheetVars.get(key);
+	}
+
+	public Map<String, String> getPreviewSheetVars()
+	{
+		return previewSheetVars;
 	}
 }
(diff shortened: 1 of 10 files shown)


code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -511,6 +511,7 @@
 			CNAbilityFactory.getCNAbility(AbilityCategory.LANGBONUS, Nature.VIRTUAL, Globals.getContext()
 				.getReferenceContext().getManufacturerId(AbilityCategory.LANGBONUS).getActiveObject("*LANGBONUS"));
 	private final CodeControl controller;
+	private Map<String, String> previewSheetVars = new HashMap<>();
```
karianna, line 514: I'd recommend wrapping this in a domain class.

code/src/java/pcgen/gui2/csheet/PreviewVariablesHandler.java
```diff
@@ -0,0 +1,80 @@
...
+            HTMLInputElementImpl element = (HTMLInputElementImpl) elements.item(i);
+            String key = getInputKey(element);
+            if (key == null || key.isEmpty())
+            {
+                continue;
+            }
+            setInputValue(element, character.getPreviewSheetVar(key));
+            element.addEventListener("change", evt ->
+            {
+                HTMLInputElement input1 = (HTMLInputElement) evt.getCurrentTarget();
```
karianna, line 42: could just be called input?

code/src/java/pcgen/io/PCGVer2Creator.java
```diff
@@ -470,6 +470,12 @@ public String createPCGString()
 		appendComment("Suppressed Biography Fields", buffer); //$NON-NLS-1$
 		appendSuppressBioFieldLines(buffer);
 
+		/*
+		 *
```
karianna, line 474: empty comment

code/src/java/pcgen/core/PlayerCharacter.java
kypvalanx: there's no database for this to persist to.  why would I use a domain object in this case?

code/src/java/pcgen/gui2/csheet/PreviewVariablesHandler.java
kypvalanx: sure can.  i renamed the conflicting variable already

code/src/java/pcgen/io/PCGVer2Creator.java
kypvalanx: added comment

code/src/java/pcgen/core/PlayerCharacter.java
karianna: Modeling is a good thing irrespective of data storage :-)

karianna: But it's lacking in other areas of the code has so this is pretty minor - I'll merge regardless

kypvalanx: so, i'm very much about doing it the right way.  if there's a good reason to do something a certain way I'm all about it, and I don't want it merged regardless.  a domain class for a single map adds complexity because there's a whole new class to consider so in this case I agree.  i think if the map had additional variables associated specifically with the map it should be wrapped in a class with those values.

PR #6294, opened 2020-06-30
Exception cleanup

kypvalanx:
ran an inspection to remove throws declarations where nothing could be thrown.  then I removed try/catch clauses that couldn't catch anything.  I did this in a few iterations.

### DIFF
--- a/code/src/java/pcgen/cdom/content/IndirectCalculation.java
+++ b/code/src/java/pcgen/cdom/content/IndirectCalculation.java
@@ -94,7 +94,7 @@ public final class IndirectCalculation<T> extends AbstractNEPCalculation<T>
 
 	@Override
 	public void isValid(FormulaSemantics semantics)
-    {
+	{
 		/*
 		 * Since this is direct (already has a reference to the object), it has no
 		 * semantic issues (barring someone violating Generics)
--- a/code/src/java/pcgen/cdom/content/IndirectCalculation.java
+++ b/code/src/java/pcgen/cdom/content/IndirectCalculation.java
@@ -22,7 +22,6 @@ import pcgen.base.calculation.BasicCalculation;
 import pcgen.base.formula.base.DependencyManager;
 import pcgen.base.formula.base.EvaluationManager;
 import pcgen.base.formula.base.FormulaSemantics;
-import pcgen.base.formula.exception.SemanticsException;
 import pcgen.base.util.Indirect;
 
 /**
@@ -94,8 +93,8 @@ public final class IndirectCalculation<T> extends AbstractNEPCalculation<T>
 	}
 
 	@Override
-	public void isValid(FormulaSemantics semantics) throws SemanticsException
-	{
+	public void isValid(FormulaSemantics semantics)
+    {
 		/*
 		 * Since this is direct (already has a reference to the object), it has no
 		 * semantic issues (barring someone violating Generics)
--- a/code/src/java/pcgen/core/analysis/SubstitutionLevelSupport.java
+++ b/code/src/java/pcgen/core/analysis/SubstitutionLevelSupport.java
@@ -26,11 +26,9 @@ import pcgen.core.Globals;
 import pcgen.core.PCClass;
 import pcgen.core.PlayerCharacter;
 import pcgen.core.SubstitutionClass;
-import pcgen.persistence.PersistenceLayerException;
 import pcgen.persistence.lst.PCClassLoader;
 import pcgen.persistence.lst.SourceEntry;
 import pcgen.persistence.lst.utils.DeferredLine;
-import pcgen.util.Logging;
 
 public final class SubstitutionLevelSupport
 {
@@ -45,14 +43,7 @@ public final class SubstitutionLevelSupport
 		final PCClassLoader classLoader = new PCClassLoader();
 		PCClass dummyClass = new PCClass();
 
-		try
-		{
-			classLoader.parseLine(Globals.getContext(), dummyClass, aLine, tempSource);
-		}
-		catch (PersistenceLayerException e)
-		{
-			Logging.errorPrint("Unable to parse line from levelArray: " + aLine);
-		}
+		classLoader.parseLine(Globals.getContext(), dummyClass, aLine, tempSource);
 		return dummyClass.getOriginalClassLevel(level).qualifies(pc, source);
 	}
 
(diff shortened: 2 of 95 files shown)


code/src/java/pcgen/cdom/content/IndirectCalculation.java
```diff
@@ -94,8 +93,8 @@ public void getDependencies(DependencyManager fdm)
 	}
 
 	@Override
-	public void isValid(FormulaSemantics semantics) throws SemanticsException
-	{
+	public void isValid(FormulaSemantics semantics)
+    {
```
karianna, line 97: looks like the whitespace is out here and in other files?

kypvalanx: doesn't the checkstyle find that stuff?  weird.  i'll take a look.

kypvalanx: oh, it's going from tabs to spaces.  what's our preference on this and can we set it up in checkstyle?

also:https://chrome.google.com/webstore/detail/render-whitespace-on-gith/ifdbipohclgnokjgpejhnbjdlgjkkhom/related

kypvalanx: @karianna i need a quick clarification on this

karianna: We have a checkstyle config set up for the project --> https://github.com/PCGen/pcgen/tree/master/code/standards - see what that has in it to start with and if we're missing a rule then we can add one.

kypvalanx: @karianna there is a rule for spaces.  I found a workaround for enforcing tabs.  what is our preference on tabs or spaces?  my personal preference is spaces because they are a fixed width.

kypvalanx: anyway. the right thing to do here is to reduce whitespace change and open a  PR with a reformat + checkstyle rule.  i guess i'll open an issue to discuss whitespace after work.  i'll also update this after work.

kypvalanx: @karianna whitespace fixed and issue created https://github.com/PCGen/pcgen/issues/6298

code/src/java/pcgen/core/analysis/SubstitutionLevelSupport.java
```diff
@@ -45,14 +43,7 @@ private static boolean levelArrayQualifies(int level, final PlayerCharacter pc,
...
 
-		try
-		{
-			classLoader.parseLine(Globals.getContext(), dummyClass, aLine, tempSource);
-		}
-		catch (PersistenceLayerException e)
-		{
-			Logging.errorPrint("Unable to parse line from levelArray: " + aLine);
-		}
+		classLoader.parseLine(Globals.getContext(), dummyClass, aLine, tempSource);
```
thpr, line 46: I'm concerned this changes the function of this method - previously, this exception would cause an error that is local to this method; now the exception will bubble up to a higher level; potentially causing it to process many less lines in a file.  The design of this (and many other items where you are removing reactions to PersistenceLayerException) are to try to have a minimal failure - as little as possible should be not processed.  In this new case, if a failure is encountered, does it completely stop processing the file (or the entire load?)

code/src/java/pcgen/gui2/converter/DefaultTokenProcessor.java
```diff
@@ -31,42 +30,35 @@
 	@Override
 	public String process(TokenProcessEvent tpe)
 	{
-		try
```
thpr, line 34: Similar to above - previous design is fail of a single token is just that - now it will bubble up to... where?

code/src/java/pcgen/persistence/lst/BioSetLoader.java
```diff
@@ -104,16 +104,8 @@ public void parseLine(LoadContext context, String lstLine, URI sourceURI)
...
-				{
-					LstUtils.processToken(context, ageSet, sourceURI,
-						colToken.nextToken());
-				}
-				catch (PersistenceLayerException e)
-				{
-					Logging.errorPrint(
-						"Error in token parse: " + e.getLocalizedMessage());
-				}
+				LstUtils.processToken(context, ageSet, sourceURI,
```
thpr, line 107: This seems clear that instead of failing by token, one will lose the entire line if an exception is thrown - since the exception will exit the while loop

code/src/java/pcgen/core/analysis/SubstitutionLevelSupport.java
karianna: We used to call this 'heroic mode' ;-), the loaders keep on chugging even on glitchy data.

kypvalanx: it doesn't change any functionality.  the exception is checked and you can investigate to see that it cannot be thrown in those places. this has been dead code for a while.  can you show that the exception is actually thrown?

code/src/java/pcgen/persistence/lst/BioSetLoader.java
kypvalanx: again, the point of this PR is to remove throws declarations that never can throw exceptions and remove catch blocks that can no longer catch anything.  it can't fail by a token if an exception cannot be thrown.  i think you find if you trace the checked exception down it is declared at a low level on methods that do not throw it and that throws declaration propagates up the stack for a bit.  these exceptions cannot possibly be thrown as the code is written.

code/src/java/pcgen/gui2/converter/DefaultTokenProcessor.java
kypvalanx: it won't bubble up, it cannot possibly be thrown

thpr:
I stand corrected; apparently various cleanups of parse* and the use of ParseResult* has resulted in only unparse or PRExxx tokens using PersistenceLayeException.  So these are unused - WooHoo!

kypvalanx:
> I stand corrected; apparently various cleanups of parse* and the use of ParseResult* has resulted in only unparse or PRExxx tokens using PersistenceLayeException. So these are unused - WooHoo!

this refactoring was exciting, more and more throws and catches were being shown as unused as i went

PR #6329, opened 2020-07-21
Eliminate Suppress BioField

thpr:
This feature is not available through the GUI, so remove for now.


karianna:
@thpr and same here

PR #6343, opened 2020-08-09
Fix Various tests in PCGen Unit Tests

thpr:


code/testsuite/csheets/35e_Alice.xml
```diff
@@ -125,7 +125,7 @@ BIO
 		</reach>
 		<region>None</region>
 		<reputation>0</reputation>
-		<residence>Peachmond</residence>
+		<residence></residence>
```
LegacyKing, line 128: Trivial change, but why?

thpr: Technically, it's listed as Alice's City, not Residence, so it's correct in being empty.  The XML file doesn't list city, so it's never shown there (we should probably update that XML template)

thpr:
Well, since new variable system all variables have a default and Deity is now a variable, it has a default.  So yes, it's intended that Deity is set to None.

Happy to discuss if you think that's an issue - we can actually change the format of things to OPTIONAL[DEITY] and thus Deity doesn't require a default... since the default is Optional.empty()... but that requires more invasive changes in code and we should do that as a separate patch later.

PR #6359, opened 2020-08-14
Fixed intellij configuration issues.

TimMcGilly:
Fix issues with application configuration in intellij caused by missing JavaFX libs. Issue was introduced in https://github.com/PCGen/pcgen/pull/6243.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,8 @@ javafx {
     modules = [ 'javafx.controls', 'javafx.web', 'javafx.swing', 'javafx.fxml', 'javafx.graphics' ]
 
     //Check if running in ide. If so bundles javafx in to jar. If not only compiles so that runtime can add different platform javafx later.
-    if (!System.getProperty('idea.active')) {
+    //Also allows for gradle property which if true will force the single platforms javafx to be bundled to create local build. Can be added in command line with "-PlocalOnly=true."
+    if (!System.getProperty('idea.active') && !(localOnly.toBoolean())) {
         configuration = "compileOnly"
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,11 @@ plugins {
 javafx {
     version = "14.0.1"
     modules = [ 'javafx.controls', 'javafx.web', 'javafx.swing', 'javafx.fxml', 'javafx.graphics' ]
-    configuration = "compileOnly"
+
+    //Check if running in ide. If so bundles javafx in to jar. If not only compiles so that runtime can add different platform javafx later.
+    if (!System.getProperty('idea.active')) {
+        configuration = "compileOnly"
+    }
 }
 
 group = 'net.sourceforge.pcgen'
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,2 @@
 version=6.09.05-SNAPSHOT
+localOnly = false


TimMcGilly:
The one issue with this fix is that finial pcgenRelase or runtime gradle tasks will fail if run from intellij but this is less important than having a working application and I suspect releases are already normally done via command line.

LegacyKing:
@TimMcGilly I'm dead in the water currently. Will this address this issue?

2020-08-14T22:54:09.049522400 SEVERE main LoggingUncaughtExceptionHandler:32 Uncaught error on thread  Thread[main,5,main] - ignoring
java.lang.NoClassDefFoundError: javafx/embed/swing/JFXPanel
        at pcgen.system.Main.startupWithGUI(Main.java:212)
        at pcgen.system.Main.main(Main.java:138)
Caused by: java.lang.ClassNotFoundException: javafx.embed.swing.JFXPanel
        at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:583)
        at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:521)
        ... 2 more

TimMcGilly:
Yep, it should fix that issue if caused when running from intellij @LegacyKing

LegacyKing:
That's the command line actually...

LegacyKing:
Also, not sure when this started popping up, but any chance we can address this? 
image

TimMcGilly:
@LegacyKing That second issue I've tracked down to this commit https://github.com/PCGen/pcgen/commit/c2390813b631c994441a471c12170588afb300dc
which is just a normal plugin update. We could add a module-info.java but I not sure one would be appropriate because as afaik pcgen is quite non-modular.

What command caused that error on start up? Also what java version are your running?

TimMcGilly:
@LegacyKing If this occured when running the jar from qbuild or some other command, this failure is due to javaFX not being bundled within the jar.

Using `gradlew run` should already work as when running I have it add the necessarily modules with jvm arguments.
```
"--module-path /mods/lib/ --add-modules javafx.controls,javafx.web,javafx.swing,javafx.fxml"
```

So that local jars can still be generated without extra need for jvm args, I've just added a new gradle property which lets you force gradle to include the os specific javafx code into your jar. You need to add `-PlocalOnly=true` to the end of a gradle command to do this. 

For example `.\gradlew assemble qbuild -PlocalOnly=true` which should give a working local jar. 

This argument should not be used on cross-platform or zip builds as it would break them.

Is there anywhere I should document this?

I looked into if I could get gradle to detect if it is a cross-platform build and then change this property itself but it turns into a chicken and the egg type problems and appears impossible to my knowledge.

LegacyKing:
I'll give it a test because only the .\gradlew run is working.

LegacyKing:
That fixed the issue

DarthBorehd:
Can somebody help me understand how to fix this?  I ran into this issue on a brand new install of Linux Mint 20

sepatel:
> Can somebody help me understand how to fix this? I ran into this issue on a brand new install of Linux Mint 20

Do the following
```
./gradlew assemble qbuild -PlocalOnly=true
```

Then in the `build/distributions` folder is a `.tar` file that you can expand anywhere and then run the `./pcgen`.

PR #6374, opened 2020-09-08
Sanitize filenames when creating them from character name

alanshutko:
Resolves CODE-3501, which was caused by using characters from a character name which are not allowed in filenames.

### DIFF
--- a/code/src/java/pcgen/gui2/PCGenFrame.java
+++ b/code/src/java/pcgen/gui2/PCGenFrame.java
@@ -111,6 +111,7 @@ import pcgen.system.Main;
 import pcgen.system.PCGenPropBundle;
 import pcgen.system.PCGenSettings;
 import pcgen.system.PropertyContext;
+import pcgen.util.FileHelper;
 import pcgen.util.Logging;
 import pcgen.util.chooser.ChoiceHandler;
 import pcgen.util.chooser.ChooserFactory;
@@ -939,7 +940,8 @@ public final class PCGenFrame extends JFrame implements UIDelegate
 		File prevFile = file;
 		if (file == null || StringUtils.isEmpty(file.getName()))
 		{
-			file = new File(parentPath, character.getNameRef().get() + Constants.EXTENSION_CHARACTER_FILE);
+			String characterName = FileHelper.sanitizeFilename(character.getNameRef().get());
+			file = new File(parentPath, characterName + Constants.EXTENSION_CHARACTER_FILE);
 		}
 		chooser.setSelectedFile(file);
 
--- a/code/src/java/pcgen/util/FileHelper.java
+++ b/code/src/java/pcgen/util/FileHelper.java
@@ -117,4 +117,22 @@ public final class FileHelper
 
 		return relative.getAbsolutePath().substring(root.length(), relative.getAbsolutePath().length());
 	}
+
+	/**
+	 * Remove characters which are not allowed by certain OSes and filesystems. The string returned can be used
+	 * as a filename on all OSes which can run pcgen.
+	 *
+	 * The following characters are removed:
+	 * <ul>
+	 *     <li>&lt;, &gt; :, ", \\, |, ?, * because they are
+	 *     <a href="https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions">restricted on Windows</a></li>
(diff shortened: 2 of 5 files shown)


LegacyKing (commented):
Looks good to me.

code/src/java/pcgen/gui2/PCGenFrame.java
```diff
@@ -939,7 +940,8 @@ boolean showSaveCharacterChooser(CharacterFacade character)
 		File prevFile = file;
 		if (file == null || StringUtils.isEmpty(file.getName()))
 		{
-			file = new File(parentPath, character.getNameRef().get() + Constants.EXTENSION_CHARACTER_FILE);
+			String characterName = FileHelper.sanitizeFilename(character.getNameRef().get());
+			file = new File(parentPath, characterName + Constants.EXTENSION_CHARACTER_FILE);
```
javydreamercsw, line 944: Any reason not to do this on the same line?

code/src/java/pcgen/util/FileHelper.java
```diff
@@ -117,4 +117,22 @@ private static String stripOffRoot(File relative)
...
+	 *     <li>&lt;, &gt; :, ", \\, |, ?, * because they are
+	 *     <a href="https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions">restricted on Windows</a></li>
+	 *     <li>/ and U+0000 (Unicode NULL) , restricted on Linux, Mac, and Windows</li>
+	 * </ul>
+	 *
+	 * @param invalidCharacterFilename Filename to sanitize
+	 * @return string with invalid characters removed
+	 */
+	public static String sanitizeFilename(String invalidCharacterFilename) {
+		return invalidCharacterFilename.replaceAll("[\u0000\\\\/:*?\"<>|]", "");
```
javydreamercsw, line 136: Null check for the parameter? Recently I started using Lombok NonNull annotation for parameters.

code/src/java/plugin/pcgtracker/PCGTrackerPlugin.java
```diff
@@ -328,8 +329,9 @@ private boolean savePC(PlayerCharacter aPC, boolean saveas)
 
 		if (aPCFileName.isEmpty())
 		{
+			String characterName = FileHelper.sanitizeFilename(aPC.getDisplay().getDisplayName());
 			prevFile = new File(PCGenSettings.getPcgDir(),
-				aPC.getDisplay().getDisplayName() + Constants.EXTENSION_CHARACTER_FILE);
+				characterName + Constants.EXTENSION_CHARACTER_FILE);
```
javydreamercsw, line 334: Any reason not to do this on the same line?

javydreamercsw (approved):

code/src/java/pcgen/gui2/PCGenFrame.java
alanshutko: I could do it on the same line. It seemed like the line was getting pretty long, though, and breaking it up looked a bit more clear.

code/src/java/pcgen/util/FileHelper.java
alanshutko: Good point. Will adjust.

alanshutko: I'm checking other parts of the code to see if there is a pattern to parameter checking. Does anyone know the preferred pattern?

We could add lombok, but I'm not sure I want to add a dependency in my first PR. 

Another option would be to do an explicit check and throw a null pointer exception with clear messaging (similar to what lombok would do with @NotNull.

Option 3 would be to check for null, send something to the system logs, but return an empty string as a safe value for downstream code.

LegacyKing:
Error is not code related. Working on cleaning up these errors.

LegacyKing:
If this is approved, we can merge it in. The CI is failing for data not code.

LegacyKing:
I do have a question. Why is this pointed at the 6.08 branch instead of the MASTER branch @alanshutko ?

alanshutko:
When I last checked, 6.09/MASTER does not prefill a filename at all.

LegacyKing:
Ah, that makes sense. I'm also disappointed we lost that feature in 6.09.

PR #6507, opened 2021-05-25
[bugfix] fix casting error with preview handler

grimreaper:
Problem

modern libraries prevent casting to HTMLInputElementImpl so we're unable
to fully produce character previews

Solution

Use the EventTarget interface instead.


karianna (approved):
LGTM

PR #6544, opened 2021-07-25
Update modernbasics_abilities_talents.lst

FerretDave:
corrected the output for the DR to be the variable, rather than static


LegacyKing:
Does this actually work?

FerretDave:
Yes ;-)   previous was a string "1/-", adding a second tough talent for DRII made no difference, now it does.

FerretDave:
going to redo this, the *DESC* for the subsequent talents needs populating instead really

PR #6561, opened 2021-08-29
Encapsulate Races

Llisandur:
Create individual race folders so the race files only exist in a single folder.
Each source pcc loads the race loader pcc.
ARG, ISR, and others no longer require Bestiary sources.


LegacyKing (approved):
Just some minor nits... Approved

PR #6585, opened 2021-09-18
[build] bump JavaFX to latest (17.0.0.1)

grimreaper:


karianna:
This would require moving all of PCGen to 17.  I think we'd need to hold off on this one?

grimreaper:
I don't think so. This is OpenJFX: not the JVM?

grimreaper:
rebased

karianna:
Update to 17.0.6 in recent PR

PR #6658, opened 2022-01-13
fix Gradle issues with JCenter outage Jan 12 2021.

tildar:


karianna:
How does this fix it?  Does the main gradle build file not supply this info already?

tildar:
I was having Gradle builds fail on my fork due to JCenter outage.
I found this as an option here: https://discuss.gradle.org/t/plugins-redirecting-to-jcenter/41909/6
Plugged it in and the build worked, and thought it would be helpful to pass it upstream.

karianna:
I think the repositories are defined in _build.gradle_ instead:

Maybe add to that block instead?

```
repositories {
    mavenCentral()
    ivy {
        url "http://pcgen.sourceforge.net/mvnrepo"
        allowInsecureProtocol = true
        patternLayout {
            artifact "[organisation]/jars/[artifact]-[revision].[ext]"
        }
    }
    ivy {
        name "fileRepo"
        url 'http://pc-gen.org/librepo/'
        allowInsecureProtocol = true
    }
    maven {
        url "https://repository.jboss.org/nexus/content/repositories/thirdparty-releases/"
    }
}
```

karianna:
This issue resolved itself

PR #6666, opened 2022-01-25
Fixing data from Pathfinder:HOTW

jasontamez:
If needed, compare with https://www.d20pfsrd.com/classes/base-classes/witch/witch-patrons


LegacyKing:
Uh, this has alterations to three different things. Not a clean PR.

- Did APG have an errata to include new CMD/CMB's?

jasontamez:
I didn't realize pushing new stuff would change the original pull request, and I was going to push the other two separately. Sorry about that. Will resubmit as separate requests.

PR #6720, opened 2022-07-25
Fix DATA-4287 ("Incorrect damage dice for Kinetic blast")

Tamdrik:
Updated Kineticist's "Kinetic Blast" damage to scale properly, at odd levels instead of even.  Ref: Jira issue DATA-4287.


karianna (approved):
LGTM

PR #6806, opened 2023-02-02
Updated Samurai Resolve ability

BlyatBeauty:
Resolve has been given CheckCount and CheckType Aspects and now appears as checkable boxes on an exported character sheet.


BlyatBeauty:
I'm struggling to squash my commits, so I'll just make a new PR

PR #6833, opened 2023-03-30
DATA-4314 - Fix #1

Vest:
Fixes the exception caused by **TEMPBONUS:VAR|FightingDefensivelyAttackPenalty**. Should be **BONUS:VAR|FightingDefensivelyAttackPenalty**.

More details are in the corresponding JIRA DATA-4314.


Vest:
The error is:
```
LSTERROR LstUtils:145 Illegal TEMPBONUS subtoken 'VAR' 'FightingDefensivelyAttackPenalty|-2' for pcgen.core.Ability Crane Style (Source: file:/Users/.../git/pcgen/data/pathfinder/paizo/roleplaying_game/ultimate_combat/uc_feats.lst )
LSTERROR LstUtils:145 Failed in parsing typeStr: TEMPBONUS VAR|FightingDefensivelyAttackPenalty|-2 for pcgen.core.Ability Crane Style
LSTERROR LstUtils:145 Illegal Token 'TEMPBONUS' 'VAR|FightingDefensivelyAttackPenalty|-2' for pcgen.core.Ability Crane Style in file:/Users/.../git/pcgen/data/pathfinder/paizo/roleplaying_game/ultimate_combat/uc_feats.lst
```
It seems that TEMPBONUS doesn’t like “VAR”: `TEMPBONUS:VAR|FightingDefensivelyAttackPenalty|-2`.

I have analyzed other temporary bonuses, and decided to use the BONUS keyword instead of TEMPBONUS.

Vest:
@BlyatBeauty, may I ask you to take a look at my PR? It fixes an error message. I guess, you are the author of

BlyatBeauty:
Hi yes, sorry for the delay! Schoolwork had me burnt out, I'm checking this right away!

BlyatBeauty:
So I swear there was logic behind why I originally wrote the bonus that way. That said, it doesn't work, whereas your solution definitely fixes the issues with Crane Style. 

I don't have the ability to accept pull requests and allow them to merge but to whoever does, I can confirm that this PR works.

BlyatBeauty (approved):

Vest:
No need to swear. Thank you for your review. 

Your code can be fixed, if you add PC or ANYPC in front of VAR. But I don’t remember, whether VAR is allowed in TEMPBONUS. Plus I check other DATA scripts before proposed this solution. 
I don’t play this setting, so I cannot validate the fix.

BlyatBeauty:
No, adding PC before VAR wouldn't make it count for some reason, I did try it though. I haven't tried with ANYPC but again, your solution works fine, it gives off the correct calculations every time.

Vest:
@karianna may I disturb you, and ask you to accept my PR?

LegacyKing:
TEMPBONUS is written incorrectly. You want to write it:
`TEMPBONUS:PC|VAR|FightingDefensivelyAttackPenalty|-2`.

LegacyKing:
Just to be clear. BONUS:VAR is only to be used for static/constant bonuses. TEMPBONUS is for circumstantial/infrequent penalties or bonuses.

data/pathfinder/paizo/roleplaying_game/ultimate_combat/uc_feats.lst
```diff
@@ -289,7 +289,7 @@ Wave Strike							CATEGORY:FEAT		TYPE:Combat
...
+Crane Style							CATEGORY:FEAT		TYPE:Combat.Style					PREABILITY:2,CATEGORY=FEAT,Dodge,Improved Unarmed Strike																																																					PREMULT:1,[PRETOTALAB:2],[PREVARGTEQ:MonkFeatQualify,1]																																																																																																																																																												DESC:Your unarmed fighting techniques blend poise with graceful defense.																																																																																																																																																																								SOURCEPAGE:p.93															BENEFIT:You take only a -2 penalty on attack rolls for fighting defensively. While using this style and fighting defensively or using the total defense action, you gain an additional +1 dodge bonus to your Armor Class.																																																																																																																																																																																																																																									TEMPBONUS:PC|COMBAT|AC|1|TYPE=Dodge		BONUS:VAR|FightingDefensivelyAttackPenalty|-2
```
LegacyKing, line 292: Fix the syntax please TEMPBONUS:PC|VAR|xxx

LegacyKing (changes requested):

Vest:
@BlyatBeauty I have updated my PR again. Unfortunately, I have several comments regarding this feat. Since I am new to this syntax, I see some drawbacks:
- in PCGen, the Crane Style (as a temp. bonus) can be picked up, when neither Fighting Defensively, nor Total Defense is selected. This feat still gives +1 Dodge bonus.
- It seems, that the attack (-2) is applied to both melee and ranged, however this should be wrong. The feet Crane Style is applied to a monk or somebody, who has "unarmed strike". So it means, you have to have empty hands (no weapons), and no ranged attack at all.

What do you think about this? Is it what can be corrected in the DATA?

data/pathfinder/paizo/roleplaying_game/ultimate_combat/uc_feats.lst
Vest: Done

BlyatBeauty:
While though Crane Style has Improved Unarmed Strike as a prerequisite, Crane Style's text never states that its reduced penalty only applies to unarmed strikes, and a monk may still choose to use other weapons besides their unarmed strike. 

I will only be able to test the PR tomorrow at the earliest.

BlyatBeauty:
> TEMPBONUS is written incorrectly. You want to write it:
> 
> `TEMPBONUS:PC|VAR|FightingDefensivelyAttackPenalty|-2`.

Ahh, thank you!!! I was wondering how to make this work right!

BlyatBeauty:
Just tested the newest PR. With `TEMPBONUS:PC|VAR|FightingDefensivelyAttackPenalty|-2` the feature seems to work exactly as intended, the Fighting Defensively Penalty is reduced by 2 when you apply Fighting Defensively, and the PC geins a +1 AC Bonus when the temporary bonus is applied.

While we could optimize this further - making the AC Bonus only apply when the player selects either the Fighting Defensively or the Total Defense conditions, I worry this would hinder certain users' way they use the program, and risks overcomplicating the entry.

Vest:
This is what I don’t like currently. You can select both - fighting defensively and total defense. That’s why the Crane Style might work incorrectly. 

I think, it will be another discussion in future, probably it isn’t related to the exception that I was trying to fix. 

Thank you for the review.

@LegacyKing , may I disturb you once again and check my PR?

BlyatBeauty (approved):

LegacyKing (approved):

PR #6837, opened 2023-04-02
Fixed the IDEA project generation

Vest:
Fixed the IDEA project generation
1. Test sources are properly included using non-deprecated method
2. JavaFX libraries (mods/lib) are properly added to the project

The generated project is able to run the application without any issues, but you have to add it as a configuration:
Run/Debug Configuration example

Previously, @karianna marked the line in build.gradle with "TODO", so I have decided to remove it, hoping that now the project generation can work well.

To test the changes, you have to clean IDEA files and generate them from the console:
`% gradle cleanIdea && gradle idea`

I don't recommend you to combine them into `gradle cleanIdea idea`, because I saw that changes are **not replaced, but merged**.

I guess, this is everything. So I hope, you can test my PR (_I tested it on Mac, if it is important_) and tell me, what might go wrong.

p.s. tests fail in IntelliJ, 663 failed tests out of 20830.


build.gradle
```diff
@@ -907,12 +908,29 @@ dependencyUpdates.resolutionStrategy {
...
+        projectLibrary.classes += files
+
+        projectLibraries += [ projectLibrary ]
+    }
+
     module {
-        testSourceDirs += project.sourceSets.itest.java.srcDirs
-        testSourceDirs += project.sourceSets.slowtest.java.srcDirs
-        testSourceDirs += project.sourceSets.test.java.srcDirs
+        testSources.from(project.sourceSets.itest.java.srcDirs,
```
karianna, line 926: Is this missing `project.sourceSets.test.java.srcDirs`?

karianna (changes requested):

karianna:
Also worth updating our README / docs so new developers know which gradle commands to fire

build.gradle
Vest: Nice catch, it is because the test folder is filled with this block. It seems, the IDEA plugin is smart enough to use it:
```groovy
    // Running regular tests
    test {
        java {
            srcDirs = ['code/src/utest', 'code/src/testcommon']
        }
        compileClasspath += sourceSets.main.output
        runtimeClasspath += sourceSets.main.output
        resources {
            srcDirs = ['code/src/testResources']
        }
    }
```

Vest: Here is the fresh project, generated by Gradle:
Sources in the Module
Look, the `utest` folder is there.

Vest:
Ok, I will update README.md, and go through docs later. Thank you for the tip.

Vest:
What do you think, if I remove files **code/PCGen.iml** and **code/PCGen.ipr** from the repository? They are old, and still use Java SE 5. I think, we don't need them.

karianna:
> What do you think, if I remove files **code/PCGen.iml** and **code/PCGen.ipr** from the repository? They are old, and still use Java SE 5. I think, we don't need them.

That would be appreciated yes!

Vest:
I have updated the PR according to comments from @karianna 
Unfortunately, I don't use Asciidoctor for the document generation, so I have just created a new page using the same stylesheet :)

karianna (approved):

PR #6838, opened 2023-04-11
DATA-4314 - Fix #2

Vest:
Removed the wrong line from _ultimate_combat.pcc, that forced PCGen to interpret support/uc_abilitzcategories_um.lst file as an ABILITY list instead of ABILITYCATEGORY.

Moved the line to the top position.

@BlyatBeauty, this is your DATA, I guess, so you should agree with me :) Probably you overlooked this line.

### DIFF
--- a/data/pathfinder/paizo/roleplaying_game/ultimate_combat/_ultimate_combat.pcc
+++ b/data/pathfinder/paizo/roleplaying_game/ultimate_combat/_ultimate_combat.pcc
@@ -58,10 +58,10 @@ ARMORPROF:uc_profs_armor.lst
 WEAPONPROF:uc_profs_weapon.lst
 
 # Support
+ABILITYCATEGORY:support/uc_abilitycategories_um.lst|PRECAMPAIGN:1,INCLUDES=Ultimate Magic
 ABILITY:support/uc_abilities_class_um.lst|PRECAMPAIGN:1,INCLUDES=Ultimate Magic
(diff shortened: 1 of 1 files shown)


BlyatBeauty:
Hey, good catch! Yeah this fix looks good to me!

Vest:
@BlyatBeauty, I have a short question. Why have you created the LST files in the support/ folder? Does it change anything, or are there any rules?

BlyatBeauty:
@Vest 

Each support file is a reference to another sourcebook. Those other sourcebooks aren't generally necessary for Ultimate Combat to work, but feats from the other books are compatible with Ultimate Combat and without a support file to let PCGen know how those feats are supposed to work together with this book's contents, PCGen won't piece together how the two are supposed to interact.

In the case of Master of Many Styles, a Monk archetype, the archetype allows a player to select Style feats as bonus feats. However, other sourcebooks introduce new Style feats. 
The way the data for Ultimate Combat was previously added in, players could only select style feats from Ultimate Combat as bonus feats. With these support files, now when a user loads another sourcebook, they're able to take the Style feats from those sourcebooks as well when they play a Master of Many Styles Monk.

For players not taking this archetype, leaving those references in the support folder means they can still select Ultimate Combat as a sourcebook for their character without needing to load every other sourcebook Paizo has to offer that includes a Style feat.

Vest:
Thank you. I hope, I will remember this difference :)

PR #6858, opened 2023-05-29
Moved PreMult-parser/-tester/-writer to the plugin folder

Vest:
Corrected failed tests, because they assumed that PREMULT tag is always active. When you run tests individually (e.g. failed ones, there is a chance that PREMULT is not available).
Some minor improvements were suggested by IntelliJ IDEA.
Some typos were fixed.

### DIFF
--- a/code/pluginbuild.xml
+++ b/code/pluginbuild.xml
@@ -8,7 +8,7 @@ STANDARD TARGETS
 
 jar-all-plugins - Generate the plugin jar files
 
- 
+
 -->
 
 <project name="pcgen-plugins">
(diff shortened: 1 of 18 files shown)


Vest:
Some tests were written with an assumption that PREMULT is always available. But it isn't always the case, because it was not explicitly added during this test run.

I know, this change is quite big, and I have decided to introduce it, because these failed tests appear in IntelliJ, if you use it for development.

If you don't mind, I plan to get rid of some warnings, where the raw Class is used. Especially in plugins, but I need this PR to be approved, because my change will be consistent - all parsers/testers/writers will be bundled into plugins properly.

@LegacyKing, if you have time, please check my PR. I apologize for pinging you directly, but I see that you are one of the most active maintainers.

karianna (approved):

PR #6861, opened 2023-06-01
Unused ACVALUE is removed from all sources

Vest:
Java, LST and HTML files, where it was used or mentioned. ACVALUE was used in OSRIC only, but not in the actual book itself.

The deprecation of ACVALUE was mentioned in 2016: 67057b8 (67057b893a514d837d7ec4949f74e92a353c9ee1). PR was #1092.
Many LST files were formatted with tabs to look more consistent.
Corrected navlistindex.html by removing a duplicate <a>.


Vest:
I don't know whether @thpr still updates this repo, but I think it is time to start removing things that didn't work since 2016 :)
If you think the opposite, probably we should consider removing the "deprecated" attribute from ACVALUE. Otherwise, it looks strange - we have the modifier that isn't used at all.

Thank you for your time.

karianna:
@Vest to be clear ACVALUE was only supported by the OSRIC game system?  I'm not sure of PCGen's overall OSRIC support but since the values are referenced in our 3.5 Edition LST files I'd assume they're valid for that edition no?

Vest:
@karianna, I saw the ACVALUE in OSRIC lst files, but I didn’t find anything in the core book itself: https://www.osricrpg.com/files/OSRIC.pdf

The size usually affects the weight and the cost. 

My DM told me (he has broader experience than I), that in some unofficial home rules something similar to the SIZE scaling was used, but it was never printed by Wizards. At least as far is I remember from my 3.5ed experience, we never faced such modifiers depending on size. Only weapons scale, but not AC. This is actually right, because a theoretical fairy would never be hit in a gargantuan full plate :). 

That is why I consider that the change, introduced by thpr is right. This symbol was never used.

LegacyKing:
ACVALUE has been around since the inception of PCGen and has zero ties to OSRIC. If you look at the code, it's modifying the actual value based upon size, which is SRD based.

Vest:
Yes, thank you. Unfortunately, I don't know which SRD edition you were talking about, but let me take 3.5 as an example:
1. weapons, yes, they scale for different sizes: https://www.d20srd.org/srd/equipment/weapons.htm#weaponSize
2. armor, no, only its weight and cost changes: https://www.d20srd.org/srd/equipment/armor.htm#armorForUnusualCreatures (but there is one exception for a tiny size - AC should be divided by 2. I am wondering, how much AC does a tiny buckler provides). But I bet, it is better to add a tiny full plate or buckler into an equipment list for such cases. Otherwise, the size doesn't affect AC at all.

karianna:
@Vest - I think the conclusion is that we'll have to close this PR without merging, unless it's easy for you to revert the removal and only leave the other whitespace changes you made?

Vest:
@karianna thank you for your time, I guess, you can close it as it is. Honestly, I didn't see a reason, why. Based on your comments, I went through the official core rule books, and I didn't find how the size affect AC value. One evidence is the "Fine" armor, and that's it. Other types of armor will be improperly scaled.

My commit was based on the commit from @thpr, who declared this property as deprecated. That's why this token was put into the *.deprecated package. If you want, I can create a PR and move it from the package back.

As another approach, I can add the interface "DeprecatedToken" to ACVALUE, because all deprecated tokens use it (or at least should use it). I have tried this once. However, there is a checker LstUtils.deprecationCheck, that will write a warning message, and you will see dozens of lines, telling you that the ACVALUE token has been used.

So, in other words, I have feeling that we have a deprecated token, but we don't want to accept that it is deprecated and thus warn a user, or remove the token, or we don't want to bring it back, because there is no rule book, where this token can be used.

If it is possible to ask other team members in Discord, I can raise a question there.
What do you think?

karianna:
Let's absolutely discuss on discord, I'm really not to sure on this one but I tend to defer to the data folks who code up the source materials 🙂

karianna:
Merged as per Discord discussion

PR #6875, opened 2023-06-24
Fixed the failed "slowtest": STARTFEATS

Vest:
The test  was failing, because PreMultParser wasn't registered for the string "PREMULT:1,[PREHD:MIN=1],[PRELEVEL:MIN=1]"
Changed the formatting of AboutDialog, because Checkmarx tests were failing.
cleaned and unused property files were updated by the Gradle tasks.


Vest:
@karianna or @LegacyKing - if you have time, please review my PR.
I didn't see that slowtests (or itests) were using in this project. They were explicitly disabled. That's why I didn't notice that one of my old PRs #6858 broke few tests as well.

The reason of this bug is that the PreMult parser must be added only to those tests, where it is required. In the past, it was assumed that the parser is always there (even, if the test doesn't use it).

karianna (approved):

PR #6876, opened 2023-06-27
Fixed the failing test - pcGenGUIElwoodTest

Vest:
There were a lot of changes. Many of them I discussed with @LegacyKing, so I hope he can review this (quite large) PR.
Unfortunately, not all integration tests are fixed with this PR. Currently I have more than 40 failed cases.


Vest:
p.s. this is how this test case can be tested:
`gradle slowtest --tests '*pcGenGUIElwoodTest*'`

LegacyKing (approved):
All OS and Data changes approved. Need Code Review.

LegacyKing:
@karianna can you review and merge if the code looks good? 

Thanks!

code/src/java/pcgen/core/Equipment.java
```diff
@@ -5033,6 +5033,7 @@ else if (sizeDiff == 0)
 			}
 		}
 
+		assert wCat != null;
```
karianna, line 5036: Do we need this assert?

code/src/test/pcgen/AbstractCharacterTestCase.java
```diff
@@ -197,7 +203,8 @@ protected void setUp() throws Exception
 		FactDefinition<?, String> fd =
 				BuildUtilities.createFact(context, "SpellType", PCClass.class);
 		fd.setSelectable(true);
-		SourceFileLoader.processFactDefinitions(context);
+		// Same line appears in finishLoad()
+		// SourceFileLoader.processFactDefinitions(context);
```
karianna, line 207: Remove commented out code?

karianna (approved):
Couple of non Blocking comments

code/src/java/pcgen/core/Equipment.java
Vest: I have decided to add it, because the code block down below fails, as it isn‘t prepared for NPEs. The null appears, when the code cannot find a wield category. I can replace it with a static check (if statement) that fails, when this happens and write into log how to fix it with LST.

code/src/test/pcgen/AbstractCharacterTestCase.java
Vest: I will, didn’t want to remove it, thought somebody might disagree with the change. That is why the comment was added.

PR #6880, opened 2023-07-01
Updated expected templates (csheet files)

Vest:
This long PR fixes approximately 26 XML files located in /csheets folder. There are multiple reasons, why this was needed:
1. in the base template, a trailing /t/t was removed, and this change affected all templates
2. A long time ago FEAT and TEMPLATE tokens were replaced withe FEATHIT/-DAMAGE and TEMPLATEHIT/-DAMAGE (appeared somewhere in 2015)
3. The deity tag was removed for an MSRD character, because she doesn't have deity in the rule book.

p.s. not all characters were updated, because for some of them big blocks of data was modified. I need some time to investigate, why this change appeared.
You can squash this PR to a single commit.

### DIFF
--- a/code/testsuite/csheets/pf_Cleric.xml
+++ b/code/testsuite/csheets/pf_Cleric.xml
@@ -797,7 +797,7 @@ BIO
 			<critical>20/x2</critical>
 			<!-- Should be changed to a variable due to improved crit -->
 		</unarmed>
-
+		
 		<weapon>
 			<common>
 				<name>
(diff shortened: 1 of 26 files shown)


code/testsuite/csheets/pf_Cleric.xml
```diff
@@ -2319,28 +2337,6 @@ BIO
 			<virtual>F</virtual>
 			<category>Special Ability</category>
 		</ability_object>
-		<ability_object>
```
LegacyKing, line 2322: This change is unexpected. I don't see anything that makes this change. Can you validate this is expected?

Vest:
@LegacyKing I have reverted the mentioned commit.

Vest:
I have reverted another commit, where the human trait disappeared.

LegacyKing (approved):

PR #6881, opened 2023-07-03
Fixed: PCClassTest - renamed CLASSTYPE to TYPE in sample data.

Vest:
I have also corrected few typos, and removed an empty line.

p.s. FACT|Abb subtoken still shows warnings. They look like
```
LSTERROR main PCClassLoader:324 Illegal FACT subtoken 'Abb' 'Clr' for pcgen.core.PCClass Cleric (Source: null )
LSTERROR main PCClassLoader:324 Failed in parsing typeStr: FACT Abb|Clr for pcgen.core.PCClass Cleric
LSTERROR main PCClassLoader:324 Illegal Token 'FACT' 'Abb|Clr' for pcgen.core.PCClass Cleric in null
```

@karianna I will appreciate if you give me an advice. The code flow is interesting (but strange to me). We add two FACTs by default: CLASSTYPE and SPELLTYPE. But somewhere in the middle, if we add a FACT|Abb, we have to commit our context, otherwise during the parsing the PCClassLoader couldn't find the FACT Abb. I don't know how to fix this (stupid) warning without an explicit commit. I have expected that if the token was parsed, but not committed yet, it can be used by the parser, because (in theory) we will commit at the end of our parsing step.

If you can play with the test case, please help me with it. You can use PCClassTest.testAddAbility to reproduce the warning.

I remember that I removed an explicit commit a long time ago, because there were two commits during setUp, then one commit in finishLoad. It looked messy to me. But I doubt that that change was related to this warning, because I couldn't find in the code where Abb FACT has ever created. Probably these warnings were always ignored.

### DIFF
--- a/code/src/java/pcgen/cdom/enumeration/FactKey.java
+++ b/code/src/java/pcgen/cdom/enumeration/FactKey.java
@@ -65,7 +65,7 @@ public final class FactKey<T>
 
 	/**
 	 * Returns the FactKey for the given String (the search for the constant is
-	 * case insensitive). If the constant does not already exist, a new FactKey
+	 * case-insensitive). If the constant does not already exist, a new FactKey
 	 * is created with the given String as the name of the FactKey.
 	 * 
 	 * @param name
--- a/code/src/java/pcgen/rules/persistence/CDOMControlLoader.java
+++ b/code/src/java/pcgen/rules/persistence/CDOMControlLoader.java
@@ -42,7 +42,7 @@ public class CDOMControlLoader extends LstLineFileLoader
 
 	public CDOMControlLoader()
 	{
-		//CONSIDER better way to load these?
+		//CONSIDER a better way to load these?
 		addLineLoader(new CDOMSubLineLoader<>("FACTDEF", FactDefinition.class));
 		addLineLoader(new CDOMSubLineLoader<>("FACTSETDEF", FactSetDefinition.class));
 		addLineLoader(new CDOMSubLineLoader<>("DEFAULTVARIABLEVALUE", DefaultVarValue.class));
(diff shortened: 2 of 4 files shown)


LegacyKing:
Are you using the default gamemode 'compatibilityDataControl.lst' file which explicitly defines all three?

```
FACTDEF:CLASS|Abb			DATAFORMAT:String		VISIBLE:YES
FACTDEF:CLASS|ClassType		DATAFORMAT:String		VISIBLE:YES
FACTDEF:CLASS|SpellType		DATAFORMAT:String		VISIBLE:YES	SELECTABLE:YES
```

karianna:
@Vest Give what @LegacyKing suggests a go and let's see if that helps.

karianna (approved):

Vest:
I have analyzed the flow, no, the compatibility LST file is not loaded. AbstractCharacterTestCase (base class for many tests) creates the environment in Java only. Here is an example, how sizes are created:

```java
// Added LoadInfo to avoid warnings during tests
LoadInfo loadable =
	gamemode.getModeContext().getReferenceContext().constructNowIfNecessary(
		LoadInfo.class, gamemode.getName());
loadable.addLoadScoreValue(0, BigDecimal.ONE);
loadable.addLoadMultiplier("LIGHT", 1/3.0f, "", 0);
loadable.addLoadMultiplier("MEDIUM", 2/3.0f, "", -3);
loadable.addLoadMultiplier("HEAVY", 1.0f, "", -6);
```

I think, this speeds up the test run, because you don't have to parse all books and all LST files.
I understood how few Java objects are created during the test run, but FACTs is a challenge for me. If you merge my PR, we can live with these warnings, but I will return back to them, after all itests are fixed.

PR #6885, opened 2023-07-06
Fixed: Updated expected templates (csheet files)

Vest:
There are two simple commits that fix the failed integration tests.
@LegacyKing you were the one, who agreed that traits should disappear from XML. At least, this is expected for you.


Vest:
p.s. you can squash the PR, if you approve it.

PR #6890, opened 2023-07-08
Fixed integration tests related to Startfinder

Vest:
Here are few words about the changes:

unfortunately, there was a regression introduced by me. There are details in Discord, but high level: it seems that Tom was using **WeakReference** in **AbstractReferenceManufacturer** in several places. I don't know the reason, but maybe he wanted to free memory. It causes a weird behavior, when the skill **CSKILL** with **TYPE=Profession** is parsed, the code tries to find this skill in the typeReferences field, and it finds the weak reference pointing to null. That is why SF tests failed randomly. Sometimes the skill is found, sometimes is not.

I'd switch from WeakReference to either SoftReference, or better to a strong one. I don't understand why weak reference was needed. It is usually used for cache, where items might expire. Current code doesn't trust these references, it verifies with `.get() == null` every time, when the value is needed. I think, this is redundant.

Now good news, the integration tests are almost working. Now only data import tests fail.
Here is the result of the latest :slowtest run:
latest slowtest result
All XML files seem valid and consistent now.


Vest:
p.s. please squash this PR again. It has some temporary garbage.

LegacyKing (approved):
Looks good to me.

karianna:
Yeah WeakReferences don't strictly need to be used here.  There's other optimizations we could apply first.

Vest:
Regarding the commit: [9f835e1](https://github.com/PCGen/pcgen/pull/6890/commits/9f835e16079b7dc465ee1ba7d669abe6de236f26). I used an idea from SO, and decided to temporarily override the **finalize** method of CDOMGroupRef:

```java
	@Override
	protected void finalize() throws Throwable {
		System.out.println(this + " collected");
		super.finalize();
	}
```

This is what I got, after I launched the test pcGenGUISFmechanicTest again:
```
INFO main SourceFileLoader:504 Loading game Starfinder and sources [Starfinder RPG Core Supplements].
CDOMTypeRef Skill Skill [Profession] collected
INFO main CharacterManager:194 Loaded character Yo-yo 
```
My blind guess was right - the skill "Profession" was collected by GC. That's why the weak reference returned me a null. Yes, this code smells, I'd refactor it in future.

PR #6896, opened 2023-07-10
Fixed a bug, when the OutputPanel couldn't show the file chooser dialog

Vest:
This happens because the underlying code works in Swing threads only. Now the entire content of the panel is running in JavaFX thread.

Few empty lines were also removed.

As a symptom, the user sees the error in the logs:

```
023-07-10T00:10:47.285913900 SEVERE JavaFX Application Thread LoggingUncaughtExceptionHandler:32 Uncaught error on thread  Thread[JavaFX Application Thread,5,main] - ignoring
pcgen.gui3.GuiAssertions$WrongThreadException: expected NOT to be on JavaFX thread - actually on: JavaFX Application Thread
        at pcgen.gui3.GuiAssertions.assertIsNotJavaFXThread(GuiAssertions.java:59)
        at pcgen.gui3.GuiUtility.runOnJavaFXThreadNow(GuiUtility.java:56)
        at pcgen.gui2.prefs.OutputPanel.onOutputSheetHTMLDefaultButton(OutputPanel.java:291)
        at javafx.base/com.sun.javafx.event.CompositeEventHandler.dispatchBubblingEvent(CompositeEventHandler.java:86)
        at javafx.base/com.sun.javafx.event.EventHandlerManager.dispatchBubblingEvent(EventHandlerManager.java:234)
        at javafx.base/com.sun.javafx.event.EventHand
```
The PR can be improved further, if (after) all UI parts are migrated to JavaFX. Plus the file chooser is not a dialog window, because JavaFX doesn't have a Scene/Stage that can be used as a parent. The entire window is managed by Swing. That's why keep an eye on an unclosed dialog.

p.p.s If the user doesn't select a file, or selects a wrong one, the false/wrong input is not put into properties, and thus not reflected on UI. Previously the input was cleared by an empty value.

### DIFF
--- a/code/src/java/pcgen/gui2/prefs/OutputPanel.java
+++ b/code/src/java/pcgen/gui2/prefs/OutputPanel.java
@@ -45,6 +45,7 @@ import javafx.scene.control.Label;
 import javafx.scene.control.TextField;
 import javafx.scene.layout.GridPane;
 import javafx.stage.FileChooser;
+import javax.swing.SwingUtilities;
 import org.apache.commons.lang3.BooleanUtils;
 
 /**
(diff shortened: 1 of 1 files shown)


Vest:
There are no JUnit tests that can test my fix, because there is a mix of JavaFX and Swing frameworks, and I didn't find a simpler way (without rewriting many parts) how to test my fix. 
Sorry, a manual monkey-testing is required.

PR #6917, opened 2023-08-24

Vest:
It breaks the image export. In addition to this change, other changes were added:

- PortraitTokenTest now uses JUnit5 assertions
- TestHelper got a helper method "evaluateToken", that was used by other test-classes. This is a small refactoring to avoid the duplicated code.


karianna:
@Vest broadly looks good but there seems to be a lot of other things mixed up in this PR - deliberate?

Vest:
@karianna my fault. I haven’t checked the PR. These changes arrived because my master branch was outdated with PCGen’s branch. 

I will recreate it.

PR #6918, opened 2023-08-24

Vest:
The image export was broken.

TestHelper got a helper method "evaluateToken", that was used by other test-classes. This is a small refactoring to avoid the duplicated code.

This commit is specific for 6.08 branch.


Vest:
@LegacyKing here is my commit for 6.08. The code compiles, shouldn't break the build.
Not all modified tests could be tested with 6.08, because they were probably explicitly commented out by someone, who was trying to fix "slowtests" in old times.

PR #6922, opened 2023-08-28

Vest:
It breaks the image export. In addition to this change, other changes were added:
- PortraitTokenTest now uses JUnit5 assertions
- TestHelper got a helper method "evaluateToken", that was used by other test-classes. This is a small refactoring to avoid the duplicated code;
- The token object is final;
- Trimmed spaces in the modified files.

### DIFF
--- a/code/src/java/plugin/exporttokens/PortraitToken.java
+++ b/code/src/java/plugin/exporttokens/PortraitToken.java
@@ -40,7 +40,7 @@ import org.apache.commons.lang3.StringUtils;
  * The Class {@code PortraitToken} supports the PORTRAIT
  * token and its and PORTRAIT.THUMB variant.
  *
- * 
+ *
  */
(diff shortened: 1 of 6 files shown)


Vest:
Hello @karianna,
here is the same PR as in #6917, but I have made few changes (because I was moving JUnit4 code to JUnit5) that I have overseen.
In my machine `gradle slowtest` doesn't fail at all, but you can verify this again.

Please merge the code with the squashed commit. Nobody needs to see my small fault. It doesn't bring any value.

Regarding the changes - I have decided to refactor the code in this PR, because I didn't want to add another duplicate (where I add a code to export tokens into a file), and then refactor the code. The method that I use is required to test a code "if token is encoded write it as encoded, otherwise write it as it is". To check the "encoded" flag, I need: writer and filter. This is the method that was used in other test cases.

PR #6934, opened 2023-09-15
Replace java.net.URL with java.net.URI in LstFileLoader.java

Vest:
The aim of this PR is to prepare to migration from Java 17 to Java 21 (LTS) that might happen in future. JDK deprecates URL constructors, and replaces them with URI.

The next step of this PR was to improve error messages (some formatting) for LST files processing. Now all LST files have UTF-8 (some of them were ANSI or UTF-8-BOM), and this forced us to use a 3rd party (Apache) library to handle different encodings.

Current tests (integration, unit etc) are "green" (at least on my machine), and all LST files don't have warnings related to a wrong/unexpected encoding. So in theory, if you create a new LST file, and its codepage doesn't match to UTF-8, you will get either an error (if it is completely weird codepage), or a warning (if it has a BOM-character).

Please review and merge this PR, but squash the changes. 

Thanks

p.s. it is just a thought, but I was thinking about a parallel handling of LST files using Java Streams. The NIO API suits this purpose better than the old API that Apache provided. Plus I want to get rid of redundant dependencies, if I can.

@karianna


Vest:
I don't want to confuse people, but I was wrong - JDK-4508058, was fixed and then the change was reverted. The official statement is to detect BOM yourself, and handle this situation as required (e.g., I just cut the first character from the string and warn the user).
Please excuse me for this confusion :(

karianna (approved):

PR #7013, opened 2023-11-26
DATA-4136: Maneuverability spelled wrong

hathawayd:
DATA-4136: Maneuverability spelled wrong causing flight skill to calculate wrong.


hathawayd:
It might also be worthwhile to check why Manueverability was calculating anything at all.

karianna (approved):

PR #7025, opened 2023-12-07
DATA-4354: Saurian Shaman was adding the Fire instead of Strength

hathawayd:
DATA-4354: Saurian Shaman was adding the Fire domain when it should have added the Strength domain. It looks like the book I have lists Fire, but both aonprd and d20pfsrd list Strength, my guess is it got changed in newer printing?


karianna:
@LegacyKing How do we normally deal with Book vs SRD conflicts?

LegacyKing:
We go with the most up to date.

LegacyKing:
Page 39—In Saurian Shaman archetype, add
“(Archetype)” after the title “Saurian Shaman”. In
the Nature Bond ability, in the second sentence,
change “Fire” to “Strength”.

PR #7077, opened 2024-04-05
Move to Java 21

karianna:
- **Initial cut for Java 21 support + lots of debug**
- **Update PMD to 7.0.0-rc4**
- **PMD now passes**
- **Checkstyle now passes**
- **grdle 8.5 for full Java 21 support**
- **Use official monocole**
- **Initial cut for Java 21 support + lots of debug**


Azhrei:
Lol, you beat me to it by 19 minutes!  I was just about to submit a PR that finally builds everything under Java+JavaFX 21.0.2! 👍

Azhrei:
I just checked the GluonHQ.com page and it appears that they have an SDK and jmods for JavaFX 22 and 23ea on linux-aarch64, but not v21.  Strange, since v17 _is_ included...

You probably ended up on this same SO page https://stackoverflow.com/questions/78263233/javafx-21-for-linux-aarch64 but they do mention that 21.0.1 is in Maven, just not 21.0.2, with the statement that not every minor release makes it to Maven.

Honestly, I don't think it's much of a stumbling block.  The `linux-aarch64` platform is just Asahi Linux right now, as it's the only distro that boots & works on Apple Silicon, AFAIK.  Kind of a niche market.

One of the things that I'd like to see in the (future) automated CI/CD system is a dynamic search to determine what the latest release is for a given platform, then constrain it by the _major.minor_ release number defined in `build.gradle`.  Thus, the config file could specify `21.0` and it would grab whichever patch level is the latest for the platform being built.

Azhrei:
Oops, I was wrong.

I checked the Maven repo directly at https://central.sonatype.com/artifact/org.openjfx/javafx and the dropdown box at the top _does include_ 21.0.2, so it's just the GluonHQ page that doesn't mention JavaFX 21 LTS support.  I specifically checked the `.pom` file and I see two entries for `linux-aarch64`, one with monocle and one without.

Vest:
@Azhrei SO said that this has something related to LTS. They don’t want to support arm for Linux. 
I have temporarily cut this architecture from the gradle build. I haven’t heard anything about Linux users who use offen on Raspberry PIs :)
It is a sad decision, but it is easier to drop this combination instead of rewriting many things (such as build JavaFX from scratch, or use another JDK distro).

karianna:
Yeah I've pinged Johan Vos (who runs Gluon) and they're kind of always resource crunched, so sometimes those point releases don't get to Maven later.  I caught the other helpful PR to merge into this one which I'll do (or manually cherry-pick) and update this again over the weekend.

karianna:
FYI Bellsoft have https://bell-sw.com/liberica_eula/ as a EULA which I'm not comfortable with on behalf of this project (as they reserve the right to change it any time)

karianna:
There are some fixes required around slowTests (36 failures, although 34 of thode are from a new Java 21 restriction on using the Security Manager)

karianna:
With the latest merge from @Vest we have a working Mac build so I'm calling this good (or good enough). Will merge and see if we can get a full release from the build server to work.

PR #7078, opened 2024-04-05
Update pcgen.sh

Azhrei:
Added directory change to code/pcgen.sh to fix double-clicking from the macOS Finder GUI.

### DIFF
--- a/code/pcgen.sh
+++ b/code/pcgen.sh
@@ -1,6 +1,20 @@
 #!/bin/sh
 set -e
 
+if [ ! -e ./pcgen.sh ]; then
+    # We're not in the directory where the script lives.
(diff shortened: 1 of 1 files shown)


code/pcgen.sh
```diff
@@ -1,6 +1,20 @@
 #!/bin/sh
 set -e
 
+if [ ! -e ./pcgen.sh ]; then
```
karianna, line 4: Can you add a comment explaining the Mac OS X use case here?

karianna (changes requested):

code/pcgen.sh
Azhrei: Done.

karianna (approved):

PR #7084, opened 2024-04-08
Fixes problem of using deprecated SecurityManager

Azhrei:
Only fixes the `build` task, there are still 36 other uses of SecurityManager in the various tests.


Azhrei:
Hm.  I thought I got this right, but maybe there are commits in here that shouldn't be.  I'll investigate.

Azhrei:
Ugh.  Some of my changes to make this work on macOS were included, and to _only_ build for mac-aarch64.  There's also an upgrade to Gradle 8.7 in here.  I need to learn how to revert individual files from a commit and I'll update the branch to fix this PR.  Sigh.

karianna:
@Azhrei You may want to fire the pull requet against my java21 PR

Azhrei:
> @Azhrei You may want to fire the pull requet against my java21 PR

Yes, will do.

I was going to try to do some editing here on GH to eliminate the two conflicts, but it's likely easier to delete my personal fork and pull from this `master, then apply my 3-file patch so I can redo the PR.

I love Git when everything works like it's supposed to, but when I screw something up, I do a _really_ good job. 🙄

PR #7085, opened 2024-04-08
6.08: Upgraded gradle wrapper from 7.1 to 8.7

Vest:
Upgraded gradle wrapper to 8.7 to support Java 17 (and actually other Javas as well).
I have executed the command:
```
gradle wrapper --gradle-version 8.7
```
Actually, maybe this is a big risk, because (theoretically) if the version 6.08 was built using Java 11 (checked build.gradle). The build error was thrown, because @LegacyKing  was using Java 17 instead of 11th on his host machine.

In other words: gradle 7.1 supports Java up to 16. Starting from gradle 7.3, Java 17 is supported.

@karianna @Azhrei I am not sure, if it was wise from me to upgrade gradle to 8.7 at all, maybe I should upgrade it to 7.3 only. If "the second", then you can execute the following command:
```
gradle wrapper --gradle-version 7.3
```
I am not sure, if we will have any build errors. On my machine 8.7 seemed running fine, but there are many warnings in the code, as it is very old and not refactored (yet).


Azhrei:
The main issue with NOT updating Gradle is that as plugins themselves are upgraded, the builds will fail with "Unable to load class, file major version ##", b/c an older Java version of Gradle will not be able to parse newer .class files.  I find this particularly annoying; one of my pet peeves, I guess...

Vest:
No, I bet it isn't plugins, but the "build system". When you run gradle, you use the current Java installation. And for LegacyKing it is Java 17. Gradle 7.1 must use something close to Java 11 (e.g., Java 11).

karianna:
I'll take a look tonight - Java 17 can )of course) compile to 11 as a target so let's see if we can support that case. It's always a bit funky when you have a Java version running your build tool and targeting something else.

karianna:
Syntax of the build scripts have changed a lot from 7.x to 8.x - going to 7.3 is safer (I'll submit a PR for this shortly if all tests pass)

PR #7184, opened 2024-10-02
temporarily change the bash working directory

James-Rocker:
# what

On Linux, attempting to run the `pcgen.sh` outside the program folder results in `error: Unable to access jarfile ./pcgen.jar`. This was a problem for making `desktop` files. Fix was to change the working directory of the file


Vest:
@James-Rocker, I don't have linux, but do how you start the project?
Do you build it from sources with Gradle to get a *.deb or *.zip file and then you run it from the console? Please give me some hints, I understand your fix, but I want to get the same error.

James-Rocker:
> @James-Rocker, I don't have linux, but do how you start the project? Do you build it from sources with Gradle to get a *.deb or *.zip file and then you run it from the console? Please give me some hints, I understand your fix, but I want to get the same error.

Sure so my methods of running was as follows

1. download the latest zip from the releases under https://github.com/PCGen/pcgen/releases
2. unpack the zip folder
3. run the unpacked `pcgen.sh` file in the terminal from my `$HOME` location

Unless my current working directory is the installation folder, this causes the script to fail.

karianna (approved):

karianna:
@Vest I assume there might be a concern when this runs in a .deb or .rpm?

Vest:
Yes, this is what I assume as well. That’s why I cannot judge this PR easily.

karianna:
> Yes, this is what I assume as well. That’s why I cannot judge this PR easily.

I have an ubuntu in a VM I can try a deb on, let me see if I can produce one of those.

karianna:
Hmm, it doesn't actually look like we produce .deb or .rpm installers at this stage

karianna:
OK, given we don't actually create Linux installers - let's land this and deal with that later

Issue #7079, opened 2024-04-07
[BUG] SecurityManager is deprecated for removal and spotbugs complains about it during build


Azhrei:
**Describe the bug**
There are ~36 situations where a SecurityManager is being constructed in the current code base.  This class is deprecated and will likely go away very soon (it was deprecated prior to Java 17, then warnings started being issued, and it's likely to go away no later than 24).

**To Reproduce**
Attempt to run `./gradlew spotbugs` and watch the error messages scroll. 🙂

**Expected behavior**
The build should finish without errors.

**Desktop (please complete the following information):**
 - OS: Any
 - Version Any

**Additional context**
All references to the `SecurityManager` are from the `SystemExitInterloper` (or is it `Intercepter`?) class.  This class is a kludge to allow a test to call `argparse4j.parseArgsOrFail()` and for the error to be caught.  

However, the latest `argparse4j` has a method named `parseArgs()` that throws an `IllegalArgumentException` instead.  This is how the tests should be written.

(I have made proof of concept changes to `Main`, `CommandLine`, and `CommandLineTest`, but I have not pushed them to my repo yet and won't be able to do so for another 24 hours or so.  It's at https://github.com/Azhrei/pcgen under the `sm-fix` branch.)

Azhrei:
I can't seem to assign this to myself (GitHub isn't providing the panel on the right that usually allows labels, assignations, projects, and so forth).  But I should have time this coming week to work on it.  I'll do a few and try to give an ETA on completion.

Azhrei:
Status update:  I found two NPEs during `slowtest` that were the result of my change to the base code.  Both will be included in this eventual PR as they're part of this same problem.

PR #7247, opened 2024-12-24
TEST to look at Gh build vs tip

karianna:


Vest:
Please wait (as usual, I cannot tell until when) and don't merge this PR

karianna:
Yeah I assume we won't merge this PR, I figured it was just the easierst way to look at a diff and see if we could spot anything.

karianna:
Super interesting those tests fail for you but not for me locally!

karianna:
> If you use your version/branch, but not mine, I bet you have this PR: #6975 It comments the most important part, when the character sheet is exported. I have uncommented it, and now the tests started failing. (mostly because LegacyKing corrected some data-files).

OK, so just a data fix required then?

Vest:
Not only:
1. DATA changes caused slowtests to fail
2. Some XML files were not trimmed
3. The template was trimmed (no leading whitespaces)
4. That's all. I am waiting for the next build :smile:

Vest:
Two commits above this comment are related to #7079. Security manager is mostly used to intercept System.exit calls;
I have created required classes to replace System.exit with my proposed class. 

I hope, you will like it.

karianna:
Looks like it all passed!  Shall we take out of Draft?

Vest:
It is unfinished. Do you insist? I need:
1. remove security manager
2. replace all System.exit calls 
3. I have disabled NSIS, but I have not moved to jpackage a windows build yet. 

If you don’t need release builds for windows for now. I would implement my PR simply to have indicators that tests are green. 

Are you sure? Because I am not. 

You can actually merge it, but I only if you plan to work with the new codebase. E.g., continue improving GH actions. If not, I would keep it for a while draft.

Vest:
The commit 59833c5 addresses the issue #7079. We don't use SecurityManager directly (maybe its exceptions are still thrown here and there), but I have created a custom interceptor, called pcgen.util.ExitInterceptor. It is used in several test-classes.
@karianna, if you want, you can link this PR to the mentioned issue #7079 
@Azhrei, I know, you were assigned to this issue. Since I haven't heard any news from you, I think, you were busy with your life. If you are curious about my implementation, can you please review this PR? Not entirely, but you can look at the class: GracefulExit, ExitInterceptor, ExitFunction and few tests that I modified.

I don't know, if I should move my three "exit"-related classes/interfaces into a nested package (just to reduce litter in utils package). If you think, it will be a good idea, I will do this. Something like: "pcgen.util.exit".

Thank you all.

Vest:
Code coverage of CommandLineArgs

Code coverage of GracefulTest

Azhrei:
Thanks.  I actually have some time today, so yes, I'll take a look.  It'll be a couple hours from now, most likely.

Since intercepting System.exit() is a no-no, I'm very curious to see how you did it!

Vest:
I don’t intercept it. I have got rid of it. There is only one place, where I allow it - GracefulExit :smile:
The rest is filtered out with CheckStyle. If you accidentally write System.exit, it won’t pass CheckStyle (a simple regexp within all sources - dumb, but solid). You can bypass it, if you overcome my regexp (e.g., use reflection + build the string at runtime). 
thanks.

karianna:
@Vest Will happily wait until you're done, thanks for tackling this!

code/src/java/pcgen/persistence/lst/LstFileLoader.java
```diff
@@ -98,40 +96,35 @@ public static String readFromURI(URI uri) throws PersistenceLayerException
 							"The file %s uses UTF-8-BOM encoding. LST files must be UTF-8".formatted(uri));
```
Azhrei, line 96: Error message should be in a properties file...
Also, at a higher level, a file that starts with a UTF-8-BOM actually *is* a UTF-8 file, so this whole area might need reworking.

code/src/java/pcgen/persistence/lst/LstFileLoader.java
```diff
@@ -98,40 +96,35 @@ public static String readFromURI(URI uri) throws PersistenceLayerException
...
+			} else
 			{
-				// Just to protect people from using web
-				// sources without their knowledge,
-				// we added a preference.
+				// Just to protect people from using web sources without their knowledge, we added a preference.
 				ShowMessageDelegate.showMessageDialog("Preferences are currently set to NOT allow\nloading of "
-					+ "sources from web links.\n" + uri + " is a web link", Constants.APPLICATION_NAME,
-					MessageType.ERROR);
+								+ "sources from web links.\n" + uri + " is a web link", Constants.APPLICATION_NAME,
```
Azhrei, line 114: Another properties file template string...?

Azhrei (changes requested):
Wow, big changeset. 🙂  A lot changes not directly related to `GracefulExit`, but I mostly skimmed those.

I like the idea of using GE as a front-end on `System.exit()`.

It all looks good to me!

code/src/java/pcgen/core/PlayerCharacter.java
```diff
@@ -5477,7 +5477,13 @@ public int processOldMaxDex()
 	{
 		final int statBonus = (int) getStatBonusTo("MISC", "MAXDEX");
 		final Load load = getHouseRuledLoadType();
-		int bonus = (load == Load.MEDIUM) ? 2 : (load == Load.HEAVY) ? 1 : (load == Load.OVERLOAD) ? 0 : statBonus;
+		int bonus = switch (load)
+		{
+			case MEDIUM -> 3;
```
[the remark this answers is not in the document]
Vest: Nice catch, but no :) I have found this bug, when my integration tests failed:

Encumbrance Effects

Max Dex must be +3

code/src/java/pcgen/persistence/lst/LstFileLoader.java
Vest: Done!

code/src/java/pcgen/persistence/lst/LstFileLoader.java
Vest: Nope.
I don't honestly know what to deal with the error here. I have explained you that we had Apache Commons something (maybe String commons, I don't remember exactly) to work with BOM-files. I wanted to optimize the build, and I was trying to get rid of dependencies that are not that important.

When I analyzed all LST files, I found out that we had BOM in (maybe) 10 files, not more. It was easier for me to remove BOM and write code that fails/warns the user, if he gives an unsupported file.

LegacyKing was the first one (we don't have many data-contributors), who stepped this mine first. He was disappointed with the error, and I had to write this code just for him.

My thoughts are to add a Gradle task to go through all LST (or DATA) files and scan for BOM. Either convert them to non-BOM, or fail the build or whatever. 

If I don't throw an error, a wrong LST will be pushed to the repo (I don't want this). If I throw an error, probably this build will fail and somebody else will be unhappy. Maybe I should convert LST files silently, but Karianna didn't like when "gradle test" modified .properties files even when nothing was changed. Maybe I'd better write an integration test that will fail, and after this PR is ready we will be able to see in the pull request, who is trying to break the build. Maybe I will also add a task such as "checkLst" to Gradle, so it will convert broken files to good ones. And in the test exception, I will write something like "BOM detected. Test failed. Use gradle checkLst to repair".

Azhrei: Sounds like you've got it under control. I wasn't aware of the backstory. 👍

Vest:
@karianna please check my PR: https://github.com/Vest/pcgen/pull/1
Do you like it?

karianna:
> @karianna please check my PR: [Vest#1](https://github.com/Vest/pcgen/pull/1) Do you like it?

Looks great!

Vest:
Ok, I'd say that we have something, that I can call "release": https://github.com/Vest/pcgen/releases/tag/v0.0.8

**Attention:** I am not a professional devops. So, I am experimenting with my workflow. Really. So, it can build releases now. I have tested a mac release only. The application starts. It isn't signed, but it can run.

My thoughts - delete NSIS completely. But I have to verify windows & linux builds somehow.

@karianna

karianna:
> Ok, I'd say that we have something, that I can call "release": https://github.com/Vest/pcgen/releases/tag/v0.0.8
> 
> **Attention:** I am not a professional devops. So, I am experimenting with my workflow. Really. So, it can build releases now. I have tested a mac release only. The application starts. It isn't signed, but it can run.
> 
> My thoughts - delete NSIS completely. But I have to verify windows & linux builds somehow.
> 
> @karianna

Is it also possible to create .deb linux distro files?  I can test on an Ubuntu system (although it'll be Arm based), I can also test Windows (Arm) but I suspect the exe file is for x86-64?

Azhrei:
It is.  GitHub makes a number of environments available for performing the build.  x86_64 is supported for Windows and Linux, and arm64 for Linux and macOS (not sure about Windows).  The Debian (including Ubuntu) and Red Hat (not sure about Fedora) Linux distros are supported.

It's not difficult to create a matrix of supported architectures and supported platforms, then let GitHub build ALL OF THEM!  Obviously, you'd want to be careful about triggering that one!

LegacyKing:
Where do we stand with this?

PR #7312, opened 2025-04-20
Fix code 1494

lordiceshard:
Added logic to allow charges to be edited from the Inventory>Equipment menu. Also added logic to disable the EditCharges menu item for items that lack charges in both the Purchase and Equipment menus.


mertonmonk:
Nice clean code! I clicked on approve and it's running the codeql check.

PR #7319, opened 2025-04-25
Info export

mertonmonk:
support INFO:DESC on spells and other objects subclassed from PObject and move getInfoToken from Token to PlayerCharacter to make it universally available in GUI as well as export functionality.


Vest:
@mertonmonk you did a great job, but in case if you want to contribute more, please execute ```gradle checkstyleMain```. I am correcting all warnings, so no offense, just keep this in mind.
thanks

code/src/java/plugin/exporttokens/SpellMemToken.java
```diff
@@ -258,7 +258,9 @@ else if ("DURATION".equals(aLabel))
 						}
 						else if ("DESC".equals(aLabel) || "EFFECT".equals(aLabel))
 						{
-							String mString = aPC.parseSpellString(selectedCSpell, aPC.getDescription(aSpell));
+							String mString = aPC.getInfoToken(".INFO.DESC", aSpell);
+							if (mString.equals(".INFO.DESC"))
```
Vest, line 262: @mertonmonk 
CheckstyleMain doesn't allow IFs without curly brackets, because the code might be misleading. Should append be within IF, or not?
What variant you meant?
image
or
image
thanks

Vest (commented):
Please answer to my question, because I have to correct your code.

code/src/java/plugin/exporttokens/SpellMemToken.java
mertonmonk: The append should be outside of the if statement so it appends mString whether it was re-defined within the if statement or not.

mertonmonk: Thanks for pointing out the gradlew target.  It's been quite a while since I did much coding for PCGen, so there is a lot of new code and I'm not particularly familiar with gradle.  I wish I had the time and energy to pull out all the interfaces - I think code should be simple so it's easier for newer developers to understand.

PR #7339, opened 2025-05-27
Fix for empty Ranks column on the Skills tab

Azhrei:
A renderer was being created that wasn't returning a usable TableCellRenderer.  By not registering a "custom" renderer, the default provided by JSpinner is used -- and that one works fine.

There are other references to the SpinnerRenderer that I have not touched but which also may need to be updated.  They are:
* PostLevelUpDialog.java:53
* PostLevelUpDialog.java:134
* EquipmentModels.java:378
* EquipmentModels.java:487
* StatTableModel.java:325

I'd like some eyeballs on both this patch and those other references.  I don't know how to get PCGen through those other code paths, so someone more familiar with PCGen should take a look at those areas to see if the spinners are working properly.

### DIFF
--- a/code/src/java/pcgen/gui2/tabs/SkillInfoTab.java
+++ b/code/src/java/pcgen/gui2/tabs/SkillInfoTab.java
@@ -123,9 +123,6 @@ public class SkillInfoTab extends FlippingSplitPane implements CharacterInfoTab,
 		setOrientation(VERTICAL_SPLIT);
 		setResizeWeight(0.70);
 
-		JSpinner spinner = new JSpinner();
-		spinner.setEditor(new JSpinner.NumberEditor(spinner, "#0.#")); //$NON-NLS-1$
-		skillTable.setDefaultRenderer(Float.class, new SpinnerRenderer(spinner));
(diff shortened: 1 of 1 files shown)


karianna:
@Azhrei What does it mean by "...not a useable TableCellRenderer"?  I'm asking becasue I wonder if the fix should be inside SpinnerRenderer

Azhrei:
The TableCellRenderer is supposed to do certain things when setting up the component for the JTable to use.  The supplied SpinnerRenderer isn’t accomplishing that (things like setting up clipping rectangle, filling in background color, etc).  By removing it from the `initComponents()` method, the JTable uses the `DefaultTableCellRenderer` (I think I have that name correct).  The default renderer for the JSpinner works just fine.

There doesn’t seem to be any need for the SpinnerRenderer for skill ranks at all — unless there is a plan to change the view associated with spinners…?  The other code references I mentioned must be accessible somehow … I’d like to have someone use the GUI until those code paths are actually used to ensure they do/don’t have issues with spinners.  I did look at the spinners used in the Inventory | Equipment tab and they worked fine, so maybe I tested the EquipmentModel references already?

PR #7388, opened 2025-08-01
[HELP-228] PF B2 Familiars not populating

Nylanfs:
COMPANIONLIST doesn't like PREALIGN


karianna:
Do we need to fix this in code so the PREALIGN works?

PR #7556, opened 2026-05-11
refactoring gradle.build: Use gradle's task parallelization and caching to speedup builds

Vest:
## Summary

Migrates the PCGen build to JPMS, replaces the Xalan XSLT processor with Saxon-HE, swaps the runtime/jpackage plugin to `org.beryx.jlink`, rewrites the JDK download/extract for cross-platform parallel builds, and enables Gradle's configuration cache.

## Java Platform Module System (JPMS) migration

- Adds `code/src/java/module-info.java` declaring the `pcgen` module with explicit `requires` (java.desktop, java.xml, javafx.*, PCGen.base, PCGen.Formula, Saxon.HE, FOP, Spring, etc.)
- Drops the deprecated `jdk.jsobject` module requirement.
- IntelliJ module visibility.
- gui3 sub-packages opened for the reflection that FXML loading requires.

## XSLT processor: Xalan → Saxon-HE

- Drops `xalan:xalan` and `xalan:serializer` from the FOP exclude block; adds `net.sf.saxon:Saxon-HE:12.9`.
- Across ~200 `.xslt` files in `outputsheets/`: replaces the Xalan-specific namespace `xmlns:xalan="http://xml.apache.org/xalan"` with the portable EXSLT namespace `xmlns:exsl="http://exslt.org/common"`.
- Wraps `substring-before(...)` in `number(...)` where the result is used in arithmetic — Saxon is stricter than Xalan about string-to-number coercion in XSLT 1.0.

## Runtime image: org.beryx.runtime → org.beryx.jlink

- Master uses the legacy `org.beryx.runtime` plugin to drive jpackage. This branch swaps to `org.beryx.jlink`, which is the maintained successor and integrates with JPMS module graphs.
- macOS app bundle: working directory set correctly via the `MacDirLauncher` script swap, gui3 packages opened for reflection.

## Cross-platform JDK download

- Rewrote JDK download/extract as **per-platform parallel tasks**.
The old `downloadJRE` / `downloadJavaFXModules` used imperative `doLast` blocks that were incompatible with the configuration cache, and `extractJDKs` was effectively broken (its checks ran at configuration time and always reported `NO-SOURCE`). The replacement is declarative `Download` + `Copy` tasks per platform with proper up-to-date checks.
- Output directory renamed `jre/` → `jdks/` to reflect that we now pull full JDKs (jlink and jpackage need `javac` and the `jmods` directory, not a stripped JRE).
[...]

PR #7558, opened 2026-05-17
Optimize build and CI pipeline, remove redundant or dead code

Vest:
## Summary

Pure cleanup of the Gradle build script and CI workflows on top of `kar_gh_build_gradle_deps`. No application code changes — every commit either removes dead code, fixes a latent bug, or makes the build faster / more debuggable.

20 small, focused commits. Net: **+125 / −477** across 8 files.

### Build performance

- **Parallel test forks** — `test` and `itest` now use `maxParallelForks = max(1, processors / 2)` instead of running
serially. Big win on multi-core runners.
- **`testcommon` promoted to a dedicated source set** — was duplicated as a `srcDir` inside `test`, `itest`, and `slowtest`, causing the same classes to be compiled three times.
- **JUnit BOM** — replaced ad-hoc `junit-jupiter-*` version pins with `platform('org.junit:junit-bom:6.0.3')`. JUnit 4 and Vintage retained explicitly because ~870 legacy tests still use the JUnit 4 API.
- **TestFX/Monocle JVM args scoped to `test`** — they were applied to every `Test` task via `tasks.withType(Test).configureEach`, which pulled headless-toolkit args into integration and slow tests that don't use TestFX.
- **Loop-generated per-game `inttest` tasks** — `sfinttest`, `pfinttest`, `rsrdinttest`, `srdinttest`, `msrdinttest` collapsed
from copy-pasted definitions into a map-driven loop. Names preserved because CI references `pfinttest`.
- **`converterJar` refactor** for clarity.

### Dead code removal

- `commitFile()` calls in `updateVersionRelease` / `updateVersionToNext` (the function was never defined — these
tasks would have thrown if invoked).
- `testZip` task (never depended on).
- `jreImage` and `programDistsImage` copySpecs (never applied; the latter pointed at a `build/launch4j/pcgen.exe` path that hasn't existed since the jpackage migration).
- ~233 lines of NSIS installer scaffolding in `release.gradle` (`nsisBaseFolder`, `baseLibs`, `lib32`/`lib64`, `pdfLibs`,
`basePlugins`, `gmgenPlugins`, `genDataList` task, etc.). NSIS was superseded by jpackage long ago; this scaffolding was dead.
- Commented-out repository declarations.
- Unused `localOnly` property.

### CI workflows (`.github/workflows/`)

- **CodeQL action versions** — `actions/checkout` v3 → v4, `github/codeql-action/{init,autobuild,analyze}` v2 → v3 (v2 is
deprecated and will stop working when GitHub retires the endpoints).
- **Cache `restore-keys` mismatch fixed** — release workflows had primary keys prefixed with `runner.os` (e.g. `Linux`) but `restore-keys` with `matrix.os` (e.g. `ubuntu-latest`). Restore-keys must be a prefix of the primary key to ever match,
so the fallback restore was dead code and `build/jre`/`build/libs` cold-started every run.
- **Triple-overlapping cache collapsed** — `setup-java cache: gradle` + `setup-gradle@v4` + manual `actions/cache` for `build/jre` and `build/libs`. Layers 1 and 2 cover the same paths; layer 3 cached build outputs across runs (unsafe — stale jars could leak between branches — and `build/jre` is no longer produced anyway). Now just `gradle/actions/setup-gradle@v4`, the canonical choice.
- **`cache-overwrite-existing: true` removed from PR workflow** — it forced every PR to overwrite the shared cache, so concurrent PRs were trashing each other's caches. Kept on release workflows since those run from tagged commits on master and should refresh.
- **PR workflow no longer runs `test` twice** — `./gradlew build` already runs `test` via the Java lifecycle; the follow-up step then re-invoked it alongside `itest datatest slowtest`.
- **Release pipeline split into stages** — the compound `clean build copyToOutput test compileSlowtest datatest pfinttest
allReports buildDist prepareRelease pcgenRelease` had four redundant tasks (all in `pcgenRelease`'s dependency graph) and gave the Actions UI no useful timeline. Split into 5 named stages so failures point at the actual broken stage.
- **Release artifacts now attached to the GitHub Release** — the `softprops/action-gh-release` step had `files:` commented out, so releases were created empty. Workflow artifacts were uploaded only for macOS, dropping Linux and Windows outputs on the floor. Now every matrix OS attaches `build/release/*` to the existing
release (idempotent across the matrix because `tag_name` is shared).
- **Workflow names disambiguated** — both `gradle-release.yml` and `gradle-release-manual.yml` declared the same `name: Create Release with Manual Tag`. Renamed the tag-triggered one to `Create Release on Tag Push`.

@karianna this PR should be merged after you merge #7556.
I know, I have polluted this PR with many commits, that's why I'd suggest you to squash the PR as well as do the same for #7556.

I hope, these changes are still bearable, because some of them I took from Claude.


Vest:
I am closing this PR. Use #7559.
Unfortunately, I couldn't foresee merge conflicts :(

PR #7561, opened 2026-05-18
Remove commons-collection4 from the project

Vest:
I have gone through multiple dependencies, and this particular dependency is not heavily used within the project. Using a modern Java JDK and existing API, I was able to eliminate some checks and make the code shorter + as I mentioned, I have removed commons-collection4.

ps. please apply this PR after #7560.


Vest:
you can squash the changes.

PR #7562, opened 2026-05-18
Drop commons-io dependency in favor of JDK equivalents

Vest:
## Summary

Remove `commons-io` as a direct dependency by migrating its remaining call sites to JDK equivalents. Also fix pre-existing Saxon XSLT compilation errors exposed after the switch from Xalan to Saxon-HE.

### commons-io removal
- **`BatchExporter`** — replaced `TeeOutputStream` fan-out with `ByteArrayOutputStream` + sequential `Files.write`; cleaned up `removeTemporaryFiles`, error reporting, and resource handling along the way.
- **`ExportUtilities.getValidFiles`** — `IOFileFilter` / `FileFilterUtils` chain → single `Predicate<File>` + `Stream.filter().sorted()`.
- **`PrintPreviewDialog.SheetLoader`** — `FileUtils.listFiles` + filter chain → `Files.walk` + `Predicate<File>`; dropped the spurious `FilenameFilter` self-implementation.
- **`build.gradle` / `module-info.java`** — dropped the `commons-io` implementation declaration, the `jlink` `forceMerge` entry, and the `requires org.apache.commons.io`.

### XSLT Saxon compatibility fixes
- **`fantasy_common.xsl`** (fantasy, 5e, starfinder, pathfinder_2, 4e, sagaborn) and **`killshot_common.xsl`** — replaced `xsl:use-attribute-sets="equipment.title"` with an equivalent `xsl:call-template name="attrib"` call. Saxon-HE rejects references to undeclared attribute sets at compile time (XTSE0710); the `attrib` template is the correct runtime skinning mechanism.
- **`XsltPdfTemplateCompilationTest`** — new unit test that compiles each fixed stylesheet under Saxon and fails on any XTSE0710/XTSE0720 error, guarding against future regressions.

## Test plan

- [x] `./gradlew compileJava compileTestJava compileItestJava compileSlowtestJava` passes
- [x] PDF character export still produces a valid file
- [x] PDF party export still produces a valid file and surfaces FOP errors
- [x] `XsltPdfTemplateCompilationTest` passes (7/7)
- [ ] Print Preview dialog lists templates and renders a preview

Print Preview is failing because of bugs not related to me.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -220,7 +220,6 @@ dependencies {
 
     implementation fileTree(dir: 'mods/lib', include: ['javafx.*.jar'])
 
-    implementation 'commons-io:commons-io:2.22.0'
     implementation 'org.springframework:spring-web:7.0.7'
     implementation 'org.springframework:spring-beans:7.0.7'
     implementation 'org.springframework:spring-core:7.0.7'
@@ -313,7 +312,7 @@ jlink {
     options = ['--strip-debug', '--no-header-files', '--no-man-pages']
     addExtraDependencies('javafx')
     forceMerge('jep', 'fop', 'Saxon-HE',
-               'commons-lang3', 'commons-io', 'spring',
+               'commons-lang3', 'spring',
                'freemarker', 'jdom2', 'argparse4j', 'xmlunit', 'controlsfx',
                'annotations', 'spotbugs', 'xmlresolver')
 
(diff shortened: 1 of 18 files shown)


Vest:
I will keep it as draft, until I merge it after Java Modules are in master.

karianna (approved):

PR #7563, opened 2026-05-19
Java sub-projects have been added, including PCGen-base and PCGen-formula repositories

Vest:
Convert PCGen-base and PCGen-Formula from external Maven dependencies to in-tree JPMS subprojects, migrate the main code and tests to their new APIs, and fix the CI release pipeline that the migration uncovered.

The two libraries are now built locally as Gradle subprojects with their own module-info.java, included via implementation project(':PCGen-base') / implementation project(':PCGen-Formula'). This unblocks coordinated
changes across the formula stack without round-tripping through external releases.

Changes

JPMS subprojects
- PCGen-base and PCGen-Formula moved into the repo as Gradle subprojects with module declarations.
- Settings, build scripts, and dependency wiring updated; their old settings.gradle files removed.
- Source code relocated under each subproject's code/src/java/...; types whose ownership changed (e.g. pcgen.base.formula.Formula → pcgen.cdom.formula.Formula, pcgen.base.calculation.* → pcgen.cdom.calculation.*) are
renamed in the main module.
- module-info.java for pcgen updated with requires pcgen.base; requires pcgen.formula; and the necessary opens pcgen.cdom.base to pcgen.base; for StagingProxy reflective access.

Main-source API migration
- Main source migrated to the new PCGen-Formula / PCGen-base APIs.
- Test sources updated to compile and run against the new APIs.
- JUnit Platform Launcher added to the subprojects' test dependencies so their suites discover correctly under Gradle 9.5.

Behavioral fixes uncovered by the migration
- SolverManagerFacet: restore processSolver after addModifier/removeModifier; resolve the global scope correctly when looking up modifiers; remove unnecessary processSolver calls after user-input channel writes.
- PlayerCharacter#clone: copy the globalVarScopedMap entry so cloned characters keep their per-scope variables.
- CaseInsensitiveString.equals: revert a String-equality short-circuit branch that produced wrong results in some scopes.

CI release pipeline
- jlink: add addExtraDependencies('javafx') so the merged module (which inherits requires javafx.graphics; requires javafx.controls; from controlsfx via forceMerge) can resolve the JavaFX modules from mods/lib during
merged-module compilation. Without it :createMergedModule failed with error: module not found: javafx.graphics.
- jlink: declare prepareMergedJarsDir.dependsOn copyToLibs to satisfy Gradle 9.5's strict implicit-dependency validation. Both tasks read/write build/libs/, and Gradle now refuses to infer the order.
- Test JVM args: move the JavaFX --module-path mods/lib --add-modules javafx.* block out of allprojects { tasks.withType(Test) { ... } } and into the root project only. PCGen-base and PCGen-Formula are pure formula
libraries, have no mods/lib, and don't depend on JavaFX — applying those args to their tests crashed JVM startup with FindException: Module javafx.web not found.
- forceMerge: drop PCGen-base and PCGen-Formula (they are real modules now, not jars to merge) and commons-collections4 (removed from the project upstream).

Tooling
- Gradle wrapper bumped to 9.5.1.
- Unused imports removed in main-source files flagged by checkstyle after the API migration.


code/src/java/pcgen/base/enumeration/TypeSafeConstant.java
Vest, line 1: Delete TypeSafeConstant from main (identical copy exists in PCGen-base).

When 5660bb56e9 ("Convert PCGen-base and PCGen-Formula to JPMS subprojects") brought PCGen-base in as a proper subproject, the file pcgen.base.enumeration.TypeSafeConstant already existed inside PCGen-base. The code/src/java copy in main was a duplicate of the same fully-qualified class name.

Under JPMS that's a hard error: two modules can't both export the same package, and you can't have the same FQN from two modules on the module path. So one had to go, and main's copy was deleted in favor of PCGen-base's authoritative copy.

This is mentioned in the commit body of 39d02837c4 too ("Relocate split-package classes to resolve module boundary conflicts") — the same fix pattern applied to other split packages.

Vest:
Comment to ca9df42ac47a3d558bd5f85ff92b074f692e8668:

Build warnings audit — all safe to ignore

Three categories of warnings appear in the build output that are not actionable:

createMergedModule: Spring ServiceLoader uses clause
> Cannot derive uses clause from service loader invocation in: org/springframework/beans/factory/serviceloader/AbstractServiceLoaderBasedFactoryBean

The jlink plugin statically analyses bytecode to derive uses clauses for the merged module. AbstractServiceLoaderBasedFactoryBean uses a reflection-based ServiceLoader call that cannot be resolved statically. This Spring class is never instantiated by PCGen, so there is no runtime risk. Silencing it would require adding a speculative uses directive for an unknown service interface.

createDelegatingModules: Terminal digits in module names
> module name component lang3 should avoid terminal digits
> module name component jdom2 should avoid terminal digits

javac style warning about auto-generated module-info.java for non-modular JARs (commons-lang3, jdom2). The module names are derived from the libraries' own Automatic-Module-Name manifest entries — they are upstream
naming choices we cannot change.

jlink: java.base not found in jmods

> java.base module not found in .../jmods, assuming the used Java toolchain has enabled JEP 493

JDK 24+ supports JEP 493 (linkable runtime images), which allows jlink to operate without a separate jmods directory. If Temurin JDK is used, it was built with this feature enabled. This is expected behaviour on JDK 25.

Vest:
@karianna ok, let's wait. I hope, it is the last change that I wanted to have in this PR.

karianna (approved):

PR #7566, opened 2026-05-24
Random-name dialog: rewrite from Swing to JavaFX

Vest:
I have changed the window that generates a random name from Swing to JavaFX:
Random Name

Pulling the engine out of the Swing panel exposed enough rot in the underlying code (never-used variable-substitution machinery, an outgrown JDOM2 dependency, ~33 silently-broken refs in the bundled data) that fixing those became part of the same delivery.

## 1. JavaFX UI

Swing `NameGenPanel` replaced by an FXML-driven JavaFX panel + controller, hosted in a Swing `JDialog`. The engine was extracted into a UI-agnostic `NameGenerator` facade first (~370 lines moved out of the UI class) so the new panel could drive it cleanly.

- Two cascading combos (Category → Title), three gender radios with sticky-fallback semantics, Generate / OK / Cancel.
- Collapsible "Advanced" pane with a Structure override — power users can pin a specific naming pattern instead of the weighted random one.
- Loaded straight into a `Stage` (not via `JFXPanelFromResource`) — the dialog is standalone-modal, not embedded, and the JFXPanel route blew up with `SceneState` NPEs.
- Cmd+Q / save-dialog freeze fix on macOS.
- Tests: pure-logic for the sticky-gender helper + TestFX smoke test that loads the FXML.

Removed: dead `RandomNamePlugin` GMGen wrapper (no callers, never packaged).

## 2. Replace `VariableHashMap` engine with record-based model (`0299fa3`)

The old engine routed every parsed XML element through a `VariableHashMap` keyed by string IDs, with `Rule extends ArrayList<String>` storing child references re-resolved on every name generation. The indirection existed to support `<VARMOD>` / `<ROLL>` / `<IF>` / `<WHILE>` — features the DTD itself calls "unsupported" and that PCGen never used.

- `NameList`, `Rule`, `RuleSet`, `NameGenData` are now records; `RulePart` is a sealed interface with `ListRef`, `RuleSetRef`, and a `Literal` enum.
- Two-pass linker resolves cross-refs at load time via shared maps. No string lookups at generation time.
- `NameGenData.unresolvedReferences()` collects dangling refs so dataset bugs surface as data instead of being silently dropped.
- 9 files deleted (~1,000 lines), 2 added.

## 3. Fix broken `GETLIST`/`GETRULE` references in bundled data (`cdbe72c`)

The new loader's clean-data assertion failed on first run, exposing 33 unresolved id refs the legacy engine had been swallowing:

- 25 wrong-tag `GETLIST` refs targeting a `<RULESET>` → switched to `GETRULE`.
- 4 wrong-tag `GETRULE` refs targeting a `<LIST>` → switched to `GETLIST`.
- 2 typo IDs: `tamrelilic-noun-a` (×7) → `tamrelic-noun-a`; `tamriel-dunmer-surname-Suffix` → `...-suffix`.
- 6 `mar-standard-rule` refs split into `-male-rule` / `-female-rule` per call-site gender.
- 3 `russian-historical-male-patrynomic-rule_alt` refs — the `_alt` variant was never authored — pointed at the existing rule.

## 4. Drop JDOM2

Loader rewritten on `javax.xml.parsers`. One real divergence: mixed-content `<VALUE>foo<SUBVALUE>...</SUBVALUE></VALUE>` elements need a `directText` helper because `getTextContent` concatenates descendant text where JDOM2's `getText` returned only direct children. Regression test pins the "Donn / brown, brown-haired" entry. Dependency, module require, about-dialog attribution strings, and a long-broken `Src.iml` jar reference all gone.

Engine package renamed `pcgen.core.doomsdaybook` → `pcgen.core.namegen` first — "Doomsday Book" was a 2003 artifact of the third-party engine PCGen originally adopted.

## 5. Lazy XML loading

The full-DOM parse of the 4 MB `Random Names` directory was running just to populate the dialog's category list. Replaced with two tiers:

- Cheap StAX **pre-scan** at startup builds a category/ruleset/list index (attribute headers only, no `<RULE>` / `<VALUE>` bodies).
- Full DOM parse runs only on the file behind the user's selection, plus files it transitively references via `GETLIST`/`GETRULE`.

`NameGenerator`'s public API is unchanged. **Cold dialog-open: ~150ms → ~40ms** on the bundled dataset, measured via JMH (benchmark removed before merging once the numbers were in).

## 6. Tests + parser hardening

- Per-ruleset smoke test: a `@TestFactory` calls `generate()` on every `final`-usage ruleset and asserts non-empty output. Catches empty lists, all-zero weights, content-free rules. +144 tests / ~0.6s to `:test`.
- Slowtest task discovery fix: `datatest` / `inttest` / per-game variants were silently running zero tests — include globs were rooted at `slowtest/` instead of `build/classes/java/slowtest/`, and `useJUnitPlatform()` was missing. Per-game variants now share a JVM, cutting `pfinttest` wall-clock ~50s → ~30s.
- DTD `weight="1"` default: a non-validating parser was producing `""` and crashing the load with `NumberFormatException`. Helper treats blank as 1.
- XXE hardening on the DOM factory (cheap, the data is trusted but the principle is right).
- `childElements` walks the `NodeList` once instead of twice.

## 7. Build cleanup

`Src.iml` (predates Gradle, references nonexistent jars), the Gradle `idea` plugin (superseded by direct Gradle import), and an unused `lobobrowser/cobra-0.98.4` artifact all dropped.


build.gradle
```diff
@@ -835,32 +830,6 @@ tasks.named("dependencyUpdates") {
     }
 }
 
-// Support for creating IntelliJ IDEA files.
-idea {
```
Vest, line 839: With better Gradle script we don't need Idea as a separate plugin anymore.

Vest:
XMLs were corrected by the bot. UI, too.

Vest:
@karianna I thought, I reduce the code base much, but I wanted to have it properly implemented, it hasn't reduced much :(
Now, we have a JavaFX window and tests

PR #7568, opened 2026-05-31
jlink 4.0.2 + native installers + namegen cleanup

Vest:
## Summary

Three independent threads bundled into one PR — they were uncovered while debugging the first one. **OK to squash** on merge: the commit history was useful while iterating but the end-state changes are small and cohesive.

### 1. `jlink` plugin upgrade and merged-module fix
- Bump `org.beryx.jlink` 4.0.1 → 4.0.2.
- 4.0.1's `createMergedModule` failed with duplicate-`requires` compile errors. Root cause: asm 9.10 (pulled in by jlink 4.0.1) parses Java 25 class files that 9.9.1 silently skipped, so the plugin's auto-derived `requires` set grew enough to collide with our explicit list under `additive = true`.
- Fix: keep only the modules the scanner can't infer on its own (`java.naming`, `java.scripting`, `java.management`, `jdk.httpserver`, `jdk.unsupported.desktop`).

### 2. Release pipeline now ships real native installers + portable zips
- **Windows was silently excluded from `fullJpackage`.** A leftover `Os.FAMILY_MAC || Os.FAMILY_UNIX` predicate in `release.gradle` dated from when Windows shipped via NSIS (`buildNsis`); NSIS was retired in `be48763530` but the jpackage hookup was never added on Windows. Removed the predicate.
- **`fullJpackage` only built the runtime image, never the installer.** It depended on `assembleJpackageImage` (or `patchMacJpackage` on Mac), not the `jpackage` task that produces `.dmg`/`.deb`/`.exe`. Releases shipped only `image-<os>.zip` from `jlinkZip` — a JVM runtime without `/data`, `/plugins`, `/preview`, `/outputsheets`, so launching it gave the "missing required folders" dialog. Made `jpackage` depend on the data-bundling step (and on Mac, on `patchMacJpackage` so the patched plist + `MacDirLauncher` go into the installer); pointed `fullJpackage` at `jpackage`.
- Pin one `installerType` per OS (`dmg` / `deb` / `exe`) — without it jpackage emits both formats per platform, doubling build time and (on Linux) requiring `rpmbuild` that isn't installed on the Ubuntu runner.
- Added `--win-shortcut --win-menu` so the `.exe` creates Start-menu and desktop entries.
- **New `portableZip` task** produces `pcgen-<version>-<os>-<arch>-portable.zip` alongside the installer — same self-contained app, but unzip-and-run for users who don't want an installer (USB sticks, locked-down environments, etc.). Wired into `assembleArtifacts` so CI uploads it to the GitHub Release.
- **Fixed Linux `.deb` not being uploaded.** `jpackage` names Debian packages `pcgen_<version>_<arch>.deb` (underscore), which the existing `pcgen-*.*` glob (hyphen) missed. Added `pcgen_*.deb` to the include set. `.rpm` (`pcgen-<version>.<arch>.rpm`) was already covered.
- **Removed `image-<os>.zip` from the release uploads.** Output of `jlinkZip` — a JVM runtime image without `/data`, `/plugins`, `/preview`, or `/outputsheets`. Misleadingly named: users who tried to run it got the "missing required folders" dialog. Autobuild keeps it via `buildDist` (separate distribution channel, unchanged).
- **Removed the `checksum` task and `SHA256-digests-*.txt` files.** GitHub Releases compute and expose a SHA-256 for every uploaded asset automatically; anyone verifying integrity can `shasum -a 256` against the value GitHub provides. The 4 generated digest files were noise.

### 3. `pcgen.core.namegen` perf / cleanup (no behavior change)
- `NameList` / `RuleSet`: O(n) two-pass weighted `pick()` → O(log n) `binarySearch` over a precomputed cumulative-weight array.
- `DataValue`: hand-rolled linked list (`DataSubValue`) → lazy `LinkedHashMap`.
- `NameGenLazyData.gendersForRuleset`: O(categories) bucket scan → O(1) lookup via `RuleSetMeta.categoryTitles()`.
- `NameGenerator.assemble`: lazy meaning / pronunciation `StringBuilder`s; only allocate when a part actually overrides.

## Final release-asset list (per platform)

After this PR, each release produces a focused, user-meaningful asset list:

| Platform | Installer | Portable |
|---|---|---|
| macOS (aarch64) | `pcgen-<v>.dmg` | `pcgen-<v>-mac-aarch64-portable.zip` |
| Linux (x64) | `pcgen_<v>_amd64.deb` | `pcgen-<v>-linux-x64-portable.zip` |
| Linux (aarch64) | `pcgen_<v>_arm64.deb` | `pcgen-<v>-linux-aarch64-portable.zip` |
| Windows (x64) | `pcgen-<v>.exe` | `pcgen-<v>-windows-x64-portable.zip` |

Plus `pcgen-<v>-sources.jar`. 9 user-meaningful assets, down from 15 mixed-purpose ones. Per-asset SHA-256 hashes are available on the GitHub Release UI/API.

## Test plan

- [x] `./gradlew createMergedModule` succeeds (was the original failure).
- [x] `./gradlew fullJpackage` on Mac produces `pcgen-6.09.08.dmg` (160 MB) with all four required folders inside `PcGen.app/Contents/app/`.
- [x] `./gradlew portableZip` on Mac produces `pcgen-6.09.08.RC1-mac-aarch64-portable.zip` (148 MB) with the same self-contained `PcGen.app`.
- [x] Manual release run on `Vest/pcgen` (run `26779067448`, tag `v6.09.08.RC1`): macOS, Ubuntu x64, Ubuntu ARM, and Windows jobs all green; produced `.dmg` + `.exe` + portable zips on all OSes. (Note: that run predates the `.deb` glob fix and the `image-*.zip` / digest-file removal — a fresh run is recommended before merge to confirm the final asset list.)
- [x] `./gradlew :test --tests 'pcgen.core.namegen.*' --tests 'pcgen.gui3.namegen.*'` (173 tests, 0 failures).
- [x] Sanity check: open Random Name dialog, generate a few names.


Vest:
I am optimizing a CI build again - I want to build artifacts for different platforms. When i am ready I will create a new pr

PR #7575, opened 2026-06-02
ci: collapse 5 sequential gradle invocations into 1

Vest:
## Summary

Stacks on top of #7568 — please review and merge that PR first.

Each `./gradlew` invocation pays ~5–10s of configuration overhead and ~2–3s of daemon startup, plus a fresh task-graph build. The release workflows ran 5 of them in series:

| Step | Notes |
|---|---|
| `./gradlew clean build` | `clean` does nothing on a fresh CI runner — `build/` doesn't exist yet |
| `./gradlew compileSlowtest datatest pfinttest` | data/integration tests |
| `./gradlew allReports` | `= checkstyleMain + pmdMain + spotbugsMain`, **all already executed by `build`** (verified in the previous run's log: `> Task :spotbugsMain` ran during `build`, then again as `UP-TO-DATE` here) |
| `./gradlew buildDist` | already a transitive dep of `pcgenRelease` |
| `./gradlew pcgenRelease` | umbrella |

Collapse to a single invocation, drop the redundant `clean` and `allReports`, and drop `buildDist`. Also set `fail-fast: false` on the matrix so a transient runner preemption (seen on the partner-supplied `ubuntu-24.04-arm` image — see #7568 verification run) doesn't kill the other three OS jobs.

Add `--stacktrace` to every `gradlew` call across all three workflows (`gradle-test.yml`, `gradle-release.yml`, `gradle-release-manual.yml`). Costs nothing on green builds (only fires on failure) and saves a diagnostic round-trip when CI fails for a non-obvious reason — recent example: a TestFX/JavaFX 25 incompatibility surfaced as `Test process … finished with non-zero exit value 1` with no further detail, requiring a separate diagnostic commit + workflow re-run to learn it was actually an `InaccessibleObjectException` from `--add-exports` vs `--add-opens`. With `--stacktrace` enabled by default, that would have been visible in the first failed log.

## Measured impact

Identical input (same `fix_packaging` HEAD with PR #7568 applied), back-to-back runs on `Vest/pcgen` with warm caches:

| OS | Baseline (5 invocations) | This PR (1 invocation) | Saving | % faster |
|---|---:|---:|---:|---:|
| ubuntu-24.04-arm | 10m50s | 9m56s | -54s | 8% |
| ubuntu-latest | 13m21s | 11m40s | -1m41s | 13% |
| macOS | 16m04s | 13m16s | -2m48s | 17% |
| windows | 25m04s | 18m33s | **-6m31s** | **26%** |
| **Wall-clock** | **~25 min** | **~19 min** | **-6 min** | **24%** |

Windows benefited the most because each `gradlew` startup is dominated by JVM warmup + slower disk I/O on that runner image.

## Test plan

- [x] All 4 OS matrix entries green on the optimized run (Vest/pcgen run `26779067448`).
- [x] Same 15 release artifacts produced as the baseline run — installers (`.dmg`, `.exe`), portable zips, jlinkZip images, sources jar, SHA256 manifests.
- [x] Failure semantics preserved: a failed task still aborts the build (no `--continue` flag).
- [x] `--stacktrace` doesn't change passing-build behavior — Gradle only emits the stack when a task throws.


Vest:
@karianna , please merge #7568 first, you can squash it! The same can be done here. 
after this commit, you can release artifacts.

PR #7580, opened 2026-06-03
TestHelper: load plugins via JUnit extension; cleanup

Vest:
## Summary

Replace the manual `TestHelper.loadPlugins()` boilerplate with a proper JUnit 5 lifecycle hook. Originally tried a static initializer (early commits on this branch); that turned out to load plugins at JVM class-load time, **before** test setup — which broke `:datatest` (settings-file race) and `:test` (TestFX tearDown trying deep reflection without `--add-opens`). Rewrote using a `BeforeAllCallback` extension instead. Also keeps the orthogonal cleanups from the original PR: dead-method removal and `parsePCClassText` deduplication.

The PR also finishes the long-pending JUnit 4 → 6 migration (last three commits) — the `getOrComputeIfAbsent` deprecation warning that surfaced while building this branch is what kicked it off.

**OK to squash**, but the commit history is split into logical groups if reviewers prefer to keep them.

### 1. JUnit 5 extension (`PCGenTestEnvironment`)

- New `code/src/testcommon/pcgen/test/PCGenTestEnvironment.java` — a `BeforeAllCallback` that runs `Main.createLoadPluginTask().run()` exactly once per JVM, keyed via `ExtensionContext.Store.GLOBAL`.
- **Opt-in** via `@ExtendWith(PCGenTestEnvironment.class)`, NOT auto-discovered. Auto-discovery would force plugin loading on tests like `SetSolverManagerTest.setUp` that explicitly assume an empty `PluginFunctionLibrary` (its `addFunction(getOther)` collides with the entry that plugin loading puts in the singleton). The original `loadPlugins()` was opt-in too — each test that wanted plugins called it explicitly. The annotation preserves that boundary while removing the boilerplate.
- Applied to `AbstractCharacterTestCase` and `AbstractJunit5CharacterTestCase`. The annotation propagates to ~114 subclasses automatically.
- Applied directly to the 10 standalone test classes that previously called `TestHelper.loadPlugins()`.
- `TestHelper.loadPlugins()` and the static initializer that briefly replaced it are both deleted. No call sites remain.

### 2. `Main.runBootstrapTasks()`

The 4-step bootstrap sequence (`createLoadPluginTask` → `GameModeFileLoader` → `CampaignFileLoader`, run via `PCGenTaskExecutor`) was duplicated verbatim in `startupWithGUI()`, `startupWithoutGUI()`, **and** open-coded twice more in `DataTest.loadGameModes()` / `DataLoadTest.loadGameModes()`. Pulled into one `Main.runBootstrapTasks(PCGenTaskListener…)` helper. All four call sites now use it. Eliminates the risk of one of them drifting out of order — the bug we just hit when the static-initializer attempt loaded plugins before `Main.loadProperties()`.

### 3. TestFX/JavaFX 25 fix

Separately, `:test` was crashing with `InaccessibleObjectException: module javafx.graphics does not "opens" com.sun.javafx.application` from TestFX's `cleanupParameters` (which uses `setAccessible(true)` on a private static field). The existing `--add-exports` flag in `build.gradle:684` wasn't enough — `setAccessible` needs `--add-opens`. Changed `--add-exports` → `--add-opens` for `com.sun.javafx.application`. The other three `--add-exports` flags in the JVM-arg block stay as-is (they're for Monocle and JavaFX logging, which only need export-level access).

### 4. `--stacktrace` on every CI gradle invocation

Costs nothing on green builds (only fires on failure). This session needed two separate diagnostic commits — one for the TestFX `InaccessibleObjectException`, one for the static-initializer-induced JVM exit — purely because Gradle's default output for a crashed test JVM is `non-zero exit value 1` with no further detail, plus a hint to use `--stacktrace`. Make that the next step happen automatically.

### 5. Original cleanup work (kept from earlier in the PR)

These were done before the static-initializer attempt and are unrelated to it. Worth keeping:

- **Dead-method removal in `TestHelper`**: `makeAbilityFromString`, `makeWeaponProf`, `loadGameModes(String)` — zero callers anywhere in the tree, last real users removed years ago (2014, 2009, 2018 respectively). Recoverable from git via `git log -S` if anyone needs them.
- **`parsePCClassText` dedup**: `PCClassTest` and `AddClassSkillsTest` carried near-identical 17-line copies. Both now route through the canonical `TestHelper.parsePCClassText`. -62/+5 lines net.
- **IDE-flagged warnings**: 4 unused imports, an unthrown `throws`, raw-type → parameterized `Class<?>`, string-based class-walk replaced by identity compare, redundant local inlined, parameterized log calls, missing Javadoc filled in.

### 6. JUnit 4 → 6 migration (final three commits)

The BOM was already on `org.junit:junit-bom:6.1.0`, but several pieces of test scaffolding still referenced JUnit 3/4 APIs. The build comment claiming "~870 legacy tests" was stale by years — actual count of pure-JUnit-4 files at the start of this work was 1, plus 86 hybrid files using `org.junit.Assert.*` static imports.

- **Deprecated APIs in test scaffolding** (`aca6bc17a5`): `ExtensionContext.Store#getOrComputeIfAbsent` was deprecated in JUnit 6.0 in favor of `#computeIfAbsent` (mirrors `java.util.Map`) — `PCGenTestEnvironment` swapped over. `AbstractFormulaTestCase` and `TestUtilities` dropped `junit.framework.TestCase.fail` for Jupiter `Assertions.fail`. `RecursiveFileFinderTest` swapped `org.junit.Test` for the Jupiter `@Test`.

- **Bulk `Assert` → `Assertions` rewrite** (`efeb5e11a5`): 86 files, 1054 message-arg position swaps. JUnit 4's `assertEquals(message, expected, actual)` becomes Jupiter's `assertEquals(expected, actual, message)`. The trap is that mass-rewriting the imports alone would compile cleanly because Jupiter's `assertEquals(Object, Object, String)` overload silently matches the old shape with `message` interpreted as `expected` and vice versa — every failure message would be wrong, but the test class would still pass when assertions hold. The script (paren-balanced parser, top-level comma split) only swaps when the first arg is a string literal and the arity matches the JUnit-4 leading-message shape (3+ for `assertEquals`/`assertSame`/`assertNotSame`/`assertArrayEquals`, 2 for the rest). Nine cases where the leading message was a `+`-concatenation expression rather than a literal were caught by the compiler (the `assertTrue(BooleanSupplier, ...)` overload doesn't accept a String first arg) and fixed by hand in `PreRaceTest` and `EquipmentSetFacadeImplTest`.

- **Drop the dead deps** (`ac6034694c`): `junit:junit:4.13.2` and `junit-vintage-engine` come out of both `build.gradle` files. Stale comment removed.

## Test plan

- [x] `./gradlew :test :itest :datatest :slowtest` — all four green locally before the JUnit 6 commits (`:slowtest` ~11 min, others a couple minutes total).
- [x] Reviewed the test-result XMLs under `build/test-results/{test,itest,datatest,slowtest}/` — no failures, no errors, no skipped tests beyond the existing baseline.
- [x] After the JUnit 6 commits: `compileTestJava`, `compileSlowtestJava`, `compileItestJava` all clean; `:test`, `:itest`, and a representative `:slowtest` set (`StatListTest`, `PrereqHandlerTest`, `PObjectTest`, `PreRaceTest`) green.
- [ ] Full `:slowtest` and `:datatest` re-run after the JUnit 6 commits — the migration is mechanical but worth confirming end-to-end.
- [ ] CI green on this PR's `Build PCGen with Gradle` workflow.

### DIFF
--- a/.github/workflows/gradle-release-manual.yml
+++ b/.github/workflows/gradle-release-manual.yml
@@ -181,7 +181,7 @@ jobs:
             jdks-
 
       - name: Build, test, and assemble release artifacts
-        run: ./gradlew build compileSlowtest datatest pfinttest pcgenRelease
+        run: ./gradlew build compileSlowtest datatest pfinttest pcgenRelease --stacktrace
 
       - name: Upload release artifacts as workflow artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/gradle-release.yml
+++ b/.github/workflows/gradle-release.yml
@@ -151,7 +151,7 @@ jobs:
             jdks-
 
       - name: Build, test, and assemble release artifacts
-        run: ./gradlew build compileSlowtest datatest pfinttest pcgenRelease
+        run: ./gradlew build compileSlowtest datatest pfinttest pcgenRelease --stacktrace
 
       - name: Upload release artifacts as workflow artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/gradle-test.yml
+++ b/.github/workflows/gradle-test.yml
@@ -32,7 +32,7 @@ jobs:
         run: ./gradlew build
 
       - name: Run integration, data, and slow tests
-        run: ./gradlew itest datatest slowtest --stacktrace --info
+        run: ./gradlew itest datatest slowtest
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
--- a/.github/workflows/gradle-test.yml
+++ b/.github/workflows/gradle-test.yml
@@ -29,10 +29,10 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
 
       - name: Build with Gradle Wrapper
-        run: ./gradlew build --stacktrace --info
+        run: ./gradlew build
 
       - name: Run integration, data, and slow tests
-        run: ./gradlew itest datatest slowtest --stacktrace --info
+        run: ./gradlew itest datatest slowtest
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
--- a/.github/workflows/gradle-test.yml
+++ b/.github/workflows/gradle-test.yml
@@ -32,7 +32,7 @@ jobs:
         run: ./gradlew build
 
       - name: Run integration, data, and slow tests
-        run: ./gradlew itest datatest slowtest
+        run: ./gradlew itest datatest slowtest --stacktrace --info
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
--- a/.github/workflows/gradle-test.yml
+++ b/.github/workflows/gradle-test.yml
@@ -29,10 +29,10 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
 
       - name: Build with Gradle Wrapper
-        run: ./gradlew build
+        run: ./gradlew build --stacktrace
 
       - name: Run integration, data, and slow tests
-        run: ./gradlew itest datatest slowtest
+        run: ./gradlew itest datatest slowtest --stacktrace
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
@@ -50,7 +50,7 @@ jobs:
             PCGen-Formula/build/test-results/**/*.json
 
       - name: Run Coverage
-        run: ./gradlew testCoverage
+        run: ./gradlew testCoverage --stacktrace
 
       - name: Upload Report to artifacts
         uses: actions/upload-artifact@v4
--- a/.github/workflows/gradle-test.yml
+++ b/.github/workflows/gradle-test.yml
@@ -29,10 +29,10 @@ jobs:
         uses: gradle/actions/setup-gradle@v4
 
       - name: Build with Gradle Wrapper
-        run: ./gradlew build
+        run: ./gradlew build --stacktrace --info
 
       - name: Run integration, data, and slow tests
-        run: ./gradlew itest datatest slowtest
+        run: ./gradlew itest datatest slowtest --stacktrace --info
 
       - name: Publish Test Results
         uses: EnricoMi/publish-unit-test-result-action@v2
--- a/PCGen-Formula/build.gradle
+++ b/PCGen-Formula/build.gradle
@@ -40,9 +40,7 @@ dependencies {
     testImplementation platform('org.junit:junit-bom:6.1.0')
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
     testImplementation 'org.junit.jupiter:junit-jupiter-params'
-    testImplementation 'junit:junit:4.13.2'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
-    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
     testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
 }
 
--- a/PCGen-Formula/code/src/test/pcgen/base/testsupport/AbstractFormulaTestCase.java
+++ b/PCGen-Formula/code/src/test/pcgen/base/testsupport/AbstractFormulaTestCase.java
@@ -18,6 +18,7 @@
 package pcgen.base.testsupport;
 
 import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.fail;
 
 import java.util.Arrays;
 import java.util.List;
@@ -27,7 +28,6 @@ import java.util.Optional;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 
-import junit.framework.TestCase;
 import pcgen.base.formatmanager.FormatManagerLibrary;
 import pcgen.base.formatmanager.FormatUtilities;
 import pcgen.base.formatmanager.OptionalFormatFactory;
@@ -130,7 +130,7 @@ public abstract class AbstractFormulaTestCase
 		boolean isStat = ((Boolean) staticVisitor.visit(node, null)).booleanValue();
 		if (isStat != b)
 		{
-			TestCase.fail("Expected Static (" + b + ") Formula: " + formula);
+			fail("Expected Static (" + b + ") Formula: " + formula);
 		}
 	}
 
@@ -186,7 +186,7 @@ public abstract class AbstractFormulaTestCase
 		{
 			result = Arrays.asList((Object[]) result);
 		}
-		TestCase.fail(
+		fail(
 			"Expected " + valueSimpleName + " (" + valueOf + ") for Formula: "
 				+ formula + ", was " + result + " (" + resultSimpleName + ")");
 	}
--- a/PCGen-Formula/code/src/test/pcgen/base/testsupport/TestUtilities.java
+++ b/PCGen-Formula/code/src/test/pcgen/base/testsupport/TestUtilities.java
@@ -25,7 +25,8 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 
-import junit.framework.TestCase;
+import static org.junit.jupiter.api.Assertions.fail;
+
 import pcgen.base.format.ArrayFormatManager;
 import pcgen.base.format.BooleanManager;
 import pcgen.base.format.NumberManager;
@@ -84,8 +85,7 @@ public final class TestUtilities
 		}
 		catch (ParseException e)
 		{
-			TestCase
-				.fail("Encountered Unexpected Exception: " + e.getMessage());
+			fail("Encountered Unexpected Exception: " + e.getMessage());
 			return null;
 		}
 	}
--- a/build.gradle
+++ b/build.gradle
@@ -240,9 +240,6 @@ dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api'
     testImplementation 'org.junit.jupiter:junit-jupiter-params'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
-    // JUnit 4 is still used by ~870 legacy tests; pulled in via vintage engine + junit jar.
-    testImplementation 'junit:junit:4.13.2'
-    testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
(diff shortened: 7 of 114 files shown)


Vest:
Diagnosed the failure — it isn't actually caused by this PR's static-initializer change. The `:datatest` step's `non-zero exit value 1` with zero reported test failures is a downstream symptom; the real crash is in `:test` (which runs as part of `./gradlew build`):

```
java.lang.reflect.InaccessibleObjectException: Unable to make field
private static java.util.Map com.sun.javafx.application.ParametersImpl.params accessible:
module javafx.graphics does not "opens com.sun.javafx.application" to unnamed module @609e8838

  at org.testfx.toolkit.impl.ToolkitServiceImpl.cleanupParameters(ToolkitServiceImpl.java:177)
  at org.testfx.toolkit.impl.ToolkitServiceImpl.cleanupApplication(ToolkitServiceImpl.java:138)
  at org.testfx.api.FxToolkit.cleanupAfterTest(FxToolkit.java:178)
  at org.testfx.framework.junit5.ApplicationExtension.afterEach(ApplicationExtension.java:87)
```

[Full stack in run 26866284491.](https://github.com/PCGen/pcgen/actions/runs/26866284491/job/79230589029) The `:datatest` JVM that runs after this inherits the broken classpath/JVM-args setup and dies at startup, hence the orphan-`java`-pid messages and zero JUnit XMLs.

### What's wrong

`build.gradle` line 684 had `--add-exports` for `javafx.graphics/com.sun.javafx.application` to work around [TestFX/TestFX#638](https://github.com/TestFX/TestFX/issues/638). But TestFX's `ApplicationExtension.afterEach` → `ToolkitServiceImpl.cleanupParameters` does `setAccessible(true)` on a **private static field** — that's deep reflection, requires `--add-opens`, not just `--add-exports`. `--add-exports` only exposes public types to compilation. JavaFX 25 GA tightened up the difference; older JavaFX versions silently accepted the weaker flag.

### Why this PR triggered a latent bug

The pre-existing `--add-exports` was load-bearing only for tests that touched JavaFX *during* their `@AfterEach`. Moving plugin loading into `TestHelper.<clinit>` means a much broader set of tests now bootstraps JavaFX via `Main.createLoadPluginTask`, which lands in TestFX's lifecycle, which then trips the cleanup hook.

### Fix

Pushed [`d487ffc6d1`](https://github.com/Vest/pcgen/commit/d487ffc6d1) which:
- Changes the single `--add-exports` for `com.sun.javafx.application` → `--add-opens` (the rest of the JVM-arg block is correct as-is — `com.sun.javafx.util` and `com.sun.javafx.logging` only need `--add-exports`).
- Reverts the temporary `--stacktrace --info` flags I added to `gradle-test.yml` for diagnosis (they were a one-shot, not a permanent setting — they balloon the log to ~100k lines).

Should be green now. Worth a re-run.

Vest:
The TestFX `--add-opens` fix worked — the earlier `InaccessibleObjectException` is gone, `:test` and `:itest` now pass. But `:datatest` is still crashing at JVM startup (~4s after fork, before any test code runs).

Looking at the diff again, this is almost certainly the original concern about the static initializer:

- `:datatest`'s test classes (`DataTest`, `DataLoadTest`) both `import pcgen.util.TestHelper`. JUnit's class loader touches `TestHelper` during test discovery → triggers `TestHelper.<clinit>` → `Main.createLoadPluginTask().run()`.
- Unlike `:test` (which sets up TestFX/Monocle/`prism.order=sw`/etc. in its `jvmArgs`), `:datatest` runs with bare-bones JVM args. So `Main.createLoadPluginTask()` is running in a JVM environment where its expectations about working directory, `ConfigurationSettings`, plugin path layout, etc. may not hold.
- In the OLD code, `loadPlugins()` was called from `@BeforeEach` setup hooks AFTER tests had configured their environment. Moving it into `<clinit>` means it runs at a much earlier (and less prepared) lifecycle point.

Just pushed [`c4f9d849`](https://github.com/Vest/pcgen/commit/c4f9d849) which adds `--stacktrace --info` to the `:datatest` invocation so the actual exception will show in the next failed log. That'll confirm whether it's an `ExceptionInInitializerError` from `Main.createLoadPluginTask`, or something else entirely. Will revert the diagnostic flag once we have the cause.

Vest:
The diagnostic flags exposed the cause. From the `:datatest` log:

\`\`\`
DataLoadTest > testLoadSources STANDARD_ERROR
    INFO: Using PCC Location of 'data'.
    SEVERE Test worker Main:263 No settingsDir specified via -s in batch mode and no default exists.
\`\`\`

Then the JVM exits — that SEVERE message is from `Main.java:263` calling `GracefulExit.exit(1)`. No exception, no stack trace; the JVM just dies.

### Why

Look at the proper init order inside `Main.startupWithGUI()`:

\`\`\`java
loadProperties(true);                              // line 185 — load properties FIRST
PCGenTaskExecutor executor = new PCGenTaskExecutor();
executor.addPCGenTask(createLoadPluginTask());     // line 192 — THEN load plugins
\`\`\`

`DataLoadTest.loadGameModes()` follows the same order:

\`\`\`java
String pccLoc = TestHelper.findDataFolder();                          // [1]
TestHelper.createDummySettingsFile(TEST_CONFIG_FILE, ...);            // [2]
configFactory.registerAndLoadPropertyContext(...);                    // [3] sets SETTINGS_FILES_PATH
Main.loadProperties(false);                                           // [4]
PCGenTask loadPluginTask = Main.createLoadPluginTask();               // [5]
loadPluginTask.run();
\`\`\`

In the OLD code, step [1] just returned the data folder path — `loadPlugins()` was a separate method that the test never called itself (Main.loadProperties handled it). Order was preserved.

In the NEW code, step [1] **triggers TestHelper.\<clinit\>** which immediately runs `Main.createLoadPluginTask().run()` — that's plugin loading happening BEFORE step [4]'s loadProperties. By the time step [4] runs, something in the plugin-loading path has perturbed system properties / `ConfigurationSettings` state enough that loadProperties no longer sees `SETTINGS_FILES_PATH`, and Main exits the JVM.

### Recommendation

Revert the static-initializer approach. The stated goal — "plugins load once automatically without each test having to call `loadPlugins()`" — is fine in principle, but `\<clinit\>` is the wrong mechanism: it runs whenever any classloader touches `TestHelper.class`, including at JUnit discovery time, which is well before any test's `@BeforeAll`/setup has run. Better alternatives:

1. **Keep the original `loaded` guard.** Old code was three lines and worked. The "I keep forgetting to call loadPlugins" annoyance is real but doesn't justify breaking the init order.
2. **JUnit 5 extension.** Write a `BeforeAllCallback` registered via `@ExtendWith(LoadPluginsExtension.class)` on test classes (or globally via `@AutoEnable`). That way plugin loading happens AFTER the JUnit lifecycle has set up its state and run user `@BeforeAll`/`@BeforeEach`, but before the test method.
3. **Move the call inside `findDataFolder()`/`createSource()`.** They're the methods that actually depend on plugins being loaded. Idempotent guard inside the method body keeps it lazy and ordered correctly.

Option 1 is the smallest change and gets the build green immediately. Options 2 and 3 are nicer if you want to address the original ergonomic concern.

Want me to push option 1 as a "revert + targeted improvements" commit on this branch, or do you want to take it from here?

Vest:
Rewrote the static-initializer approach as five logical commits ([572dde5..d02509c](https://github.com/Vest/pcgen/compare/3fa6ce3ea8..d02509c61f)). All four test source sets (\`:test\`, \`:itest\`, \`:datatest\`, \`:slowtest\`) pass locally.

### Summary

| # | Commit | Change |
|---|---|---|
| 1 | [572dde5](https://github.com/Vest/pcgen/commit/572dde58f6) | \`Main.runBootstrapTasks()\` extracted from \`startupWithGUI\`/\`startupWithoutGUI\`. One canonical definition of the loader sequence; production behaviour-preserving. |
| 2 | [05f9bde](https://github.com/Vest/pcgen/commit/05f9bde0fa) | New \`PCGenTestEnvironment\` JUnit 5 \`BeforeAllCallback\` in \`code/src/testcommon/pcgen/test/\`. Loads plugins once per JVM. |
| 3 | [08c8d56](https://github.com/Vest/pcgen/commit/08c8d5666c) | Static initializer removed from \`TestHelper\`; \`loadPlugins()\` kept as a deprecated no-op. |
| 4 | [73a4d7b](https://github.com/Vest/pcgen/commit/73a4d7b720) | \`DataTest\`/\`DataLoadTest\`'s open-coded loader dance replaced with \`Main.runBootstrapTasks()\`. |
| 5 | [d02509c](https://github.com/Vest/pcgen/commit/d02509c61f) | Plugin loading is **opt-in** via \`@ExtendWith(PCGenTestEnvironment.class)\`, not auto-discovered. |

### Why opt-in instead of auto-discovery

I initially wired auto-discovery via \`META-INF/services\` so every test class would get plugins for free. That broke \`SetSolverManagerTest\`:

\`\`\`
java.lang.IllegalArgumentException: Cannot load two functions of name: getOther
  at pcgen.base.formula.inst.SimpleFunctionLibrary.addFunction
  at pcgen.rules.context.VariableContext.addFunction
  at pcgen.base.solver.SetSolverManagerTest.setUp(SetSolverManagerTest.java:78)
\`\`\`

\`VariableContext\`'s constructor copies all functions from the global \`PluginFunctionLibrary\` singleton (\`code/src/java/pcgen/rules/context/VariableContext.java:147\`). \`SetSolverManagerTest\` builds a fresh \`VariableContext\` and adds \`GetOtherFunction\` itself — assuming the singleton was empty. The test was relying on plugins NOT being loaded. Auto-discovery breaks that assumption.

The OLD code already had this property: \`loadPlugins()\` was opt-in (each test that wanted plugins called it from \`@BeforeEach\`/\`@BeforeAll\`). Tests that didn't call it ran without plugins, and \`SetSolverManagerTest\` was one of them. The \`@ExtendWith\` annotation preserves that boundary while removing the boilerplate.

### Where the annotation is applied

- \`AbstractCharacterTestCase\` and \`AbstractJunit5CharacterTestCase\` — the two abstract bases that previously called \`loadPlugins()\` from their setup. The annotation propagates to ~114 subclasses automatically.
- 10 standalone test classes that previously called \`loadPlugins()\` directly.
- All call sites of \`TestHelper.loadPlugins()\` are gone (the method is kept as a deprecated no-op for any external callers).

### Verification

\`\`\`
$ ./gradlew :test :itest :datatest :slowtest
BUILD SUCCESSFUL
\`\`\`

(\`:slowtest\` took 11 min; the other three a couple minutes total.)

PR #7583, opened 2026-06-04
build.gradle: fix clean/download race and merged-module missing JDK requires

Vest:
Two independent build.gradle fixes that together make `gradle clean fullJpackage` produce a runnable app on a fresh checkout.

## 1. Order download tasks after clean to avoid a parallel-execution race

Running `gradle clean fullJpackage` (or any task graph that combines `clean` with the download tasks) could fail with two races, because the download tasks had no ordering relative to clean and Gradle was free to interleave them:

* `:clean` aborted with `Unable to delete directory 'build': New files were found` because `:downloadJdk` / `:downloadJavaFXLocal` were already streaming `.part` files into `build/download-task/` while `:clean` was deleting `build/`.
* `:downloadJavaFXLocal` aborted moving its `.part` into `mods/` because `:cleanMods` had just wiped `mods/`.

Failures are timing-sensitive (CPU/disk speed, worker availability), which is why some machines hit them and others don't.

Add `mustRunAfter clean` to `:downloadJdk`, `:downloadJfxMods`, and `:downloadJavaFXLocal` — same pattern already used by `:copyToLibs` and `:build` in this file.

`mustRunAfter` (rather than `dependsOn`) is correct: it only constrains order when both tasks are already in the graph, so plain `gradle fullJpackage` still works incrementally without forcing an unwanted clean.

## 2. Merged module needs the JDK requires that jdeps misses on Temurin (JEP 493)

After the build succeeds, launching the produced app crashes at startup on Temurin:

```
Exception in thread "main" java.lang.IllegalAccessError: class freemarker.log._JULLoggerFactory
(in module net.sourceforge.pcgen.merged.module) cannot access class java.util.logging.Logger
(in module java.logging) because module net.sourceforge.pcgen.merged.module does not read module java.logging
    at freemarker.log._JULLoggerFactory.getLogger(...)
    at freemarker.template.Configuration.<clinit>(...)
    at pcgen.io.ExportUtilities.getObjectWrapper(...)
    at pcgen.system.PCGenPropBundle.<clinit>(...)
    at pcgen.system.Main.<clinit>(...)
Failed to launch JVM
```

#7581 attempted to fix this by switching to `useJdeps = 'exclusively'` on the assumption that the JDK's bundled `jdeps` would reliably resolve JDK packages → modules. That fix is incomplete on JDKs that don't ship JDK jmods on disk.

### Why Temurin 25 specifically breaks here

JEP 493 ("Linking Run-Time Images without JMODs") allows a JDK distribution to omit the `.jmod` files normally found under `$JDK/jmods/`, and Temurin 25 takes that option for the macOS/aarch64 builds we download via `:downloadJdk` (`api.adoptium.net/v3/binary/latest/25/ga/mac/aarch64/jdk/hotspot/normal/eclipse`). After `:extractJdk`, `jdks/jdk_mac_aarch64/Contents/Home/jmods/` is empty; the JDK can still link new images via `lib/modules` (jimage), which is what jpackage/jlink uses internally.

`org.beryx.jlink` v4.0.2's `JdepsExecutor` invokes jdeps as:

```
$javaHome/bin/jdeps -v --generate-module-info <out> \
    --module-path $javaHome/jmods:$jlinkJarsDir <merged.jar>
```

Crucially it passes only `$javaHome/jmods` — not `lib/modules` — as the JDK module source. On Temurin 25 that directory is then populated *only* with the JavaFX jmods that `:downloadJfxMods` extracts into it. jdeps therefore sees:

* javafx.base, javafx.controls, javafx.graphics, javafx.media — present as jmods, so it emits `requires javafx.*`,
* java.base, java.logging, java.desktop, java.sql, … — **not present anywhere on the supplied module-path**, so it cannot resolve them and silently omits the corresponding `requires` directives from the generated merged `module-info.java`.

The merged module then loads at runtime without those reads, and the first reflective call into `java.util.logging` (from freemarker's `_JULLoggerFactory`) fails with the IllegalAccessError above before `main()` even runs.

This is why the build worked for some maintainers and not for others:

| JDK distribution | `$JDK/jmods/` ships JDK jmods? | jdeps resolves JDK requires? | Packaged app launches? |
|---|---|---|---|
| SapMachine 25 | yes (full set) | yes | yes |
| Temurin 25 (macOS aarch64) | no (JEP 493) | no — silently omits them | no — IllegalAccessError on `java.util.logging.Logger` |

Verified on this machine: running the JDK's `jdeps -s` directly against `build/jlinkbase/tmpmerged/pcgen.merged.module-null.jar` produces the full expected list (`java.compiler`, `java.desktop`, `java.logging`, `java.prefs`, `java.rmi`, `java.sql`, `java.xml`, `jdk.jfr`, `jdk.unsupported`, `jdk.xml.dom`), but the `module-info.class` the plugin compiles into the merged module contains *only* the javafx ones — confirming the missing-requires happens inside the plugin's jdeps invocation, not in jdeps itself.

### Fix

Add the missing JDK modules to the `mergedModule { … }` block. The plugin's `additive = true` mode appends them on top of jdeps' output, which on the JEP 493 JDKs we ship does not contain them — so no duplicate-requires errors at javac time. On a hypothetical JDK with full jmods, jdeps would derive the same set and we'd get duplicates; this is acceptable because `:extractJdk` always pulls the configured Temurin distribution (no JDK jmods) into `jdks/jdk_<os>_<arch>` and `$javaHome` for the jpackage flow points there. SapMachine et al. are not on the build's jpackage code path.

If a future change points `:extractJdk` at a JDK that *does* ship jmods, the manual `requires` list will need to drop those entries (or the build will need to start invoking jdeps against `lib/modules` directly and skip the plugin's jdeps step).

### Companion change: pin `jlink.javaHome` to the downloaded Temurin

The `org.beryx.jlink` plugin defaults `javaHome` to whatever Gradle's toolchain auto-discovery resolves (SapMachine on a SDKMAN machine, Temurin in CI, etc.), so the merged-module behavior described above could differ between developer machines: SapMachine ships full JDK jmods and jdeps emits a complete `requires` list; a local Temurin run hits the JEP 493 blind spot above.

Set `jlink.javaHome` to the same path `targetPlatform` uses (the Temurin extracted by `:extractJdk`) so every plugin task — `prepareMergedJarsDir`, `createMergedModule`, `createDelegatingModules`, `prepareModulesDir`, `jlink`, `jpackageImage`, `jpackage`, `suggestMergedModuleInfo` — invokes `jdeps`/`javac`/`jlink`/`jpackage` against the same `$javaHome` on every host. Each of those tasks now `dependsOn extractJdk` so the Temurin is on disk before the plugin reads it. `compileJava`/`test`/`run` and IDE debug still use whatever the toolchain resolves.

## 3. Drive-by build-script cleanups

While in the file, several stale or noisy bits got cleaned up. Each is independent of the fixes above and behavior-preserving:

- **Stale `clean` TODO + unused `binDir`** — the TODO predated `cleanOutput`; `binDir = "code/bin"` was referenced nowhere.
- **Unused `ext.classpath` + dead Class-Path attributes** — removed; jars now declare only the entries they actually need.
- **`com.github.ben-manes.versions` plugin** — dropped along with the unused `dependencyUpdates` task.
- **Unused `ivy` (`pc-gen.org/librepo/`) and JBoss repositories** — no dependencies resolve from them.
- **Redundant comments** around `mavenCentral` and a stray blank line in `dependencies`.
- **`Closure.memoize()` → explicit cache map** in `latestJavaVersion` for clearer caching semantics.
- **`segments[1..-1]` → `Arrays.copyOfRange(..., String[].class)`** — Groovy range slicing was triggering IDE type-inference noise.
- **Closure parameter types** added to `createJarTask` in `plugins.gradle` and `eachFile` in `build.gradle` to clear "cannot infer type" warnings.
- **`code/sbin/fromdos.exe`** — leftover Cygwin CRLF→LF utility from a pre-Gradle build flow; nothing references it. Companion to commit `12fd0ecf36` (2017) which removed the rest of `code/sbin/`.
- **`xdocs/` + `installers/release-notes/genreleasenotes.pl`** — Maven-era documentation directory and the Perl script that consumed `xdocs/changes.xml`. The script hasn't been functionally edited since 2017 and the data files in `xdocs/` were last touched between 2006 and 2017; pre-built release-notes HTML in `installers/release-notes/` is authored without it.
- **Top-level `apex-ruleset.xml` and `ruleset.xml`** — 26-byte stubs at the repo root containing only the literal text `code/standards/ruleset.xml`. PMD points directly at `code/standards/ruleset.xml` via `code/gradle/reporting.gradle` (and the equivalents in PCGen-base/PCGen-Formula); the stubs are referenced nowhere.
- **Unused locals in `plugins.gradle`** — `manifestFile = "code/manifest"` (untracked leftover JAR template, never read) and `srcJavaDir = "code/src/java"` (defined but never referenced).

## Test plan

- [x] `gradle clean fullJpackage` completes without `:clean FAILED` / `:downloadJavaFXLocal FAILED` (verified on macOS aarch64, Temurin 25.0.3).
- [x] `gradle fullJpackage` (no clean) still runs incrementally and does not trigger `:clean`.
- [x] `gradle clean` alone still works.
- [x] Generated `build/jlinkbase/jlinkjars/pcgen.merged.module-null.jar`'s `module-info.class` lists all 12 expected JDK requires plus the 3 jdeps-derived JavaFX requires; no duplicates.
- [x] The packaged app launches without the freemarker `IllegalAccessError` — verified `build/jpackage/PcGen.app/Contents/MacOS/PcGen` reaches the JavaFX UI init phase on Temurin 25.0.3 macOS aarch64.
- [x] With SapMachine as the toolchain JDK, `gradle clean fullJpackage` still uses Temurin for the jlink/jpackage path (`:jlink` logs the JEP 493 message against `jdks/jdk_mac_aarch64/Contents/Home/jmods`) and the packaged app launches.


Vest:
@karianna please push the button :)

Vest:
### Why we keep `useJdeps = 'exclusively'` instead of reverting to the default `'no'`

Now that we list every JDK `requires` manually, a fair question is whether we still need `useJdeps = 'exclusively'` at all — couldn't we drop back to the plugin's default `'no'` and let our explicit list cover everything?

Short answer: no. The default `'no'` path is strictly worse on Temurin, for a different reason than jdeps' blind spot.

#### How `useJdeps = 'no'` resolves packages → modules

The plugin's default path doesn't call `jdeps`. Instead it uses its own resolver in `SuggestedMergedModuleInfoBuilder` / `ModuleManager`:

1. `PackageUseScanner` walks every class in the merged jar and collects the set of external packages it references.
2. `ModuleManager.getExportsMap(...)` builds a *package → module* map by scanning files under the supplied module paths — specifically `$javaHome/jmods` plus the project's modular dependency jars — and reading each file's `ModuleDescriptor.exports()` (only for things that look like a `.jar` or `.jmod` file). Packages that don't appear in any scanned descriptor are not in the map.
3. For each external package, the resolver looks it up in that map and emits `requires <module>;`. Packages it cannot find produce the famous \"Cannot find module exporting …\" log lines — and silently no `requires` directive.

That's it. There is **no built-in fallback** for \"this looks like a JDK package — assume `java.base` / `java.logging` / `java.desktop` / etc.\" If a JDK module doesn't show up as a `.jmod` file on disk under `$javaHome/jmods/`, its packages are treated as unknown.

#### Why that breaks on Temurin 25

JEP 493 (\"Linking Run-Time Images without JMODs\") lets a JDK distribution omit the per-module `.jmod` files. Temurin 25 takes that option for the macOS/aarch64 build we download via `:downloadJdk`. After `:extractJdk`, `jdks/jdk_mac_aarch64/Contents/Home/jmods/` exists but is empty. `:downloadJfxMods` then drops the JavaFX `.jmod` files into that same directory.

So when the `'no'` resolver scans `$javaHome/jmods` it sees only:

```
javafx.base.jmod, javafx.controls.jmod, javafx.fxml.jmod, javafx.graphics.jmod,
javafx.media.jmod, javafx.swing.jmod, javafx.web.jmod, jdk.jsobject.jmod,
jfx.incubator.input.jmod, jfx.incubator.richtext.jmod
```

And nothing else. Its package → module map ends up populated for `javafx.*` (and the project's third-party `.jar` dependencies that have a `module-info`), but **every JDK package — `java.lang`, `java.io`, `java.util`, `java.util.logging`, `java.awt`, `java.sql`, …— is missing from the map**. That is why the original `useJdeps = 'no'` build produced log lines like:

> Cannot find module exporting java.logging
> Cannot find module exporting java.io
> Cannot find module exporting java.lang
> …

and emitted a merged `module-info` that omitted every JDK requires. Switching to `'exclusively'` was meant to bypass that resolver — but jdeps reads modules from the same `$javaHome/jmods` directory, so it inherits the same blind spot for the JDK modules. That's why we still need the manual list.

#### Why `'exclusively'` is still better than `'no'` even with our manual list

If both modes miss JDK requires, why prefer one over the other? Because they don't behave identically for **non-JDK** packages — and `'no'` is worse there:

| Aspect | `useJdeps = 'no'` (badass resolver) | `useJdeps = 'exclusively'` (jdeps) |
|---|---|---|
| Resolves JDK requires when `$javaHome/jmods/` lacks JDK jmods | ✗ misses (no fallback) | ✗ misses (same root cause) |
| Resolves JavaFX requires when jmods present | ✗ historically required `addExtraDependencies('javafx')` and various `forceMerge` workarounds because the resolver mis-attributed JavaFX packages | ✓ jdeps reads JavaFX jmods directly and emits `requires javafx.base / javafx.controls / javafx.graphics / javafx.media` correctly |
| Reading a properly-formed `module-info.class` from each input | Re-implemented via `PackageUseScanner` + `ModuleManager.getExportsMap` (a custom Java program the plugin compiles and runs) | Native `jdeps`, which is the JDK's reference tool for exactly this question |
| Determinism | Depends on the resolver's heuristics and which jars are visible | Depends only on the contents of the supplied module-path |
| Plugin's stated direction | Legacy fallback retained for compatibility | The mode added in plugin 4.x specifically because it's more reliable on modern JDKs |

Concretely for our build, switching back to `'no'` would mean:

1. We'd still need every manual `requires 'java.logging'` etc. we just added (because the JDK blind spot is the same).
2. We'd additionally need `requires 'javafx.base'`, `'javafx.controls'`, `'javafx.graphics'`, `'javafx.media'` — because the badass resolver's track record with JavaFX in this project is what motivated `addExtraDependencies('javafx')` in the first place. Right now jdeps is finding those four for us automatically and emitting them with no help.
3. We'd revert to a custom heuristic for everything, where jdeps gives us the JDK's own answer for the things it can see.

Strictly worse, no upside. So `'exclusively'` stays — explicit requires fill in the JEP 493 blind spot, and jdeps handles JavaFX and the modular third-party jars as designed.

Vest:
With this commit we no longer update our dependencies ourselves and rely on GitHub Actions Dependabot to propose dependency upgrades.

PR #7596, opened 2026-06-15
Add nightly build workflow and overhaul manual release workflow

karianna:
## Summary

Adds two GitHub Actions workflows for automated releases.

### 1. `gradle-nightly.yml` (new)

Daily scheduled workflow that produces full release artifacts for all four supported platforms whenever `master` has received new commits in the previous 24 hours.

- **Trigger:** cron `0 2 * * *` (02:00 UTC daily) + `workflow_dispatch` for ad-hoc runs.
- **Change detection:** `git rev-list --count --since='24 hours ago' HEAD` on `master`. Zero commits → short-circuits and skips the build entirely.
- **Versioning:** `<base>-NIGHTLY.<YYYYMMDD>` where `<base>` is the `gradle.properties` version with any `-SNAPSHOT` suffix stripped (e.g. `6.09.08-NIGHTLY.20260615`).
- **Publishing:** GitHub **pre-release** tagged `v<nightly_version>` with all 4 platform artifacts attached.
- **Retention:** keeps the seven most recent nightly pre-releases; deletes older releases **and** their underlying git tags via `gh release delete --cleanup-tag`.

### 2. `gradle-release-manual.yml` (replaces existing)

Owns the full release lifecycle from version bump to next-dev bump.

- **Trigger:** `workflow_dispatch` only.
- **Inputs:**
  - `release_version` — e.g. `6.09.08` or `6.09.08.RC1`. Validated against `^\d+\.\d+\.\d+(\.[A-Za-z0-9]+)?$`. No `v` prefix, no `-SNAPSHOT` suffix.
  - `prerelease` — boolean (default `false`), useful for RCs.
- **Jobs:**
  1. `prepare_release` — validates the version, verifies the tag is free, commits the `gradle.properties` bump to the dispatched branch as `github-actions[bot]`, tags the commit, and creates the GitHub Release.
  2. `build_release` — 4-platform matrix mirroring `gradle-release.yml`; checks out the new tag and runs `./gradlew build compileSlowtest datatest pfinttest pcgenRelease --stacktrace`, uploading artifacts to the Release.
  3. `bump_to_next_dev` — only runs when every matrix job succeeded. Invokes `./gradlew --no-daemon --no-configuration-cache updateVersionToNext` (reuses the existing Groovy helper in `code/gradle/releaseUtils.groovy`) and commits the new `-SNAPSHOT` version back to the dispatched branch so the source tree is ready for further development.

### Untouched

`gradle-release.yml` (tag-push path), `gradle-test.yml`, `codeql-analysis.yml`, and all Gradle build logic.

## Validation

- `actionlint 1.7.12` + `shellcheck 0.11.0` integration — 0 findings on all 5 workflows.
- All YAML files parse cleanly.
- `updateVersionToNext` confirmed reachable from the top-level build (`build.gradle:853` applies `code/gradle/release.gradle`).
- Cannot smoke-test the workflows end-to-end without a real run; first CI run will exercise them.

## Caveats

- **Version bump rule:** `updateVersion` in `releaseUtils.groovy` bumps the trailing numeric token via regex `(\d+)([^\d]*$)`. So `6.09.08` → `6.09.09-SNAPSHOT` (patch bump), but `6.09.08.RC1` → `6.09.08.RC02-SNAPSHOT` (RC bump, not patch bump). Operators promoting a final release should pass the final version (e.g. `6.09.08`), not an RC tag. The workflow input help text documents this.
- **Branch protection:** if `master` requires PR review, direct pushes from `github-actions[bot]` (in both `prepare_release` and `bump_to_next_dev`) will fail. We can switch the bump step to open a PR instead if the first real run reveals this.
- **jpackage `appVersion`** already strips `-SNAPSHOT` and non-numeric suffixes (`build.gradle:349`), so the nightly suffix doesn't break installer naming — no installer-side changes needed.
- **Same-day collision:** manually dispatching the nightly workflow twice on the same UTC day will fail the second run at "tag already exists". Acceptable given the schedule fires once per day.

## Test plan

Once merged:
1. Trigger `gradle-nightly.yml` via `workflow_dispatch` to validate the happy path without waiting for cron.
2. Trigger `gradle-release-manual.yml` with an RC version (e.g. the next `.RC` candidate, with `prerelease: true`) to validate the full pipeline end-to-end.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.github/workflows/gradle-nightly.yml
+++ b/.github/workflows/gradle-nightly.yml
@@ -1,0 +1,271 @@
+name: Nightly Build
+
+# Runs daily at 02:00 UTC. Skips the build entirely if there have been
+# zero new commits on master in the last 24 hours (so no PRs merged and no
+# direct commits = no nightly release). When changes are present, builds
+# full release artifacts for all four supported platforms and publishes
+# them as a GitHub pre-release tagged v<base>-NIGHTLY.<YYYYMMDD>.
+#
+# The base version comes from gradle.properties with any existing
+# -SNAPSHOT suffix stripped, e.g. 6.09.08-SNAPSHOT -> base 6.09.08 ->
+# nightly version 6.09.08-NIGHTLY.20260615.
+#
+# The workflow keeps the seven most recent nightly pre-releases; older
+# nightly releases and their underlying git tags are deleted automatically
+# at the end of every successful run.
+
+on:
+  schedule:
+    # Daily at 02:00 UTC. Approx 14:00 NZDT / 21:00 ET / 18:00 PT.
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+concurrency:
+  # Never run two nightly builds in parallel; never cancel one in flight,
+  # so we don't end up with a half-published release.
+  group: nightly
+  cancel-in-progress: false
+
+jobs:
+  check_changes:
+    name: Check for changes
+    runs-on: ubuntu-latest
+    outputs:
+      should_build: ${{ steps.detect.outputs.should_build }}
+      nightly_version: ${{ steps.detect.outputs.nightly_version }}
+      tag_name: ${{ steps.detect.outputs.tag_name }}
+      commit_sha: ${{ steps.detect.outputs.commit_sha }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: master
+
+      - name: Detect changes and compute nightly version
+        id: detect
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -euo pipefail
+
+          NEW_COMMITS=$(git rev-list --count --since='24 hours ago' HEAD)
+          COMMIT_SHA=$(git rev-parse HEAD)
+          SHORT_SHA=$(git rev-parse --short HEAD)
+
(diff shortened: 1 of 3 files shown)


Vest:
@karianna — small inconsistency between the workflow's inline comment and AGENTS.md / the PR body, found while reviewing the doc updates. Could you (or Copilot) double-check?

In `.github/workflows/gradle-release-manual.yml` the header comment says:

> For 6.09.08.RC1 it yields **6.09.08.RC2-SNAPSHOT**.

But `code/gradle/releaseUtils.groovy::updateVersion` builds the next version with:

```groovy
m.replaceAll("${ String.format('%02d', (m[0][1] as int) + 1) }${ m[0][2] }")
```

`%02d` always zero-pads to two digits, so `RC1` actually becomes **`RC02-SNAPSHOT`**, not `RC2-SNAPSHOT`. The PR body and the new AGENTS.md text both correctly say `RC02` / "zero-padded to two digits" — only the workflow's inline comment is out of sync.

Two-character bugfix: change `RC2-SNAPSHOT` → `RC02-SNAPSHOT` in the YAML header comment. Everything else looks consistent.

(Side note for Copilot: please verify by tracing `String.format('%02d', ...)` in `releaseUtils.groovy` and confirming the workflow comment is the only inaccurate spot.)

karianna:
@Vest going for the two digit implementation, updated accordingly

PR #7607, opened 2026-06-17
ci: split build and integration-tests into parallel jobs

Vest:
> **Stacked on top of #7603.** PR #7603 fixes the cross-fork race in PcgenFtlTestCase. Without it, master's slowtest is flaky under any cores/2 parallelism, including this PR's own CI runs. Once #7603 merges, rebasing this branch onto master leaves only the workflow-yaml diff.

## Summary

- Split the single sequential CI job into two jobs that run in parallel: `build` (compile + plugin jars + spotbugs + unit `:test` + coverage) and `integration-tests` (`itest` + `datatest` + `slowtest`).
- Both share the Gradle dependency cache via `setup-gradle@v4`. `slowtest` does not depend on `:test` or `:spotbugs*` (verified with `./gradlew slowtest --dry-run`), so the integration-tests job re-pays only ~30 s of `code` + `slowtest` compile, against ~4 min saved on the critical path.
- Each job runs its own `publish-unit-test-result-action` with a disjoint file glob, so the two checks show only their own results instead of one job blocking on the other's reports.

## Why

Today the wall clock is `build (~4 min) → itest+datatest+slowtest (~16 min) ≈ 20 min`. Static analysis on master sources has nothing to do with whether the slow character-export tests pass; running them back-to-back in one job just sums the time. Expected wall clock after this split: `max(~5 min, ~16 min) ≈ 16 min` today, dropping to `max(~5 min, ~7-8 min) ≈ 8 min` once #7603 lands and slowtest can use cores/2 parallelism without racing.

## Test plan
- [ ] CI run on this PR shows two jobs (`build`, `integration-tests`) running in parallel.
- [ ] Both jobs succeed.
- [ ] Two separate test-result checks appear on the PR (`Unit Test Results`, `Integration Test Results`).
- [ ] Coverage comment is posted from the `build` job.
- [ ] Wall-clock is meaningfully shorter than the previous sequential run for the same commit.

### DIFF
--- a/.github/workflows/gradle-test.yml
+++ b/.github/workflows/gradle-test.yml
@@ -1,5 +1,9 @@
-# This workflow will build a Java project with Gradle and cache/restore any dependencies to improve the workflow execution time
-# For more information see: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-gradle
+# This workflow builds PCGen with Gradle. The `build` and `integration-tests`
+# jobs run in parallel — they share the Gradle build cache via setup-gradle@v4
+# but otherwise compile their inputs independently. Splitting the previous
+# single-job sequence (`./gradlew build` then `./gradlew itest datatest slowtest`)
+# shaves the ~4 min compile/spotbugs/unit-test phase off the critical path of
+# the ~16 min slow-test phase.
 
 name: Build PCGen with Gradle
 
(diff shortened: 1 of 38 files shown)


Vest:
After #7603 is merged, I want to test here, whether the CI is getting faster.

Vest:
I have also cleaned some warnings, made tests more isolated - they can be run in parallel, removed dead code. I hope, I made it better :)

PR #7608, opened 2026-06-17
Fix macOS portable .app being unlaunchable after unzip

Vest:
## Summary

The macOS nightly artifact (`pcgen-*-mac-*-portable.zip`) ships a `.app` bundle that cannot be launched after extraction. Double-clicking it gives the generic "The application \"PcGen.app\" can't be opened" Finder dialog.

## Root cause

`build.gradle:466` uses Gradle's built-in `Zip` task to pack the bundle. Plain ZIP strips POSIX file modes from archive entries, so on extraction:

- `PcGen.app/Contents/MacOS/MacDirLauncher` (the bash shim that `cd`s into `Contents/app` and execs the real binary) loses its `0755` mode.
- `PcGen.app/Contents/MacOS/PcGen` (the Mach-O launched by jpackage) loses it too.

macOS refuses to launch a bundle whose `CFBundleExecutable` isn't executable, hence the dialog. Verified locally with the latest nightly: both files arrived as `-rw-r--r--` with `Feb 1 1980` timestamps (the ZIP epoch default).

## Fix

Use `ditto -c -k --keepParent` for the macOS branch instead of `Zip`. `ditto` is Apple's canonical archiver for `.app` bundles — it preserves Unix modes, symlinks, extended attributes, and code-signature metadata that plain ZIP discards. Linux/Windows continue to use Gradle's `Zip` task; their executables don't depend on a Unix exec bit.

## Verification

Locally on macOS arm64, after the fix:

```
$ unzip -Z build/jpackage/pcgen-*-mac-*-portable.zip 'PcGen.app/Contents/MacOS/*'
-rwxr-xr-x  2.1 unx   223936 bX defN 26-Jun-17 22:44 PcGen.app/Contents/MacOS/PcGen
-rwxr-xr-x  2.1 unx       57 bX defN 26-Jun-17 22:45 PcGen.app/Contents/MacOS/MacDirLauncher
```

- Exec bits preserved (`-rwxr-xr-x`)
- Real timestamps, not 1980
- `_CodeSignature/` block intact (matters once builds start being signed)

Extracting with `ditto -x -k` round-trips cleanly and the bundle launches.

## Notes

- Gatekeeper will still reject unsigned nightlies (`source=no usable signature`); that's a separate Apple Developer ID notarisation issue, not packaging. Users currently work around it with `xattr -dr com.apple.quarantine`.
- `Files.deleteIfExists` is used in `doFirst` rather than `project.delete` because the latter trips the Gradle configuration cache (`Invocation of 'delete' references a Gradle script object from a Groovy closure at execution time`).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,8 @@
 
 import groovy.json.JsonSlurper
 
+import java.nio.file.Files
+
 import org.apache.tools.ant.filters.FixCrLfFilter
 import org.apache.tools.ant.taskdefs.condition.Os
 
@@ -384,10 +386,10 @@ tasks.named("jpackageImage") {
     def jpackageDir = layout.buildDirectory.dir("jpackage")
     doFirst {
         def root = jpackageDir.get().asFile.toPath()
-        if (java.nio.file.Files.exists(root)) {
-            java.nio.file.Files.walk(root).withCloseable { stream ->
+        if (Files.exists(root)) {
+            Files.walk(root).withCloseable { stream ->
                 stream.filter { it.fileName.toString() == ".DS_Store" }
-                      .forEach { java.nio.file.Files.deleteIfExists(it) }
+                      .forEach { Files.deleteIfExists(it) }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


Vest:
@karianna I have noticed that your nightly build was broken for Mac: https://github.com/PCGen/pcgen/releases/tag/v6.09.08.RC1-NIGHTLY.20260617
Probably, it is my fault, but I don't know for sure - my bot detected, when we ZIP .app files, we lose 0500 flag for shell script, and if we want to zip it properly, we must use ditto (for mac).
I have vibecoded a bit, can you give it a try, merge it and check the nightly build later?

Vest:
@karianna — would value your opinion on the packaging approach here before this gets merged. There are at least three reasonable ways to fix the broken-bit problem, and I'm not certain `ditto` is the right one long-term:

| Approach | Pros | Cons |
|---|---|---|
| **`ditto -c -k --keepParent`** *(this PR)* | Reads modes off disk — no hard-coded executable list. Preserves symlinks, xattrs, and code signatures, so it's future-proof for notarisation/frameworks. Apple-canonical. | Needs an `Exec` task and a `mac`/non-mac branch in `build.gradle`. macOS-only, so it can't be tested on Linux/Windows runners. |
| **Gradle `Zip` + `filePermissions { unix(0755) }` via `eachFile`** | Smallest diff, no new tooling, matches the rest of `build.gradle`. Stays cross-platform. | Hard-codes `MacDirLauncher` and `PcGen` as the only files needing the exec bit — silently breaks if jpackage emits another helper binary. Plain ZIP also drops xattrs and `_CodeSignature` integrity, which matters once builds are signed. |
| **Gradle `Tar` (gzip)** | Tar natively preserves modes and symlinks; one Gradle-native task; no platform branch. | `.tar.gz` isn't the macOS distribution convention — Linux/Windows nightly artifacts are `.zip`, and Mac users expect a double-clickable `.zip`/`.dmg`. Also drops xattrs without the BSD-tar-only `--xattrs` flag. |

For PCGen *today* (unsigned builds, no embedded frameworks, no symlinks in the jpackage output), the Gradle-`Zip` option would arguably be the smallest and simplest fix. `ditto` is what you'd want once signing/notarisation lands — but that may or may not be on your roadmap.

Happy to switch the PR to whichever approach you prefer. Any thoughts?

karianna:
The ditto approach is probably the best as we'll want to eventually notarize.

PR #7611, opened 2026-06-18
Pin javaVersion + javafxVersion in gradle.properties

Vest:
## Summary

- Replace the configuration-time Adoptium API lookup (`latestJavaVersion` closure in `build.gradle`) with two static properties in `gradle.properties`: `javaVersion=25` and `javafxVersion=25.0.3`.
- Drops the `JsonSlurper` import and ~20 lines of caching indirection. The four call sites (`downloadJfxMods` × 2, `downloadJavaFXLocal` × 2) now interpolate the property directly.
- Documents the bump procedure in `README.md` ("Bumping Java / JavaFX versions") covering patch vs. major bumps, what to verify, and a `curl` reference command for querying the latest JDK patch from Adoptium.

Addresses the "preferred" option from the TODO.md item *"Remove configuration-time HTTP call to Adoptium"*. The `5e55cc7321` per-invocation cache reduced the network call to one per Gradle run, but the call still happened during task **configuration**, so configuration-cache compatibility was incomplete and offline builds still touched the network on a cold cache.

## Test plan

- [x] `./gradlew help` configures cleanly, configuration cache stores without warnings.
- [x] `./gradlew downloadJfxMods downloadJavaFXLocal` resolves URLs to `25.0.3` from the property (no Adoptium version lookup).
- [x] `./gradlew compileJava compileTestJava` succeeds.
- [x] CI (full build + tests) green.


Vest:
I have decided to move the Java and JavaFX versions to the property file, as Gradle recommends. This will help us to have the JDK upgrades in a central place, then it reduces some runtime noise from API calls.
I'd stay with this approach for a while.
@karianna

karianna:
@copilot resolve the merge conflicts in this pull request

PR #7618, opened 2026-06-20
Fix Domains tab showing object/prereq debug text instead of domain names (#6978)

### LINKED ISSUE #6978: [BUG]
**Describe the bug**
in the domains tab the section "Domain" doesn't display the domain name but some object fotmatting tags.

**To Reproduce**
start pcgen, load SRD 3.5 (for example)
1. Go to 'Domain' tab
2. See on the "Domain" box the object formatting tags.

**Expected behavior**
in the Domain box the names of the domains should appears

**Screenshots**
immagine

**Desktop (please complete the following information):**
 - OS: windows 11 22h2
 - pcgen Version 6.09.05

Vest:
## Summary

- Fixes #6978: the Domains tab listed entries like `Object:Air, Prereq:[<prereq operator="GTEQ" operand="1" ...>]` instead of the domain name.
- Root cause: `QualifiedObject.toString()` was a leftover auto-generated debug stub (`"Object:" + theObject + ", Prereq:" + getPrerequisiteList()`). `DomainInfoTab.DomainRenderer` is a `DefaultTableCellRenderer`, which falls back to `value.toString()` for the cell text, so the debug string was rendered directly. The same path also broke `SearchFilterPanel`, which matches search input against `element.toString()`.
- Fix: delegate `QualifiedObject.toString()` to the wrapped object (returning `""` when the wrapped object is `null`). Domain inherits `CDOMObject.toString()` which returns the display name, so the cell now shows "Air", "Animal", … and search starts matching domain names again.
- No callers in `code/src/java` or `code/src/test` rely on the previous debug-format output (verified by search), so this is a safe behavior change.

## Test plan

- [x] New unit test `pcgen.core.QualifiedObjectTest` covers delegation and the null-wrapped case (`./gradlew :test --tests pcgen.core.QualifiedObjectTest` passes).
- [x] Manual: launch PCGen, load *SRD 3.5 for Players*, hit **New**, open the **Domains** tab, and confirm the right-hand list shows plain domain names (Air, Animal, Artifice, …) and that the **Search** field filters by name.

### DIFF
--- a/code/src/java/pcgen/core/QualifiedObject.java
+++ b/code/src/java/pcgen/core/QualifiedObject.java
@@ -108,11 +108,7 @@ public class QualifiedObject<T> extends ConcretePrereqObject implements Qualifyi
 	@Override
 	public String toString()
 	{
-		// TODO Auto-generated method stub
-		return "Object:"
-				+ theObject.toString()
-				+ ", Prereq:"
-				+ getPrerequisiteList().toString();
+		return theObject == null ? "" : theObject.toString();
 	}
 
 	@Override
--- a/code/src/test/pcgen/core/QualifiedObjectTest.java
+++ b/code/src/test/pcgen/core/QualifiedObjectTest.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2026 (C) Vest <Vest@users.noreply.github.com>
+ *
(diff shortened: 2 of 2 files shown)


Vest:
I have fixed the bug #6978 
fix-image

PR #7624, opened 2026-06-22
UserChooseInformation: make immutable, drop setTitle

Vest:
Pass the title via constructor; drop the post-construction setter and the null-default branch in `getTitle()`. The sole caller (`UserInputToken.parseToken`) computes the title up-front and now constructs once.

Makes the class `final` and the `title` field `final`. Eliminates the EI_EXPOSE_REP2 hazard on this `Chooser<String>` implementation: all eight `Chooser<T>` implementations in the codebase are now state-free or constructor-injected only.

## What this looks like

Before:

```java
public class UserChooseInformation ... {
    private String title = null;

    public String getTitle() {
        return title == null ? "Provide User Input" : title;
    }

    public void setTitle(String chooseTitle) {
        title = chooseTitle;
    }
}
```

After:

```java
public final class UserChooseInformation ... {
    private static final String DEFAULT_TITLE = "Provide User Input";
    private final String title;

    public UserChooseInformation() { this(DEFAULT_TITLE); }

    public UserChooseInformation(String title) {
        this.title = Objects.requireNonNullElse(title, DEFAULT_TITLE);
    }

    public String getTitle() { return title; }
}
```

`UserInputToken.parseToken` collapses two `setTitle(...)` branches into a single up-front `title` local and one `new UserChooseInformation(title)` call.

## Verification

- `./gradlew compileJava` clean.
- `:test --tests "plugin.lsttokens.choose.*" --tests "pcgen.cdom.base.*"` — 2149 tests, 0 failures.

## Scope note

Standalone refactor — independent of the other two SpotBugs PRs (#TBD EI suppression, #TBD IMPROPER_UNICODE rewrite).

### DIFF
--- a/code/src/java/pcgen/cdom/base/UserChooseInformation.java
+++ b/code/src/java/pcgen/cdom/base/UserChooseInformation.java
@@ -20,6 +20,7 @@ package pcgen.cdom.base;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import pcgen.cdom.enumeration.AssociationListKey;
 import pcgen.cdom.enumeration.GroupingState;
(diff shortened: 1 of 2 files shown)


Vest:
I think, this check #7625 is very stupid. I couldn't find a better word. It checks, whether our classes are immutable, if not - it suggests us to call a .clone of the object.
I have "fixed" the mentioned class in this PR, but there are over 100 classes that must be repaired, but my bot tells me, that it isn't worth it.
It is better/safer to disable this stupid check.

As alternative solutions, I could bring an `@Immutable` annotation (a new dependency), or clone, or remove all setters (also stupid).
This is a minor cosmetic fix.

karianna:
I'll merge this but I think a future upgrade will be to move most of our data classes to Records/Value Types.

Vest:
If we survive until that very moment, then you are right.

PR #7627, opened 2026-06-23
CDOM: mark 20 leaf classes final to clear CT_CONSTRUCTOR_THROW

Vest:
SpotBugs flags constructors that throw exceptions when the class is not final, since a malicious subclass could observe a partially-constructed instance via finalize(). All 20 classes are leaf classes (no extenders in code/src/java or code/src/test), so marking them final is safe and clears the SpotBugs findings without any behavior change.

Verified locally:
- compileJava: BUILD SUCCESSFUL
- tests (pcgen.cdom.*, plugin.lsttokens.choose.*): 4495 tests, 0 failures, 0 errors
- spotbugsMain delta: 470 -> 446 total (-24); all 20 target classes now have 0 CT_CONSTRUCTOR_THROW findings; 5 CT_CONSTRUCTOR_THROW remain in other classes (unchanged baseline)

### DIFF
--- a/code/src/java/pcgen/cdom/base/BasicClassIdentity.java
+++ b/code/src/java/pcgen/cdom/base/BasicClassIdentity.java
@@ -26,7 +26,7 @@ import java.util.Objects;
  * @param <T>
  *            The Format (Class) of the object represented by this ClassIdentity.
  */
-public class BasicClassIdentity<T> implements ClassIdentity<T>
+public final class BasicClassIdentity<T> implements ClassIdentity<T>
 {
 
 	/**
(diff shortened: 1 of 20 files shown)


Vest:
@karianna , I don't mind excluding CT_CONSTRUCTOR_THROW from SpotBugs, I think, we don't develop libraries, and if you agree with me, you/me can exclude this check from SpotBugs.

It is recommended to add a final prefix to a class, that throws an exception in its constructor — this is a vulnerability: https://cmu-sei.github.io/secure-coding-standards/sei-cert-oracle-coding-standard-for-java/rules/object-orientation-obj/obj11-j/

But I don't know, for **whom**.

karianna:
Not so much a vulnerability that might be exploited, but a race condition if we have multu threaded code.  Let's merge this.

Vest:
@karianna then, let's merge it.

PR #7642, opened 2026-06-26
itest: drop 52 JUnit 4-era dummyTest() methods

Vest:
Every concrete `*IntegrationTest` class under `code/src/itest/plugin/lsttokens/editcontext/**` carried an empty

```java
@Test
public void dummyTest()
{
    // Just to get Eclipse to recognize this as a JUnit 4.0 Test Case
}
```

Modern Eclipse + JUnit 5 recognise these classes via their inherited `@Test` methods on the `Abstract*IntegrationTestCase` parents (each parent declares 5–18 real tests). The dummies have not been needed for years; each one padded the report by one passing test that asserted nothing.

### Verification

- `./gradlew compileItestJava` — clean.
- `./gradlew itest --tests "plugin.lsttokens.editcontext.KitIntegrationTest"` — **18 tests, 0 failures, 0 skipped** (was 19 with the dummy; the 18 inherited real tests still run).
- `FactIntegrationTest` and `FactSetIntegrationTest` extend the scaffold-only `AbstractIntegrationTestCase` and therefore inherit no tests — they keep their 7 own real `@Test` methods each, verified running.
- All 52 dummies removed; `grep -rln "dummyTest" code/src/itest` returns nothing.

### Import cleanup

Also drops the now-unused `import org.junit.jupiter.api.Test;` from the 40 files where `dummyTest` was the only local `@Test` usage. The 12 files with additional local `@Test` methods keep the import.

### Stats

```
 52 files changed, 350 deletions(-)
```

Pure cleanup — no production code touched, no test coverage lost.


Vest:
This PR exists, because I am against of useless tests :)
We use JUnit 6, so I hope everything is better.

Vest (commented):
@mertonmonk done!
Thank you!

PR #7644, opened 2026-06-26
Replace pcgen.base.logging SPI with java.lang.System.Logger

Vest:
## Context

PCGen-base shipped a tiny custom logging SPI (`pcgen.base.logging.Logging` + `Severity` enum) whose explicit purpose, per its Javadoc, was to keep the base library free of external dependencies while letting host applications plug a real backend in via `setProcessor(...)`. The JDK has shipped exactly this SPI since Java 9 as `java.lang.System.Logger` + `System.LoggerFinder`, and PCGen's toolchain is now Java 25 (`PCGen-base/build.gradle`, `PCGen-Formula/build.gradle`).

So the custom abstraction is pure tax: a maintenance liability that duplicates a JDK-standard SPI which contributors already recognise on sight.

## What this change does

- **Deletes** `PCGen-base/code/src/java/pcgen/base/logging/Logging.java` and `Severity.java`.
- **Removes** `exports pcgen.base.logging;` from `PCGen-base/code/src/java/module-info.java`.
- **Migrates** the only two callers (both in PCGen-Formula) to `System.Logger`:
  - `pcgen.base.formula.function.GetOptionalFunction`
  - `pcgen.base.formula.visitor.EvaluateVisitor`

The migration pattern is one-for-one:

```diff
-import pcgen.base.logging.Logging;
-import pcgen.base.logging.Severity;
+import java.lang.System.Logger;
+import java.lang.System.Logger.Level;
 …
+private static final Logger LOG = System.getLogger(MyClass.class.getName());
 …
-Logging.log(Severity.WARNING, () -> "msg");
+LOG.log(Level.WARNING, () -> "msg");
```

`System.Logger` provides the `Supplier<String>` overload, so the existing lazy-evaluation pattern carries over unchanged.

## Runtime behaviour

**Unchanged.** The JDK's default `LoggerFinder` routes `System.Logger` to `java.util.logging`, which is exactly the pipeline root `code/`'s `pcgen.util.Logging` (and its `static {}` bootstrap of `logging.properties`, `SourceLogFormatter`, `LoggingRecorder`, …) already configures. The new logger names (`pcgen.base.formula.function.GetOptionalFunction`, `pcgen.base.formula.visitor.EvaluateVisitor`) sit under the existing `pcgen.*` JUL parent, so they pick up current config with no `logging.properties` edits.

`PCGen-base` does **not** need `requires java.logging;` because `System.Logger` / `System.LoggerFinder` / `System.getLogger` all live in `java.base`. The base module is still dependency-free; if a host ever wants to plug in SLF4J / Log4j2 instead of JUL, it does so via the JDK-standard `LoggerFinder` SPI rather than a custom `setProcessor` call.

## Scope, intentional non-scope

In scope:
- PCGen-base SPI deletion + module-info edit
- Two PCGen-Formula call-site migrations

Out of scope (deliberately — separate, independent follow-ups):
- Root `code/`'s `pcgen.util.Logging` is **untouched**. It continues to own JUL bootstrap, `PCGenLogLevel.LST_*` levels, `isDebugMode` toggle, localization, error-count helpers, and `LoggingRecorder` wiring. None of that disappears by switching the base SPI.
- No backend swap (JUL → Log4j2 / Logback / SLF4J).

## Verification

- `grep -rn "pcgen\.base\.logging" PCGen-base PCGen-Formula code` → no matches.
- `./gradlew :PCGen-base:compileJava :PCGen-Formula:compileJava` → BUILD SUCCESSFUL.
- `./gradlew :PCGen-base:test :PCGen-Formula:test` → BUILD SUCCESSFUL.

Diff: **5 files changed, +14 / −126**.


Vest:
I'd like to have a single logging system for the entire project. Until now, we have 3.

PR #7649, opened 2026-06-29
Remove unnecessary 'public' modifier from JUnit 5 test methods and classes

Vest:
## Summary

- Removes the redundant `public` modifier from `@Test`, `@ParameterizedTest`, `@BeforeEach`, `@AfterEach`, `@BeforeAll`, and `@AfterAll` methods across all test source sets
- Removes `public` from top-level concrete test class declarations
- Fixes Sonar rule **S5786** ("Remove this 'public' modifier") across the codebase

JUnit 5 discovers tests by annotation, not by visibility — `public` on test methods and classes is noise, not a requirement.

**Scope:** 903 files changed in `PCGen-base`, `PCGen-Formula`, and the main `code` module (test, testcommon, slowtest, itest source sets).

**Preserved (not changed):**
- `@Override` methods — visibility must not narrow the parent's declaration
- `abstract` base test classes — may be subclassed from other packages
- Helper/mock classes without `@Test` annotations — used across packages

## Test plan

- [x] `./gradlew :PCGen-base:test :PCGen-Formula:test test` passes
- [x] Test count unchanged: **18,490** before and after


Vest:
Since we use JUnit 6, I think, we must be modern - all tests don't need "public" anymore (thanks to Sonar).
Less letters in the code, easier to read :)

mertonmonk (approved):

PR #7686, opened 2026-08-04
Refactor: Replace System.out/err with pcgen.util.Logging to resolve PMD warnings

cpstence:
…MD warnings

@Vest

### DIFF
--- a/code/src/java/pcgen/cdom/base/CDOMObject.java
+++ b/code/src/java/pcgen/cdom/base/CDOMObject.java
@@ -830,45 +830,45 @@ public abstract class CDOMObject extends ConcretePrereqObject
 		}
 		if (integerChar == null ? cdo.integerChar != null : !integerChar.equals(cdo.integerChar))
 		{
-			// System.err.println("CDOM Inequality Integer");
-			// System.err.println(integerChar + " " + cdo.integerChar);
+			// pcgen.util.Logging.log(java.util.logging.Level.FINE, "CDOM Inequality Integer");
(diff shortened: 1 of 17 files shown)


Vest:
few comments:
if you can use "import", do it. 
Logging.errorPrint is better than pcgen.util.Logging.errorPrint (everywhere).
The same goes to java.util.logging.Level.INFO, we have some constants in Logging as well.

These were my findings, and here are from my bot:

> **ShowMessageDelegate**  is a behavior risk. That  println  is the headless/batch fallback for showing a message when there are no GUI observers. Routing it to  Logging.log(INFO)  means it can be filtered/suppressed by log config — a user-facing message in non-GUI mode could silently disappear. Needs care, not a blind swap.
> 
> **Tips.java** half-converted & inconsistent. It's a standalone dev/translation tool that legitimately uses  System.out  — it still keeps  **e.printStackTrace()**  and the now-inaccurate `@SuppressWarnings("UseOfSystemOutOrSystemErr")`. Either leave it fully alone or clean it properly; don't do half.

I need you to leave suppresswarnings and don't touch this and/or you can analyze the Tips.java file, check where it is used, how and see, whether my bot's comment is applicable here.

Regarding the ShowMessageDelegate -- this is serious, I agree with the bot: try to run integration tests (you can try this with the debugger, or change the code like "Vest was here and said: " + whatever is in the code and read the logs, whether your fix doesn't break anything.

Vest (changes requested):
The text is in the PR.

Vest:
The PR was abandoned by the user.

PR #7688, opened 2026-08-09
ARG fixes for FCB

LegacyKing:


Vest:
For the history and Google bot:

```
Broken build analysis — PR #7688 "ARG fixes for FCB" (LegacyKing)

What failed: The slowtest job — specifically 4 Pathfinder character-sheet integration tests:

┌──────────────────────────────┬──────────┬────────┐
│             Test             │ Expected │ Actual │
├──────────────────────────────┼──────────┼────────┤
│ pcGenGUIPfrpgClericTest      │ 107      │ 109    │
├──────────────────────────────┼──────────┼────────┤
│ pcGenGUIPfrpgPaladinTest     │ 113      │ 115    │
├──────────────────────────────┼──────────┼────────┤
│ pcGenGUIPfrpgRogueTest       │ 119      │ 121    │
├──────────────────────────────┼──────────┼────────┤
│ pcGenGUIPfrpgGoldielocksTest │ 131      │ 133    │
└──────────────────────────────┴──────────┴────────┘

All identical in shape: +2 child nodes under /character/ability_objects in the rendered XML.
Every Pathfinder character now carries 2 extra ability objects than the fixtures expect.

Root cause: This line in data/pathfinder/.../core_essentials/ce_abilities.lst:

-CATEGORY=Internal|Default.MOD  ABILITY:Internal|AUTOMATIC|Movement
+CATEGORY=Internal|Default.MOD  ABILITY:Internal|AUTOMATIC|Movement  ABILITY:Internal|AUTOMATIC|Tracker ~ Favored Class Bonus

Default.MOD on the internal category applies to every character globally (which is why the non-caster Rogue is hit,
not just casters). It now auto-grants the Tracker ~ Favored Class Bonus internal ability to all PF characters. That ability
plus its knock-on object account for the +2 in the exported ability list.

Why it's a build break, not necessarily a logic bug: These integration tests compare rendered output against golden
fixtures in code/testsuite/csheets/*.xml. LK's data change legitimately alters every PF character's ability list, but the
fixtures weren't regenerated, so the comparison fails. This is exactly the class of change your memory note covers
("Regenerate csheet fixtures, don't hand-edit").

Two things worth flagging:

1. The fix is to regenerate the csheet fixtures for the 4 affected characters (Cleric, Paladin, Rogue, Goldielocks) —
the +2 abilities are intended, the golden files just need updating. That clears the build.
2. Design question worth raising: granting Tracker ~ Favored Class Bonus to every character via global Default.MOD — including a Rogue and a Cleric who have no FCB tracking need — is a broad hammer. If the intent was only to make the tracker available for FCB abilities to .MOD, it may not need to be AUTOMATIC on everyone. But if it's a deliberate infrastructure ability (a hidden internal tracker), then it's fine and only the fixtures need updating. Worth a one-line confirmation from LK on intent.
```

PR #7700, opened 2026-08-15
Resolve bundled data & settings paths from java.home (fix #7678)

Vest:
## Summary

Fixes PCGen's settings/data path resolution for packaged (jpackage) builds, and cleans up the "options.ini location" dialog. Resolves the root cause behind #7678: bundled data and settings were resolved against `user.dir`, which is `/` for a Finder/double-click launch.

## Changes

**Path resolution (the core fix)**
- Resolve `@`-paths (`@data`, `@plugins`, …) and the "PCGen Dir" settings path from `java.home` (walking up to the install root that holds `data/`+`system/`) instead of `user.dir`. Falls back to `user.dir` for dev/IDE runs.
- Drops the old macOS `user.dir`-ends-with-`MacOS` hack.

**Packaging**
- Embed runtime assets via jpackage `--app-content`, so jpackage builds and signs the whole image in one pass.
- Remove the now-unneeded macOS `MacDirLauncher` wrapper + `Info.plist` rewrite + ad-hoc re-sign tasks (they existed only to work around the old `user.dir` resolution).

**"options.ini location" dialog**
- Fix `//settings` shown for "PCGen Dir" (now resolves to the real install dir).
- Disable the `…` chooser button until "Select a directory" is chosen; open the chooser at the nearest existing ancestor so it no longer silently no-ops.
- Disable "PCGen Dir" when the install dir isn't writable (installed/DMG app), with a "(not writable)" hint.
- Anchor + pad + wrap the layout so the dialog resizes cleanly without truncating labels.
- Correct swapped `userData`: "Mac User Dir" and "Home Dir" mapped to each other's settings directories.

**Misc**
- Add `ConfigurationSettingsTest` covering install-root discovery across the mac/linux/windows bundle layouts and the writability check.
- Clear two pre-existing Sonar findings in `ConfigurationSettings`.

## Verification
- `./gradlew compileJava compileTestJava test` — green (incl. new `ConfigurationSettingsTest`).
- `./gradlew fullJpackage` on macOS: bundle passes `codesign --verify --deep --strict`; launcher starts with no path/data errors; assets embedded at `Contents/app`.
- Verified on the packaged bundle that "PCGen Dir" is enabled from a writable copy and disabled from a read-only DMG mount.

⚠️ Windows/Linux packaged builds were not exercised on the (macOS) dev host — please confirm `fullJpackage` on CI for those before merge.

Refs #7678.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -316,6 +316,9 @@ def hostJdkExt = (hostOs == "windows") ? "zip" : "tar.gz"
 def hostJfxOsPackage = (hostOs == "mac") ? "osx" : hostOs
 def hostJfxArchSuffix = "-${hostArch}"
 
+// Assets staged here (dir named "app") are fed to jpackage via --app-content.
+def jpackageStageDir = layout.buildDirectory.dir("jpackageStage/app")
+
 jlink {
     options = ['--strip-debug', '--no-header-files', '--no-man-pages']
     addExtraDependencies('javafx')
@@ -405,6 +408,9 @@ jlink {
 
         installerOptions = opts
         imageName = "PcGen"
+
+        // Embed runtime assets via --app-content so jpackage builds and signs the whole image.
+        imageOptions = ['--app-content', jpackageStageDir.get().asFile.absolutePath]
     }
 }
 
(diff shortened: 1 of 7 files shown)


Vest:
I hope, I won't break nightly builds. I will approve this PR when it is green.

PR #7711, opened 2026-08-21
perf: improve performance of isType/Type.getConstant and Collator on campaign load

Vest:
## Summary

Targeted, low-risk performance changes on the PCGen campaign-load path, driven by CPU profiling. Each is isolated, behavior-preserving, and backed by a JMH benchmark under a new code/src/jmh source set. Plus two small build/refactor tidy-ups (noted below).

## Findings (from profiling)

Campaign-load CPU showed PObject.isType(String) → Type.getConstant → CaseInsensitiveMap.computeIfAbsent as the dominant leaf, reached from two hot loops (AbilityCategory.populate and
AbstractReferenceManufacturer.resolveGroupReferences). A secondary DataSet.initLists → sortPObjectListByName tower constructed a Collator per comparison.

### Before
image

### After
image

### Performance
1. PObject.isType(Type) fast path — skips uppercase/tokenize/intern when the caller already holds an interned Type; used in AbilityCategory.populate. Micro-bench: single-token 25.8 → 3.0 ns, multi-token 98.6 → 7.8 ns.
2. Loadable.isType(Type) + resolveGroupReferences — default method delegating to isType(String) (so Equipment's EqMod-merged semantics are preserved by not overriding); fast overrides on PObject/ClassSkillList/ClassSpellList. Group-key tokens interned once per manufacturer instead of per object. Group-resolution loop bench: 2989 → 747 ns (~4×). See detailed walkthrough below.
3. Shared Collator in name-sort comparators (CDOMObject.P_OBJECT_NAME_COMP, Comparators.TreeTableNodeComparator). Honest note: the JDK caches Collator rules, so this is an allocation win, not CPU — -prof gc shows 1.87M → 1.32M B/op (~30% less garbage) per 200-element sort; wall time moved only ~5% (noise).

### Build / refactor (not performance)
4. Benchmark task: removed the JavaFX module path and the extractJavaFXLocal dependency — the benchmarks exercise only core data-path classes (PObject/CDOMObject/Type), which never load JavaFX, so there is no reason to pull JavaFX into the benchmark run.
5. Eager-init the `types` set in ClassSkillList/ClassSpellList/DomainSpellList (final field, empty HashSet) and drop the null checks in addType/isType. Behavior-preserving readability cleanup, not a perf change.

## How the resolveGroupReferences optimization works (change 2)

resolveGroupReferences assigns each loaded object to the "TYPE=" group references it qualifies for. It is a nested scan:

```
for each object (getAllObjects)
    for each group reference (typeReferences)
        for each type token in that group's key
            if !object.isType(token) -> object is not in this group
```

A group key is a FixedStringList such as `["Martial","Melee"]` (from a reference like `TYPE=Martial.Melee`), and its elements are plain strings.

**Before:** the inner loop called `object.isType(String)` on each string token. `isType(String)` per call uppercases the string, tokenizes on `.`, and calls `Type.getConstant(token)` to intern each piece into the shared `Type` constant. `Type.getConstant` is a `CaseInsensitiveMap` lookup (hash + case-insensitive equals). Sitting in the innermost loop, that interning ran **objects × groups × tokens** times — e.g. 5000 abilities × 20 group refs × 2 tokens = 200,000 interning calls — even though the group tokens are tiny and fixed for the whole scan.

**After:** intern each group key's tokens to `Type` constants **once**, before the object loop:

```
"Martial" -> Type.getConstant("Martial")   // done once
"Melee"   -> Type.getConstant("Melee")      // done once
```

`getTypeReference` (which builds these keys) already guarantees each element is a single, plain token — it rejects `.`, `=`, `,`, `|` — so each string maps to exactly one `Type.getConstant`, no tokenizing or edge cases. The inner loop then calls the `isType(Type)` fast path (a direct `containsInList(ListKey.TYPE, type)`), a cheap membership test instead of re-interning a string.

Net effect: `Type.getConstant` calls drop from **O(objects × groups × tokens)** to **O(groups × tokens)**. Correctness is preserved for every element type: implementers with special type semantics (notably `Equipment`, whose `isType` merges EqMod-derived types) do **not** override `isType(Type)` and inherit `Loadable`'s default, which delegates back to `isType(String)`.

## Verification

- Guardrails green: Pre{Type,ArmorType,SpellType,Charactertype}RoundRobin, PObjectTest, PreArmorTypeTest, DataSetTest; and the list-class token tests (Mon{C,CC}SkillTokenTest, spell ClassesTokenTest, Spell{known,Level}LstTest) for change 5.
- New unit test AbstractReferenceManufacturerGroupResolutionTest (5 tests) directly pins TYPE-group membership (the resolveGroupReferences path change 2 relies on): single/multi-token matching, case-insensitivity, multi-group membership, and the dangling TYPE= case.
- End-to-end: srdinttest (SRD game-mode integration tests) — 7 tests, 0 failures, confirming TYPE-group resolution is unchanged.
- Benchmarks: ./gradlew benchmark --args='TypeBenchmark ...' / 'SortBenchmark ...'.
- Not done: an end-to-end re-profile of total startup (the isType change is visible in the profiler; the Collator change won't visibly shrink CPU).

## Out of scope

Larger structural towers — LST parsing, plugin classloading, GUI bootstrap — are not addressed here.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -865,6 +865,7 @@ apply from: 'code/gradle/autobuild.gradle'    // depends on distribution.gradle
 apply from: 'code/gradle/reporting.gradle'
 apply from: 'code/gradle/release.gradle'
 apply from: 'code/gradle/plugins.gradle'
+apply from: 'code/gradle/benchmark.gradle'
 
 tasks.register("allTasks") {
     dependsOn build, slowtest, javadoc, allReports
--- a/code/gradle/benchmark.gradle
+++ b/code/gradle/benchmark.gradle
@@ -1,0 +1,40 @@
+/* PCGen JMH micro-benchmarks. Defines the `jmh` source set (code/src/jmh),
+ * its JMH dependencies, and the `benchmark` task that runs them. Kept separate
+ * from build.gradle as it is a developer-only concern, not part of `build`.
+ * Called from the main build.gradle file.
+ *
+ * Usage: ./gradlew benchmark
+ *        ./gradlew benchmark --args='TypeBenchmark -f 1 -wi 3 -i 5 -prof gc'
+ */
+
+sourceSets {
+    jmh {
+        java {
+            srcDirs = ['code/src/jmh']
+        }
+        compileClasspath += sourceSets.main.output + sourceSets.main.compileClasspath
+        runtimeClasspath += sourceSets.main.output + sourceSets.main.runtimeClasspath
+    }
+}
+
+// Declaring the jmh source set above auto-creates these configurations.
+dependencies {
+    jmhImplementation 'org.openjdk.jmh:jmh-core:1.37'
+    jmhAnnotationProcessor 'org.openjdk.jmh:jmh-generator-annprocess:1.37'
+}
+
+tasks.register("benchmark", JavaExec) {
+    group = 'verification'
+    description = 'Runs the JMH micro-benchmarks (code/src/jmh).'
+    dependsOn tasks.named('jmhClasses'), extractJavaFXLocal
+    mainClass = 'org.openjdk.jmh.Main'
+    classpath = sourceSets.jmh.runtimeClasspath
+    // Benchmarks exercise main-source classes, which need JavaFX on the module
+    // path at runtime (same reason the Test.configureEach block in build.gradle
+    // adds it).
+    jvmArgs = [
+            '-Djava.awt.headless=true',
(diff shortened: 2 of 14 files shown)


Vest:
Earlier we were stuck with few micro things that slowed down our parsing, now, only the parsing (mostly) is the slowest part of the PCGen startup.
The PR is draft because I am afraid of merging it now :) I'd better review some things later.

PR #7713, opened 2026-08-22
data: add the SRD 5.2.1 dataset for the 2024 rules

shnaps:
Draft — opened for review of the approach before it is finished off.

Adds `data/5e/wizards_of_the_coast/srd52/`, the **System Reference Document 5.2.1** (1 May 2025) converted for the `5e` game mode. It loads on its own and requires nothing else. Data only; no code is touched.

## What is in it

| | |
| --- | --- |
| Spells | 339 |
| Classes | 12, with subclasses and 20-level progressions |
| Species | 9, with their lineage and ancestry choices |
| Backgrounds / Feats | 4 / 19 |
| Equipment | 38 weapons, 13 armours, 291 items of gear |
| Magic items | 353, from 258 printed entries |
| Kits | 393 — starting equipment, the seven packs, alignments, creature gear |
| Creatures | 235 monsters and 96 animals |
| Creature traits and actions | 1,291 |
| Creature attacks | 398 |

## Creatures are modelled rather than summarised

Each creature locks the six ability scores its stat block prints (`DEFINESTAT:LOCK`, so a Bugbear Stalker is Strength 17 whatever the player rolled), takes its Proficiency Bonus from its Challenge Rating rather than from its Hit Dice, and carries its senses, languages, skills, gear and alignment.

Its traits and actions are objects of their own, split on the labelled Traits / Actions / Bonus Actions / Reactions / Legendary Actions headings the 2024 layout introduced and the 2014 one ran together. Its attacks are weapons it holds, carrying the printed to-hit and damage rather than numbers rebuilt from the creature — a Bugbear Stalker's sheet shows Javelin +5 for 3d6+3 and Morningstar +5 for 2d8+3, which is page 271.

The 5.1 set's shape was the model for this; it defines 71 trait objects across 16 creatures and leaves 325 pointing at a placeholder.

## Licence and provenance

CC-BY-4.0. The `.pcc` carries the required attribution, `ISLICENSED:YES` and `ISOGL:NO`.

Built from the SRD PDF and from Open5e, both CC-BY. **No book content is included**, and nothing from any non-CC-BY source has ever entered this folder — the licence claim is made of nothing but provenance, so that rule is enforced rather than intended. Every object cites the book it was read from, and 3,238 of 3,250 cite the page.

## How it was checked

Two ways, because they answer different questions.

A check suite reads the *written data* back — independently of the generators that wrote it — and holds it against second readings: the printed pages, Open5e, and 5etools' record of SRD 5.2 membership. It also confirms the build against Wizards' published list of what 5.2 adds over 5.1: all 77 entries present, none of the 14 withdrawn. Every check is itself proven breakable by a harness that reintroduces the defect it was written for, so a check that cannot fail is reported rather than trusted.

And loading is not computing, so the numbers were read off characters exported through PCGen: proficiency bonus, the six saves, all 18 skills, Armor Class through each armour category, spell slots and save DCs, species lineages putting spells on the sheet, class features, kits reaching the inventory, and creature attacks.

It loads with **no `SEVERE`, no `LSTERROR` and no `LSTWARN`**.

## Notes for reviewers

- Marked `STATUS:ALPHA`.
- **Do not enable the SRD 5.1 source at the same time.** Both live under the shared `5e` game mode, and both define a `First Level` template and a `Default` internal ability under the same keys; PCGen merges those silently. This is a property of the two sets sharing a game mode, not something either dataset can prevent.
- The 12 objects citing no page are names the book never prints — PCGen plumbing such as `Barbarian Multiclass Training`, and choosers derived from a sentence.
- 27 class features stay as description text on purpose: they are per-use rolls no sheet slot holds. Each is recorded by name with its reason, and the suite fails on a claim recorded nowhere.

Happy to split this differently, rename anything, or match conventions I have missed — hence the draft.


data/5e/wizards_of_the_coast/srd52/_srd52.pcc
```diff
@@ -0,0 +1,149 @@
...
+STAT:srd52__stats.lst
+SIZE:srd52__sizes.lst
+
+# Ability categories must load before the abilities that belong to them, and
+# the base categories before the categories built on top of them.
+ABILITYCATEGORY:srd52_abilitycategories_base.lst
+ABILITYCATEGORY:srd52_abilitycategories.lst
+ABILITYCATEGORY:srd52_abilitycategories_background.lst
+ABILITYCATEGORY:srd52_abilitycategories_class.lst
+ABILITYCATEGORY:srd52_abilitycategories_species.lst
```
LegacyKing, line 84: This should all be contained within one file.

data/5e/wizards_of_the_coast/srd52/_srd52.pcc
```diff
@@ -0,0 +1,149 @@
...
+
+# Species traits load before the species that grant them.
+ABILITY:srd52_abilities_species.lst
+ABILITY:srd52_abilities_lineage.lst
+# And a creature's traits and actions before the creature.
+ABILITY:srd52_abilities_monster.lst
+LANGUAGE:srd52_languages.lst
+BIOSET:srd52_biosettings.lst
+RACE:srd52_races.lst
+RACE:srd52_races_unselected.lst
```
LegacyKing, line 106: Why is there an "unselected" group?

data/5e/wizards_of_the_coast/srd52/_srd52.pcc
```diff
@@ -0,0 +1,149 @@
...
+SPELL:srd52_spells.lst
+
+# Class features load before the classes that grant them by name, and the
+# first-level templates after the classes their PRECLASS names.
+ABILITY:srd52_abilities_class.lst
+
+# The weapon masteries, one per weapon. Nothing grants them by name: the five
+# Weapon Mastery features raise a pool and these are what it is spent on, so
+# their only tie to the rest is the PREPROFWITHWEAPON each one carries.
+ABILITY:srd52_abilities_mastery.lst
```
LegacyKing, line 119: If it's part of a class feature it should exist within the same file as the rest. No reason to create a separate file.

data/5e/wizards_of_the_coast/srd52/srd52__align.lst
```diff
@@ -0,0 +1,15 @@
+# The 2024 rules give player characters no alignment mechanics, but stat
+# blocks still print one and the game mode still requires the axis to
+# exist with a default. Nothing in this dataset reads it.
+
```
LegacyKing, line 4: if the options don't matter, then a single line of 'n/a' should be fine, and I believe alignments can be disabled.

data/5e/wizards_of_the_coast/srd52/srd52_abilities_class.lst
```diff
@@ -0,0 +1,413 @@
...
+Unarmored Defense	KEY:Barbarian ~ Unarmored Defense	CATEGORY:Special Ability	TYPE:Barbarian Class Feature.Class Feature.SpecialQuality	SOURCEPAGE:SRD 5.2.1, p.29	BONUS:COMBAT|AC|CON|TYPE=Ability.STACK|!PREEQUIP:1,TYPE=Armor	DESC:While you aren’t wearing any armor, your base Armor Class equals 10 plus your Dexterity and Constitution modifiers. You can use a Shield and still gain this benefit.
```
LegacyKing, line 10: Just noticed, all the SOURCEPAGE entries are added the source name, remove that and only use page numbers, p.x format is fine. Leaving this as is will have messy sources with "SRD 5.2.1 SRD 5.2.1, p.29" on the outputs that pull in SourceDisplay and exports

LegacyKing (changes requested):
Partial Review

shnaps:
Moving this to PCGen/pcgen-newsources#634 as suggested — thanks for the review.

All four of your other points are applied there:

- `SOURCEPAGE` now carries the page alone. You were right about the doubling: `SourceFormat.java:230` assembles `publisher - source, <page>` itself, so sheets were printing the source name twice.
- The five ability-category files are one.
- The weapon masteries moved into `srd52_abilities_class.lst` with the rest of the class features.
- The unselected race is now the first line of the species file.

On alignment: you were reading a comment of mine that was **wrong**, and I have corrected it rather than acted on it. It claimed the 2024 rules give player characters no alignment mechanics and that nothing reads it. Choosing one is step 4 of character creation (p.21), all nine plus Unaligned are defined on pp.21-22, and three things read it — the Sprite's Heart Sight (p.329), Nondetection and the Ring of Mind Shielding (pp.149, 238), and sentient magic items (pp.207-8). For creatures the book says an alignment is "suggested for the monster, with the GM determining its actual alignment" (p.188), which is what the creature kits do. The file is also identical to `srd5__align.lst`, and `ALIGNMENTFEATURE:YES` lives in the shared `5e` game mode, so disabling would reach the 5.1 set too. Still happy to drop it if you would rather — it is your call, and the details are in the new PR.

That mistake prompted a sweep of every factual claim shipped in these files; it found three more and one real data defect (the Epic Boons raise a score to 30 and nothing raised the cap). There is now a check that prose stating a rule must cite a page or quote the book.

PR #7714, opened 2026-08-23
Fix tips logging

cpstence:
Summary
Addresses reviewer feedback regarding the centralized logging refactor by updating the output routing in Tips.java.

Changes Made
Refactored Tips.java to route all tip and localization message outputs through pcgen.util.Logging instead of relying on standard system output streams.

Ensured that routing these messages to Logging.log(INFO) maintains visibility for user-facing messages during non-GUI mode operations.

Testing
Compiled successfully and ran integration checks via Gradle.

Verified that boot sequence messages and tips are properly captured by the logging utility without being suppressed when the graphical interface is bypassed.

### DIFF
--- a/code/src/java/translation/util/Tips.java
+++ b/code/src/java/translation/util/Tips.java
@@ -16,6 +16,10 @@
  * Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA 02111-1307 USA
  */
 package translation.util;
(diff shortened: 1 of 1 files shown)


Vest (changes requested):
check the imports, don't use fqdm for Logging.

Vest:
The file Tips.java has been already refactored.

PR #7730, opened 2026-09-03
DATA-4391: Fix Kitsune starting age and aging milestones

cpstence:
**Issue:**
* The PCGen engine erroneously treated age 7-14 as the adult age range for Kitsune characters, and treated anything past 15 as "middle-aged"[cite: 4].
* The engine evaluated age 0 as a "youth" category, which automatically applied round-the-board -2 stat modifiers[cite: 4].
* The *Everyman Gaming Kit Compendium* forced subtype traits to be mandatory rather than optional[cite: 4].


cpstence:
**Fix:**
* Cleared illegal `AGE` tags from the main race definition in `kitsune_races.lst`.
* Rewrote the demographic brackets in `kitsune_biosettings.lst` to remove the `Undefined` (0) and `Youth` (7) tiers. Kitsune adulthood now correctly begins at `BASEAGE:15` to align with standard Pathfinder milestone math.
* Injected a `None (Standard Paizo Kitsune)` choice into the `Kitsune Racial Subtype` ability pool in `egkc_abilities_race.lst`. This creates a clean opt-out path so players can select the base race without triggering validation errors from 3rd-party clan requirements.

